PR #5, opened 2017-09-26
Results writer

moldover:
This implements the bulk of the results writer functionality.  Still remaining is displaying the winner (pending the format we'll use) and basic "header" data such as election name and total votes cast.


src/com/rcv/RCVTester.java
```diff
@@ -37,6 +37,9 @@ public static int runTests() {
           Tabulator.OvervoteRule.EXHAUST_IF_ANY_CONTINUING);
       try {
         testTabulator.tabulate();
+        testTabulator.generateSummarySpreadsheet(TEST_RESULTS_XLSX_FILE);
+
```
tarheel: delete blank lines

src/com/rcv/ResultsWriter.java
```diff
@@ -0,0 +1,228 @@
...
+ *
+ * Helper class takes tabulation results data as input and generates results file
+ * Currently we support an xlsx spreadsheet which can be visualized in a web browser
+ *
+ */
+public class ResultsWriter {
+
+  // takes tabulation round tallies and generates a spreadsheet from them
+  public boolean generateSummarySpreadsheet(
+      int finalRound,
```
tarheel: reduce indent

src/com/rcv/ResultsWriter.java
```diff
@@ -0,0 +1,228 @@
...
+    // TODO: write the election header info here (contest name, total votes cast, threhold, etc..)
+
+    /////////////////////////////////////
+    // Round-by-round elections results
+    /////////////////////////////////////
+
+    // Row 0: Round headers
+    int rowCounter = 0;
+    org.apache.poi.ss.usermodel.Row headerRow = worksheet.createRow(rowCounter++);
+    for(int round = 1; round <= finalRound; round++) {
```
tarheel: nit: space after for

src/com/rcv/ResultsWriter.java
```diff
@@ -0,0 +1,228 @@
...
+      sb.append(round).append(": ");
+      List<String> eliminated = roundToCandidatesEliminated.get(round);
+      if (eliminated != null) {
+        String cellText = String.join(", ", eliminated);
+        int cellIndex = ((round-1)*2)+1;
+        Cell cell = eliminationsRow.createCell(cellIndex);
+        cell.setCellValue(cellText);
+        sb.append(cellText);
+      } else {
+        sb.append("none");
```
tarheel: Should this be possible? Seems like an error condition.

src/com/rcv/ResultsWriter.java
```diff
@@ -0,0 +1,228 @@
...
+        String percentageText = String.format("%.2f%%", lastRoundPercentage);
+        percentageCell.setCellValue(percentageText);
+        sb.append(percentageText);
+      }
+    }
+
+    // write to log
+    RCVLogger.log(sb.toString());
+    
+    // write xls to disk
```
tarheel: have you heard of this "cloud" thing?

src/com/rcv/ResultsWriter.java
```diff
@@ -0,0 +1,228 @@
...
+    List<String> sortedCandidates = new LinkedList<String>();
+    for (Map.Entry<String, Integer> entry : entries) {
+      sortedCandidates.add(entry.getKey());
+    }
+    return sortedCandidates;
+  }
+
+
+
+}
```
tarheel: shame

src/com/rcv/Tabulator.java
```diff
@@ -25,6 +25,18 @@
   private Integer maxNumberOfSkippedRanks = 1;
   private OvervoteRule overvoteRule = OvervoteRule.EXHAUST_IF_ANY_CONTINUING;
 
+  public Integer roundsToDecideWinner = -1;
```
tarheel: remove

tarheel:
Same thing here

src/com/rcv/ResultsWriter.java
moldover: if we stop at the penultimate round then this would not happen.  if we have a column for the final round (as I would prefer, just for simplicity) then that round will have no eliminations

src/com/rcv/ResultsWriter.java
moldover: I can't go there, apologies.  if you want to change it that is fine.

src/com/rcv/ResultsWriter.java
tarheel: So in the penultimate round, the runner-up is eliminated? What happens in the final round? (How is it different from the penultimate round?)

src/com/rcv/ResultsWriter.java
tarheel: Ha, OK. If you don't mind me fixing random things like this when I encounter them in the code, then I won't nitpick them in PRs.

src/com/rcv/ResultsWriter.java
moldover: no, I got that wrong.  there should still be an elimination(s) in each round.  there's a bug where the final elimination isn't recorded by the tabulator (the main loop just exits) and so the final elimination isn't written to the spreadsheet

tarheel: Ah. Are you going to send out a PR for the bug fix? Or I can take a look later.

moldover: Just pushed a fix to this same branch.

tarheel: Thanks. Seems legit.

tarheel:
I'll wait for you to merge and then I'll resolve conflicts and merge mine.

PR #18, opened 2017-11-20
Audit refactor 1

moldover:
Here's the basics for doing the audit trail.  Turned out to be pretty straight forward.


src/com/rcv/CVRReader.java
```diff
@@ -48,26 +48,20 @@ public boolean parseCVRFile(String excelFilePath,
...
     // Iterate through all rows and create a CastVoteRecord for each row
     while (iterator.hasNext()) {
       org.apache.poi.ss.usermodel.Row castVoteRecord = iterator.next();
 
-      // create object for this row
-      ArrayList<ContestRanking> ballot = new ArrayList<ContestRanking>();
+      // TODO: determine how ballot IDs will be handled for different ballot styles
+      String ballotID =  String.format("%s(%d)",cvrFileName,ballotIndex++);
 
-      // parse ID
```
tarheel: I think we should keep this logic. When the ballot does contain IDs, let's use them in addition to our own IDs. We could just have two separate vars in the CVR class.

src/com/rcv/CastVoteRecord.java
```diff
@@ -1,23 +1,50 @@
...
+  // followed by reason for exhaustion if it is ever exhausted
+  public List<String> mDescriptionsByRound = new ArrayList<>();
 
-  public CastVoteRecord() {}
+  // adds the string to this CVR round by round descriptions for auditing
+  public void addRoundDescription(String description) {
+    mDescriptionsByRound.add(description);
+  }
+  
+  // output is our rankings sorted from high to low
```
tarheel: You should clarify what you mean by "high to low", since that's ambiguous when you're talking about rankings. Maybe "more preferred to less preferred"?

src/com/rcv/CastVoteRecord.java
```diff
@@ -1,23 +1,50 @@
...
+  SortedMap<Integer, Set<String>> mSortedRankings;
 
-  @JsonAnySetter
-  public void add(String key, ContestRankings value) {
-    Integer electionId = Integer.parseInt(key);
-    rankings.put(electionId, value);
-  }
+  // mDescriptionsByRound contains who this ballot counted for in each round
+  // followed by reason for exhaustion if it is ever exhausted
+  public List<String> mDescriptionsByRound = new ArrayList<>();
```
tarheel: I recognize that technically an ArrayList can work here, but I'd rather use a map from Integer to String to make it explicit which round maps to which description. And then we wouldn't have to reference the round directly in the description string, and could instead store it in a structured way.

src/com/rcv/CastVoteRecord.java
```diff
@@ -1,23 +1,50 @@
...
 
-  public CastVoteRecord() {}
+  // adds the string to this CVR round by round descriptions for auditing
+  public void addRoundDescription(String description) {
```
tarheel: And then I'd prefer to have this method take an int round as an arg also.

src/com/rcv/Tabulator.java
```diff
@@ -456,16 +459,16 @@ private boolean hasContinuingCandidates(
       }
     }
 
-    // loop over the ballots and count first-place votes, considering only continuing candidates
-    for (int i = 0; i < allSortedRankings.size(); i++) {
+    // loop over the ballots and count votes for continuing candidates
+    for (int i = 0; i < castVoteRecords.size(); i++) {
       if (exhaustedBallots.get(i) != null) {
```
tarheel: Now that we're storing tabulation state in the CVR, we should maybe move this in there also. (No need to do that in this diff.)

src/com/rcv/Tabulator.java
```diff
@@ -552,4 +530,16 @@ private int numCandidates() {
     }
     return num;
   }
+
+  public void doAudit(List<CastVoteRecord> castVoteRecords) {
+    for(CastVoteRecord cvr : castVoteRecords) {
+      String auditString = String.format("%s:",cvr.mBallotID);
```
tarheel: Use a StringBuilder instead. Because Strings are immutable in Java, building up a long string through concatenation is very inefficient -- it will instantiate a new object each time you concatenate.

moldover: will do

src/com/rcv/CastVoteRecord.java
moldover: done

src/com/rcv/CastVoteRecord.java
moldover: done

src/com/rcv/CVRReader.java
moldover: Going to leave as is until we hear otherwise from product.  It's easy to put back if we need to.

PR #40, opened 2018-04-08
Initial GUI implementation

HEdingfield:


src/com/rcv/Logger.java
```diff
@@ -49,6 +50,11 @@ public static void setup(String file_output_path) throws IOException {
     // controls logging console output filters
     ConsoleHandler consoleHandler = new ConsoleHandler();
     consoleHandler.setFormatter(formatter);
+    // remove any existing handlers to prevent duplicate logging
+    for (Handler h : logger.getHandlers()) {
+      logger.removeHandler(h);
+    }
+    // add new handlers
```
moldover: Nice.  I've run into this in python coding when running scripts from test harness so this makes sense.

src/com/rcv/Main.java
```diff
@@ -23,75 +23,89 @@
...
+      // generate visualizer spreadsheet data
+      tabulator.generateSummarySpreadsheet();
+      // generate audit data
+      tabulator.doAudit(castVoteRecords);
+    } catch (Exception exception) {
+      response = String.format("ERROR during tabulation: %s", exception.toString());
+      Logger.log(response);
+    }
+    return response;
+  }
```
moldover: Cool, looks good.  It does seem a little odd to return a user-facing status string, but it's fine for now.

src/com/rcv/RcvGui.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+
+  class TabulateListener implements ActionListener {
+    public void actionPerformed(ActionEvent event) {
+      String response = Main.executeTabulation(config);
+      labelStatus.setText(response);
+    }
+  }
+}
```
moldover: Looks good.  The EAC coding standards are a bit pedantic in that they want: 
a header comment block for every function including purpose, params and return
also a comment for EVERY variable (only exception is loop index variables)
and a comment for EVERY constant (except 1, -1, and 0)

src/com/rcv/Logger.java
```diff
@@ -1,37 +1,38 @@
-/**
- * Created by Jonathan Moldover and Louis Eisenberg
- * Copyright 2018 Bright Spots
- * Purpose: Wrapper for console and file logging functions
- * All logging and audit output should use these methods to ensure output goes into audit file
- * Version: 1.0
+/*
+  Created by Jonathan Moldover and Louis Eisenberg
```
tarheel: Feel free to add your name to these headers if you'd like.

src/com/rcv/Logger.java
```diff
@@ -1,37 +1,38 @@
-/**
- * Created by Jonathan Moldover and Louis Eisenberg
- * Copyright 2018 Bright Spots
- * Purpose: Wrapper for console and file logging functions
- * All logging and audit output should use these methods to ensure output goes into audit file
- * Version: 1.0
+/*
```
tarheel: Here and elsewhere: the style guide says that each of the internal lines in a multi-line comment should start with an asterisk.

https://google.github.io/styleguide/javaguide.html#s4.8.6-comments

src/com/rcv/Main.java
```diff
@@ -23,75 +23,89 @@
...
           Logger.log("Please see the README.txt for details.");
+          config = null;
         }
       }
-    } else {
-      System.err.println("No election configuration file specified as input.");
-      System.err.println("Please see the README.txt for details.");
+    } catch (IOException exception) {
+      System.err.print(
+          String.format(
```
tarheel: here and next few lines: we've been indenting with 2 spaces, not 4

src/com/rcv/Main.java
```diff
@@ -23,75 +23,89 @@
...
   }
 
   // function: makeElectionConfig
-  // purpose: create config object from command line args
-  // param: args command line arguments passed into main program
+  // purpose: create config object
+  // param: path to config file
   // returns: the new ElectionConfig object or null if there was a problem
-  public static ElectionConfig makeElectionConfig(String [] args) {
+  static ElectionConfig makeElectionConfig(String configPath) {
```
tarheel: should we call this `loadElectionConfig`?

src/com/rcv/Main.java
```diff
@@ -23,75 +23,89 @@
...
+    }
+    return response;
+  }
+
   // function: parseCastVoteRecords
   // purpose: parse cvr files referenced in the ElectionConfig object into a list of CastVoteRecords
   // param: config object containing cvr file paths to parse
   // returns: list of all CastVoteRecord objects parsed from cvr files
-  public static List<CastVoteRecord> parseCastVoteRecords(ElectionConfig config)
+  private static List<CastVoteRecord> parseCastVoteRecords(ElectionConfig config)
```
tarheel: If a function signature doesn't fit in 100 chars, format it like this:

```
private static List<CastVoteRecord> parseCastVoteRecords(
  ElectionConfig config
) throws Exception {
```

src/com/rcv/RcvGui.java
```diff
@@ -0,0 +1,71 @@
+package com.rcv;
+
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.awt.BorderLayout;
+import javax.swing.*;
```
tarheel: Style guide says no wildcards. :/

https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

src/com/rcv/Logger.java
HEdingfield: Cool, fixing this, but still changing /** to /* because the former indicates a docstring comment and isn't recommended by the style guide for normal block comments.

I also loaded the Google style guide config into IntelliJ from here: https://github.com/google/styleguide#google-style-guides

src/com/rcv/RcvGui.java
HEdingfield: Good to know, thanks. What if I document a function like this:

`// returns: String indicating whether or not execution was successful`
`static String executeTabulation(ElectionConfig config) {`

Do I also need to apply the same comment inside the function when actually defining the response variable?

`// String indicating whether or not execution was successful`
`String response = "Tabulation successful!";`

And then again when I actually call the function?

`// String indicating whether or not execution was successful`
`String response = Main.executeTabulation(config);`

src/com/rcv/RcvGui.java
HEdingfield: Good! Wildcard imports suuuuuuuuuuuck

PR #41, opened 2018-04-12
Address IntelliJ warnings

HEdingfield:


src/com/rcv/Tabulator.java
```diff
@@ -112,10 +112,10 @@
   }
 
   // map of round to TieBreak objects to record how tiebreaks were decided
-  private Map<Integer, TieBreak> roundToTieBreak = new HashMap<>();
+  private final Map<Integer, TieBreak> roundToTieBreak = new HashMap<>();
```
HEdingfield: roundToTieBreak is updated here but never queried... should we do anything about this?

src/com/rcv/CastVoteRecord.java
```diff
@@ -149,22 +144,24 @@ private void sortRankings(List<Pair<Integer, String>> rankings) {
...
 
   // function: getAuditString
-  // purpose: return a formatted string describing this cvr and how state changes over
+  // purpose: return a formatted string describing this CVR and how state changes over
   //  the course of the tabulation.  Used for audit output.
   // returns: the formatted string for audit output
   String getAuditString() {
     // use a string builder for more efficient string creation
     StringBuilder auditStringBuilder = new StringBuilder();
-    auditStringBuilder.append(AUDIT_LABEL_SOURCE);
```
tarheel: As I recall, we pulled these out into constants to try to comply with the code standards. I agree that it's not generally helpful to have constants for strings like this that you only use in one place.

src/com/rcv/Tabulator.java
tarheel: Yeah, I would just delete this map.

src/com/rcv/CastVoteRecord.java
tarheel: > I didn't see any other instances like this in the code (even though there are other cases where String text is input directly into functions). Is it ok to go through with this change and see if anyone complains? It really doesn't make any sense, as you pointed out.

Yeah, go ahead and change it for now.

> Tabulator.java line 489: would it be better / safer to compare < 0, rather than == -1?

Yes, this would be safer.

> ResultsWriter.java line 357: the value changed at rowCounter++ is never used; was this intended to be ++rowCounter?

No, the postfix is intended. We do this a number of times in this method; it happens that this is the last time we do it, but we might add a new reference to it below here in the future. I'd prefer to ignore IntelliJ's warning in this case

PR #50, opened 2018-04-22
fix division logic (again) and update ResultsWriter headers to reflect multi-winner properly

tarheel:
I realized that the division logic was still wrong when I started rendering the threshold percentages in the results spreadsheet. I think it's finally correct now.

### DIFF
--- a/src/com/rcv/ElectionConfig.java
+++ b/src/com/rcv/ElectionConfig.java
@@ -172,6 +172,15 @@ class ElectionConfig {
         rawConfig.rules.decimalPlacesForVoteArithmetic;
   }
 
+  // function: divide
+  // purpose: perform a division operation according to the config settings
+  // param: dividend is the numerator in the division operation
+  // param: divisor is the denominator in the division operation
+  // returns: the quotient
+  public BigDecimal divide(BigDecimal dividend, BigDecimal divisor) {
+    return dividend.divide(divisor, decimalPlacesForVoteArithmetic(), RoundingMode.HALF_EVEN);
+  }
+
   // function: roundDecimal
   // purpose: round a number according to the config settings
   // returns: the rounded value
--- a/src/com/rcv/ResultsWriter.java
+++ b/src/com/rcv/ResultsWriter.java
@@ -259,8 +259,7 @@ public class ResultsWriter {
         //  total active votes in this round
         BigDecimal totalActiveVotes = totalActiveVotesPerRound.get(dataUseRound);
         // fractional percent
-        BigDecimal fraction =
-            config.roundDecimal(thisRoundTally.divide(totalActiveVotes, RoundingMode.HALF_EVEN));
+        BigDecimal fraction = config.divide(thisRoundTally, totalActiveVotes);
         // percentage of active votes
         BigDecimal percentage = fraction.signum() == 1 ?
             fraction.multiply(new BigDecimal(100)) :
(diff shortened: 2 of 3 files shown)


src/com/rcv/ResultsWriter.java
```diff
@@ -479,6 +476,20 @@ private int addHeaderRows(
...
+    String electionType = config.numberOfWinners() > 1 ? "multi-winner" : "single-winner";
+    // number reflecting threshold percentage to win
+    BigDecimal thresholdPercentage = config.divide(
+        new BigDecimal(100),
+        new BigDecimal(config.numberOfWinners() + 1)
+    );
+    // string for threshold percentage
+    String thresholdString = thresholdPercentage.toString() + "%";
+    // formula string
+    String formula = config.numberOfWinners() > 1 ?
```
HEdingfield: Maybe this should instead check electionType, rather than duplicate the same logic here?

src/com/rcv/ResultsWriter.java
```diff
@@ -479,6 +476,20 @@ private int addHeaderRows(
...
+        "More than 1/" + (config.numberOfWinners() + 1) + " of total votes cast for office" :
+        "Half of total votes cast for office + 1";
```
HEdingfield: Make wording consistent between these formulas, e.g. either "more than half of total votes cast for office" or "More than 1/* of total votes cast for office + 1".

src/com/rcv/ResultsWriter.java
```diff
@@ -496,16 +507,15 @@ private int addHeaderRows(
...
-        {"Graph threshold label", "50% of votes required to win", OutputType.STRING},
+        {"Threshold number", thresholdString, OutputType.STRING},
+        {"Graph threshold label", thresholdString + " of votes required to win", OutputType.STRING},
         {null, null, OutputType.STRING},
         {"Contest summary data", null, OutputType.STRING},
         {"Tally detail", null, OutputType.STRING},
-        {"Single/multi winner", "single-winner", OutputType.STRING},
-        // TODO: update for multi-winners
-        {"Number to be elected", 1, OutputType.INT},
+        {"Single/multi winner", electionType, OutputType.STRING},
```
HEdingfield: "Single-/multi-winner"?

src/com/rcv/ResultsWriter.java
tarheel: I think checking electionType would be a bit more fragile, because if I changed that string, I'd have to change it here also.

src/com/rcv/ResultsWriter.java
tarheel: OK, fair enough. I was trying to preserve the language that Whitney had given us, but yeah, we should just be consistent.

src/com/rcv/ElectionConfig.java
```diff
@@ -172,6 +172,13 @@ private Integer decimalPlacesForVoteArithmetic() {
         rawConfig.rules.decimalPlacesForVoteArithmetic;
   }
 
+  // function: divide
+  // purpose: perform a division operation according to the config settings
+  // returns: the quotient
+  public BigDecimal divide(BigDecimal dividend, BigDecimal divisor) {
+    return dividend.divide(divisor, decimalPlacesForVoteArithmetic(), RoundingMode.HALF_EVEN);
```
moldover: just needs param comments

PR #51, opened 2018-04-23
add logic to Tabulator to compile the round tallies per precinct

tarheel:
Next step in per-precinct output. The final step will be to refactor ResultsWriter to display this computed data.

### DIFF
--- a/src/com/rcv/CastVoteRecord.java
+++ b/src/com/rcv/CastVoteRecord.java
@@ -119,6 +119,12 @@ class CastVoteRecord {
     this.currentRecipientOfVote = currentRecipientOfVote;
   }
 
+  // function: getPrecinct
+  // purpose: getter for precicnt
+  // returns: value of field
+  public String getPrecinct() {
+    return precinct;
+  }
 
   // function: sortRankings
   // purpose: create a map of ranking to candidates selected at that rank
@@ -150,8 +156,10 @@ class CastVoteRecord {
     auditStringBuilder.append(sourceName);
     auditStringBuilder.append(" [Ballot ID] ");
     auditStringBuilder.append(cvrID);
-    auditStringBuilder.append(" [Precinct] ");
-    auditStringBuilder.append(precinct);
+    if (precinct != null) {
+      auditStringBuilder.append(" [Precinct] ");
+      auditStringBuilder.append(precinct);
+    }
     auditStringBuilder.append(" [Round by Round Report] |");
     // index to to iterate over all round descriptions
     for (Integer round : descriptionsByRound.keySet()) {
--- a/src/com/rcv/Tabulator.java
+++ b/src/com/rcv/Tabulator.java
@@ -21,7 +21,7 @@ import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
 
-public class Tabulator {
+class Tabulator {
 
   // When the CVR contains an overvote we "normalize" it to use this string
   static final String explicitOvervoteLabel = "overvote";
@@ -35,6 +35,9 @@ public class Tabulator {
   // e.g. roundTallies[1] contains a map of all candidate ID -> votes for each candidate in round 1
   // this structure is computed over the course of tabulation
   private final Map<Integer, Map<String, BigDecimal>> roundTallies = new HashMap<>();
+  // precinctRoundTallies is a map from precinct to roundTallies for that precinct
+  private final Map<String, Map<Integer, Map<String, BigDecimal>>> precinctRoundTallies =
+      new HashMap<>();
   // candidateToRoundEliminated is a map from candidate ID to round in which they were eliminated
   private final Map<String, Integer> candidateToRoundEliminated = new HashMap<>();
   // map from candidate ID to the round in which they won
@@ -55,6 +58,7 @@ public class Tabulator {
     this.castVoteRecords = castVoteRecords;
     this.candidateIDs = config.getCandidateCodeList();
     this.config = config;
+    initPrecinctRoundTallies();
   }
 
   // function: log
@@ -77,7 +81,7 @@ public class Tabulator {
   // param: shouldLog is set to log to console and log file
   // return: sorted map of tally to List of candidateIDs drawn from the input data and excluding
   //   candidates not appearing in candidatesToInclude)
-  public static SortedMap<BigDecimal, LinkedList<String>> buildTallyToCandidates(
+  static SortedMap<BigDecimal, LinkedList<String>> buildTallyToCandidates(
       Map<String, BigDecimal> roundTally,
       Set<String> candidatesToInclude,
       boolean shouldLog
@@ -104,10 +108,10 @@ public class Tabulator {
     return tallyToCandidates;
   }
 
+  // function: tabulate
   // purpose: run the main tabulation routine to determine election results
   //  this is the high-level control of the tabulation algorithm
-  public void tabulate() {
-
+  void tabulate() {
     log("Beginning tabulation for contest.");
     log("There are %d candidates for this contest:", config.numCandidates());
     // string indexes over all candidate IDs to log them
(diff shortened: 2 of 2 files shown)


src/com/rcv/Tabulator.java
```diff
@@ -55,6 +58,7 @@
     this.castVoteRecords = castVoteRecords;
     this.candidateIDs = config.getCandidateCodeList();
     this.config = config;
+    initPrecinctRoundTallies();
```
HEdingfield: Might be more optimized if we check to see in the first place if the config specifies that we should group by precinct before initializing this.

src/com/rcv/Tabulator.java
```diff
@@ -585,14 +592,12 @@ private boolean hasContinuingCandidates(SortedMap<Integer, Set<String>> rankToCa
...
+    Map<String, BigDecimal> roundTally = getNewTally();
 
-    // initialize round tallies to 0 for all continuing candidates
-    // candidateID indexes through all candidateIDs
-    for (String candidateID : candidateIDs) {
-      if (getCandidateStatus(candidateID) == CandidateStatus.CONTINUING) {
-        roundTally.put(candidateID, BigDecimal.ZERO);
-      }
+    // map of tallies per precinct for this round
+    Map<String, Map<String, BigDecimal>> roundTallyByPrecinct = new HashMap<>();
```
HEdingfield: Again, should we check if grouping by precincts is specified in the config first before doing this?

src/com/rcv/Tabulator.java
```diff
@@ -675,13 +680,62 @@ private boolean hasContinuingCandidates(SortedMap<Integer, Set<String>> rankToCa
...
+      Map<Integer, Map<String, BigDecimal>> roundTalliesForPrecinct =
+          precinctRoundTallies.get(precinct);
+      roundTalliesForPrecinct.put(currentRound, roundTallyByPrecinct.get(precinct));
+    }
+
     return roundTally;
   }
 
+  // function: getNewTally
+  // purpose: create a new initialized tally with all continuing candidates
```
HEdingfield: Needs params documented.

src/com/rcv/Tabulator.java
```diff
@@ -675,13 +680,62 @@ private boolean hasContinuingCandidates(SortedMap<Integer, Set<String>> rankToCa
...
+  ) {
+    // current tally for this candidate
+    BigDecimal currentTally = tally.get(selectedCandidateID);
+    // new tally after adding this vote
+    BigDecimal newTally = currentTally.add(cvr.getFractionalTransferValue());
+    tally.put(selectedCandidateID, newTally);
+  }
+
+  // function: initPrecinctRoundTallies
+  // purpose: initialize the map tracking per-precinct round tallies
```
HEdingfield: Needs params documented.

src/com/rcv/Tabulator.java
```diff
@@ -675,13 +680,62 @@ private boolean hasContinuingCandidates(SortedMap<Integer, Set<String>> rankToCa
...
+    for (String candidateID : candidateIDs) {
+      if (getCandidateStatus(candidateID) == CandidateStatus.CONTINUING) {
+        tally.put(candidateID, BigDecimal.ZERO);
+      }
+    }
+    return tally;
+  }
+
+  // function: incrementTally
+  // purpose: add a vote (or fractional share of a vote) to a tally
```
HEdingfield: Needs params documented.

src/com/rcv/CastVoteRecord.java
```diff
@@ -119,6 +119,12 @@ public void setCurrentRecipientOfVote(String currentRecipientOfVote) {
     this.currentRecipientOfVote = currentRecipientOfVote;
   }
 
+  // function: getPrecinct
+  // purpose: getter for precicnt
+  // returns: value of field
+  public String getPrecinct() {
```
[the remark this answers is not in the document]
tarheel: What's the advantage of using the getter instead of referencing the instance variable directly? I googled this for a minute and found arguments in both directions. The highlight was definitely from [here](https://www.reddit.com/r/java/comments/2f3flb/is_it_considered_better_practice_for_a_class_to/): "Your analysis is akin to comparing the practice to wearing a condom while masturbating."

src/com/rcv/Tabulator.java
tarheel: There's no specific setting for this in the config; there's just an optional field to indicate which column has the precinct in the CVR file. But if no CVRs have precincts, the precinct tallying logic will basically be a no-op... so I think it's fine?

src/com/rcv/Tabulator.java
tarheel: Oops. Thanks for flagging (here and below).

src/com/rcv/Tabulator.java
tarheel: Actually, this is the only one of the three new methods I introduced that has any params.

src/com/rcv/Tabulator.java
tarheel: Actually, this is the only one of the three new methods I introduced that has any params.

src/com/rcv/CastVoteRecord.java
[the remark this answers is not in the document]
HEdingfield: Wondering why you added it here in the first place? I think the general philosophy is that it's good practice to use getters and setters rather than modifying variables directly [for the reasons outlined here](https://stackoverflow.com/a/1568230/3846321). From personal experience, wanting to add validation later without having to change the references everywhere in your code is really nice (i.e. encapsulation). But since there's not much previous precedent for doing that in this project, maybe don't bother starting until you're actually doing validation of some sort.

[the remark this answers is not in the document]
HEdingfield: On the [other hand](https://stackoverflow.com/a/7093579/3846321)... I'd lean towards not using them I guess... either way, consistency is good. I still need to read Effective Java!

src/com/rcv/Tabulator.java
HEdingfield: Ahh yeah, good point. I've been including all the fields (including return for void) but putting "N/A" in cases like this.

src/com/rcv/Tabulator.java
HEdingfield: Hmm, I might be wrong but looks like it's still O(n) where n is size of castVoteRecords. Maybe a good idea to create an issue to add this option as something we can specify in the config file in the future? I'll leave it up to you.

src/com/rcv/CastVoteRecord.java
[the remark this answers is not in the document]
moldover: This is getting divisive.  I'd propose allow direct access for more convenient coding but once we have accessors use them exclusively.  And FWIW, I ALWAYS wear a condom while masturbating.

src/com/rcv/Tabulator.java
```diff
@@ -35,6 +35,9 @@
   // e.g. roundTallies[1] contains a map of all candidate ID -> votes for each candidate in round 1
   // this structure is computed over the course of tabulation
   private final Map<Integer, Map<String, BigDecimal>> roundTallies = new HashMap<>();
+  // precinctRoundTallies is like roundTallies, but bucketed by precinct
+  private final Map<String, Map<Integer, Map<String, BigDecimal>>> precinctRoundTallies =
```
moldover: Looks good.  Maybe just clarify the comment: precinctRoundTallies is a map of precinct -> roundTallies for that precinct.

moldover:
Nice work.

src/com/rcv/Tabulator.java
moldover: I'll ask if we need a global setting for tabulate by precincts (my guess is yes).  Performance isn't an issue (I doubt it will ever be) but more for clarity in the code.

tarheel: OK, I opened an issue for this. Seems reasonable to have a setting for it.

src/com/rcv/CastVoteRecord.java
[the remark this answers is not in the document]
tarheel: I added the getter here because we need to access this value from Tabulator. I guess my stance is that when we have a getter for an instance variable, it should just be returning the value and not doing anything else anyway. I'm happy to revisit this and change this class and other classes that have a similar issue if we agree on a different approach, but for now will just leave as is for consistency.

PR #64, opened 2018-04-28
generate precinct spreadsheets

tarheel:
This completes the work to add support for per-precinct results summaries. I'm sure we'll need to tweak the details based on feedback, but the key functionality seems to be working.

I also resolved this issue: https://github.com/BrightSpots/rcv/issues/63

Oh, and since I was updating the config documentation anyway, I sorted it alphabetically.

### DIFF
--- a/docs/config_file_documentation.txt
+++ b/docs/config_file_documentation.txt
@@ -1,4 +1,4 @@
-This document lists the parameters that can be included in a config file. Required parameters are denoted with an asterisk.
+This document lists the parameters that can be included in a config file. Required parameters are denoted with an asterisk (*).
 - *auditOutput: the filename for the audit file that the program will create. Example: portland_2015_mayor_audit_log.txt
 - *candidates: a list of the candidates who appear in the cast vote records. Each candidate must have a name. If the cast vote records use candidate codes, each candidate must also have a code.
 - contestName: the name of the election. Example: Portland Mayoral Race
@@ -10,7 +10,7 @@ This document lists the parameters that can be included in a config file. Requir
 - date: the date of the election. Example: 2015-11-03
 - jurisdiction: the jurisdiction in which this election is being contested. Example: Portland, ME
 - office: the office being contested. Example: Mayor
-- precinctOutputDirectory: the path to where the precinct spreadsheets should be saved. Example: portland_2015_precinct_output (defaults to the current directory)
+- outputDirectory: the directory where all output files should be saved. Example: portland_2015_output (defaults to the current directory)
 - *rules: a set of options that determine exactly how the program should implement the instant runoff.
   - *batchElimination: whether batch elimination should be used. Valid options: true, false.
   - decimalPlacesForVoteArithmetic: how many places after the decimal to use in rounding for fractional vote transfer arithmetic in multi-winner elections. (defaults to 4)
--- a/docs/config_file_documentation.txt
+++ b/docs/config_file_documentation.txt
@@ -1,32 +1,29 @@
 This document lists the parameters that can be included in a config file. Required parameters are denoted with an asterisk.
 - *auditOutput: the filename for the audit file that the program will create. Example: portland_2015_mayor_audit_log.txt
-- *visualizerOutput: the filename for the Excel spreadsheet with a summary of the results that the program will create. Example: portland_2015_mayor_visualizer.xlsx
-- contestName: the name of the election. Example: Portland Mayoral Race
-- jurisdiction: the jurisdiction in which this election is being contested. Example: Portland, ME
-- office: the office being contested. Example: Mayor
-- date: the date of the election. Example: 2015-11-03
-- *maxRankingsAllowed: the maximum number of candidates that a ballot is allowed to rank. Example: 15
-- *rules: a set of options that determine exactly how the program should implement the instant runoff.
-  - description: the name of the set of the rules. Example: Maine Rules
-  - *batchElimination: whether batch elimination should be used. Valid options: true, false.
-  - *overvoteRule: how the program should handle an overvote when it encounters one. Valid options: alwaysSkipToNextRank, exhaustImmediately, exhaustIfAnyContinuing, ignoreIfAnyContinuing, exhaustIfMultipleContinuing, ignoreIfMultipleContinuing.
-  - *maxSkippedRanksAllowed: the maximum number of skipped ranks (i.e. undervotes) on a ballot before the ballot should be considered exhausted. Example: 1
-  - *tiebreakMode: how the program should decide which candidate to eliminate when multiple candidates are tied for last place among continuing candidates in a given round. Valid options: random, interactive, previousRoundCountsThenRandom, previousRoundCountsThenInteractive.
-  - minimumVoteThreshold: the minimum number of votes a candidate must receive in the first round to avoid automatic elimination. Example: 150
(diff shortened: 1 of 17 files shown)


test/config_minneapolis_2013.json
```diff
@@ -1,7 +1,11 @@
 {
-  "auditOutput":"Minneapolis2013_audit_log.txt",
-  "visualizerOutput":"Minneapolis2013_visualizer.xlsx",
+  "auditOutput":"minneapolis_2013_audit_log.txt",
+  "visualizerOutput":"minneapolis_2013_visualizer.xlsx",
   "contestName":"2013 Minneapolis Mayor",
+  "jurisdiction":"Minneapolis",
+  "office":"Mayor",
+  "tabulateByPrecinct":true,
+  "precinctOutputDirectory":"test_output/minneapolis_2013_precincts",
```
HEdingfield: See above comment on `precinctOutputDirectory`. If we can fit it into this branch before merging, we should also add the `outputDirectory` field to each existing config file.

docs/config_file_documentation.txt
```diff
@@ -1,32 +1,29 @@
...
+  - *filePath: the location of the file. Example: test/2015-portland-mayor-cvr.xlsx
+  - *firstVoteColumnIndex: the index of the column that contains each ballot’s top-ranked candidate. Example: 3
+  - precinctColumnIndex: the index of the column that contains each ballot's precinct name. Example: 2
+  - *provider: the voting machine format of the file. Example: ES&S
+- date: the date of the election. Example: 2015-11-03
 - jurisdiction: the jurisdiction in which this election is being contested. Example: Portland, ME
 - office: the office being contested. Example: Mayor
-- date: the date of the election. Example: 2015-11-03
-- *maxRankingsAllowed: the maximum number of candidates that a ballot is allowed to rank. Example: 15
+- precinctOutputDirectory: the path to where the precinct spreadsheets should be saved. Example: portland_2015_precinct_output (defaults to the current directory)
```
HEdingfield: I think that we should have a general `outputDirectory` field (i.e. where the log and visualizer files go) in each config file instead of a special one for precincts only. Been wanting to make this happen for a little while now. Think we can work it into this branch before the merge?

docs/config_file_documentation.txt
```diff
@@ -1,32 +1,29 @@
 This document lists the parameters that can be included in a config file. Required parameters are denoted with an asterisk.
```
HEdingfield: Maybe better to say "* = required parameter" here at the top or at the bottom. I personally did a ctrl-f for "*" and didn't find this note.

src/com/rcv/ElectionConfig.java
```diff
@@ -237,6 +237,22 @@ public String electionDate() {
     return rawConfig.date;
   }
 
+  // function: tabulateByPrecinct
+  // purpose: getter for tabulateByPrecinct
+  // returns: true iff we should tabulate by precinct
+  public boolean tabulateByPrecinct() {
```
HEdingfield: Should these two getters be called get*() instead?

src/com/rcv/ElectionConfig.java
```diff
@@ -237,6 +237,22 @@ public String electionDate() {
     return rawConfig.date;
   }
 
+  // function: tabulateByPrecinct
+  // purpose: getter for tabulateByPrecinct
+  // returns: true iff we should tabulate by precinct
```
HEdingfield: "iff" meaning "if and only if" or is this a typo? Probably better to spell it out if it's the former.

src/com/rcv/ResultsWriter.java
```diff
@@ -103,15 +95,61 @@ public ResultsWriter setWinnerToRound(Map<String, Integer> winnerToRound) {
...
+
+  // function: generatePrecinctSummarySpreadsheet
+  // purpose: creates a summary spreadsheet for the votes in a particular precinct
+  // param: roundTallies is map from precinct to the round-by-round vote count in the precinct
+  void generatePrecinctSummarySpreadsheets(
+      Map<String, Map<Integer, Map<String, BigDecimal>>> precinctRoundTallies
+  ) {
+    // location for output
+    String outputDir = config.precinctOutputDirectory();
+    if (outputDir != null) {
```
HEdingfield: What if outputDir == null?

src/com/rcv/ResultsWriter.java
```diff
@@ -103,15 +95,61 @@ public ResultsWriter setWinnerToRound(Map<String, Integer> winnerToRound) {
...
+      // dir is the directory to write these files
+      File dir = new File(config.precinctOutputDirectory());
+      if (!dir.mkdirs()) {
+        Logger.log("Failed to create precinct output dir: " + outputDir);
```
HEdingfield: Should we throw an exception here?

tarheel:
OK, I think I've addressed @HEdingfield's comments (and then some).

src/com/rcv/Main.java
```diff
@@ -47,33 +49,59 @@ public static void main(String[] args) {
...
-      rawConfig = JsonParser.parseObjectFromFile(configPath, RawElectionConfig.class);
-      // create and validate new ElectionConfig
-      if (rawConfig != null) {
-        config = new ElectionConfig(rawConfig);
-        // set up log output
-        Logger.setup(config.auditOutput());
+
+    // hitError tracks whether anything went wrong during loading
+    boolean hitError = false;
+
```
moldover: "error" would be clearer

src/com/rcv/RawElectionConfig.java
```diff
@@ -34,6 +37,8 @@
   public List<Candidate> candidates;
   // list of all cast vote record source files
   public List<CVRSource> cvrFileSources;
+  // should we report round-by-round results by precinct also?
+  public Boolean tabulateByPrecinct;
```
moldover: So.... to use native boolean or not?  I'm fine with either but would like to be consistent.

docs/config_file_documentation.txt
moldover: This is fine, and I'm thinking maybe we should go one step further and stop letting the user specify output file names entirely to simplify things.  They just enter the output folder, and we generate names for the outputs like 
<guid>_visualizer.xls,  <guid>_audit.txt and so on (new guid for every tabulation)

src/com/rcv/ElectionConfig.java
moldover: This seems a bit non-standard?  Not a big deal to me, but it does make me wonder what is being communicated here beyond a regular "if"?

src/com/rcv/ResultsWriter.java
moldover: we should validate that in the config settings, then this code can assume a valid output.

src/com/rcv/ResultsWriter.java
moldover: I'd propose validate this in config validate as well.

src/com/rcv/RawElectionConfig.java
tarheel: Oops. I had assumed that we had to handle the possibility of a null for this, but it turns out that's not the case: if you omit a boolean field from the config file, the parser will just default it to false.

src/com/rcv/Main.java
tarheel: I disagree -- if it were called `error`, I would expect it to be a string or something with info about the error. I changed it to `encounteredError` to make it even clearer that it's a boolean.

PR #68, opened 2018-05-20
improved tracking and reporting of fractional transfers in multi-winner

tarheel:
I changed a few things related to how we track and report multi-winner results:
1. Each CVR now keeps a record of what fraction of its vote is going to whom.
2. The audit file indicates when a candidate gets only a fraction of a CVR's vote.
3. The results spreadsheet shows that a winner maintains their winning vote total in subsequent rounds.

The way I did (3) isn't efficient -- in each round, I actually count up the votes for each of the past winners -- but it doesn't change the big-O running time, and it's kind of reassuring as a sanity check. But we could easily optimize this by just storing those counts somewhere and inserting them directly into the tallies each round instead of actually going through the CVRs each time.

### DIFF
--- a/src/com/rcv/Tabulator.java
+++ b/src/com/rcv/Tabulator.java
@@ -146,14 +146,15 @@ class Tabulator {
       BigDecimal winningThresholdThisRound = getThreshold(currentRoundCandidateToTally);
 
       // currentRoundTallyToCandidates is a sorted map from tally to candidate(s) with that tally.
-      SortedMap<BigDecimal, LinkedList<String>> currentRoundTallyToCandidates = buildTallyToCandidates(
-          currentRoundCandidateToTally,
-          currentRoundCandidateToTally.keySet(),
-          true
-      );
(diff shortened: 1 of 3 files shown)


src/com/rcv/Tabulator.java
```diff
@@ -207,6 +206,60 @@ void tabulate() {
...
+
+      // TODO: avoid recomputing these every round (since they should be static for a given
+      // candidate once that candidate has won)?
+      updateWinnerTallies();
+    }
+  }
+
+  // function: updateWinnerTallies
+  // purpose: update the tally for the just-completed round to reflect the tallies for candidates
+  // who won in a past round
```
moldover: Looks good.  Could you please add a bit more comments on why we need to update past winners?  I know the context so I get it, but to anyone else trying to navigate this code it would be helpful.   Also (annoyingly) the pastWinners local variable below needs an inline comment too.  Thanks.

tarheel: Will do.

Issue #75, opened 2018-06-02
add support for permutation-based tie-breaking


tarheel:
This is somewhat connected to #60. If we allow the operator to specify the tie-breaking permutation in the config file, they need to list out the candidates, of course.

For simplicity, my recommendation is to punt on #60 for now and rely on the candidate ordering in the config file to define the permutation for tie-breaking when this tie-breaking mode is enabled.

moldover:
Sounds right.  #60 is also something we should run past Caleb and George before doing, I suspect there may be good reasons not to do it at all (like, it could incorrectly treat undeclared write-ins like declared candidates for some cvrs)

CalebKleppner:
I think this is a great example of you guys coming up with a very efficient way to interpret something (permutation based tie-breaking) that most vendors aren't even thinking about.  Nice job.

PR #76, opened 2018-06-09
add permutation-based tie-breaking

tarheel:
This implements #75.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,8 @@ test/2013-minneapolis-mayor-tabulation-results.xlsx
 test/~*.xlsx
 test/Portland-Tabulation-Results.xlsx
 *_audit_log.txt
+*_audit.log
+rcv.log
 *_visualizer.xlsx
+*_summary.xlsx
 test_output
--- a/src/com/rcv/ElectionConfig.java
+++ b/src/com/rcv/ElectionConfig.java
@@ -10,9 +10,13 @@
 
 package com.rcv;
 
+import com.rcv.Tabulator.TieBreakMode;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -22,6 +26,8 @@ class ElectionConfig {
   final RawElectionConfig rawConfig;
   // mapping from candidate code to full name
   private Map<String, String> candidateCodeToNameMap;
+  // this is used if we have a permutation-based tie-break mode
+  private ArrayList<String> candidatePermutation = new ArrayList<>();
   // minimum vote threshold if one is specified
   private BigDecimal minimumVoteThreshold;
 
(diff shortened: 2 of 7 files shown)


src/com/rcv/TieBreak.java
```diff
@@ -87,41 +93,61 @@ String nonLosingCandidateDescription() {
...
+    // start from end of list and look for a matching candidate
+    for (int i = candidatePermutation.size() - 1; i >= 0; i--) {
+      // the current candidate we're considering from the permutation
+      String candidate = candidatePermutation.get(i);
+      if (tiedCandidatesSet.contains(candidate)) {
+        selectedCandidate = candidate;
+        break;
+      }
+    }
+    explanation = "The losing candidate appeared last in the tie-breaking permutation list.";
```
moldover: losing candidate appeared "later"

.gitignore
```diff
@@ -7,5 +7,8 @@ test/2013-minneapolis-mayor-tabulation-results.xlsx
 test/~*.xlsx
 test/Portland-Tabulation-Results.xlsx
 *_audit_log.txt
+*_audit.log
+rcv.log
 *_visualizer.xlsx
```
moldover: nice

src/com/rcv/ElectionConfig.java
```diff
@@ -382,6 +397,13 @@ String getNameForCandidateID(String candidateID) {
...
 
+  // function: getCandidatePermutation
+  // purpose: getter for ordered list of candidates for tie-breaking
+  // returns: ordered list of candidates
+  ArrayList<String> getCandidatePermutation() {
+    return candidatePermutation;
+  }
+
   // function: processCandidateData
   // purpose: builds map of candidate ID to candidate name
```
moldover: add comment about also generating tiebreak permutation

src/com/rcv/TieBreak.java
moldover: This looks solid, I only wonder what happens if UWI is involved in the tiebreak?

moldover: I was thinking we could add optional "tieBreakOrder" field into the Candidate object of RawElectionConfig.  If user specifies USE_PERMUTATION_IN_CONFIG then validation code can make sure user ordering is valid before proceeding.

tarheel: > losing candidate appeared "later"

I changed it to "latest." (Remember that there can be more than 2 people tied.)

> This looks solid, I only wonder what happens if UWI is involved in the tiebreak?

The tabulator drops UWIs before it does anything else, so they can't be involved in a tie-break scenario.

> I was thinking we could add optional "tieBreakOrder" field into the Candidate object of RawElectionConfig. If user specifies USE_PERMUTATION_IN_CONFIG then validation code can make sure user ordering is valid before proceeding.

I thought about that, but decided to start with this approach (just use the actual ordering in the file) because it's simpler. I'm fine with changing to do it the way you suggested later if the group agrees that it's better.

moldover: Yeah, your approach is simpler so that's fine.

Issue #79, opened 2018-07-15
read CVR IDs from input files


tarheel:
Right now we're ignoring them and just creating our own IDs.

PR #80, opened 2018-07-19
read IDs from CVR files

tarheel:
This addresses #79.

I also cleaned up a couple related minor things:
- We were showing the source file name twice in each audit line.
- We were interpreting numbers read from the input file as doubles. For now, I think it's safe to assume that they're integers. This might change in the future.

### DIFF
--- a/src/com/rcv/CastVoteRecord.java
+++ b/src/com/rcv/CastVoteRecord.java
@@ -23,10 +23,10 @@ import javafx.util.Pair;
 
 class CastVoteRecord {
 
-  // name of the vendor, this becomes part of the audit output but is not used in tabulation
-  private final String sourceName;
-  // unique identifier for this cast vote record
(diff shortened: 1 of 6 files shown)


src/com/rcv/CastVoteRecord.java
```diff
@@ -45,19 +45,19 @@
 
   // function: CastVoteRecord
   // purpose: create a new CVR object
-  // param: source what vendor created the CVR file from which this CVR was parsed
-  // param: ballotID unique ID of this ballot
+  // param: computedID is our computed unique ID for this CVR
+  // param: suppliedID is the (ostensibly unique) ID from the input data
```
moldover: +1 for using ostensibly correctly

tarheel: Ostensibly.

PR #81, opened 2018-07-19
better config validation

tarheel:
I've been meaning to do this forever (see #69). Now, if you try to load config_invalid.json, you get output like this:

```
$ java -jar out/artifacts/rcv_jar/rcv.jar test/config_invalid.json
Thu Jul 19 16:32:28 CEST 2018 SEVERE: Invalid config: Config must contain at least one declared candidate.
Thu Jul 19 16:32:28 CEST 2018 SEVERE: Invalid config: When overvoteLabel is supplied, overvoteRule must be either exhaustImmediately or alwaysSkipToNextRank.
Thu Jul 19 16:32:28 CEST 2018 SEVERE: Invalid config: Invalid tie-break mode.
Thu Jul 19 16:32:28 CEST 2018 SEVERE: Invalid config: maxSkippedRanksAllowed can't be negative.
Thu Jul 19 16:32:28 CEST 2018 SEVERE: Invalid config: maxRankingsAllowed must be positive.
Thu Jul 19 16:32:28 CEST 2018 SEVERE: Invalid config: continueUntilTwoCandidatesRemain can't be true in a multi-winner election.
Thu Jul 19 16:32:28 CEST 2018 SEVERE: Invalid config: batchElimination can't be true in a multi-winner election.
Thu Jul 19 16:32:28 CEST 2018 SEVERE: Invalid config: decimalPlacesForVoteArithmetic must be between 0 and 20.
Thu Jul 19 16:32:28 CEST 2018 SEVERE: Invalid config: Invalid multiSeatTransferRule.
Thu Jul 19 16:32:28 CEST 2018 SEVERE: Aborting because config is invalid.
```

### DIFF
--- a/src/com/rcv/CVRReader.java
+++ b/src/com/rcv/CVRReader.java
@@ -11,29 +11,198 @@ package com.rcv;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import javafx.util.Pair;
 import org.apache.poi.ss.usermodel.Cell;
 import org.apache.poi.ss.usermodel.CellType;
+import org.apache.poi.ss.usermodel.Row;
 import org.apache.poi.ss.usermodel.Sheet;
 import org.apache.poi.ss.usermodel.Workbook;
 import org.apache.poi.xssf.usermodel.XSSFWorkbook;
 
 class CVRReader {
 
-  // container for all CastVoteRecords parsed from the input file
-  final List<CastVoteRecord> castVoteRecords = new ArrayList<>();
+  // config for the election
+  private ElectionConfig config;
+  // path of the source file
+  private String excelFilePath;
+  // column index of first ranking
+  private int firstVoteColumnIndex;
+  // column index of CVR ID (if present)
+  private Integer idColumnIndex;
+  // column index of precinct name (if present)
+  private Integer precinctColumnIndex;
 
+  // function: CVRReader
+  // param: config an ElectionConfig object specifying rules for interpreting CVR file data
+  // param: excelFilePath path to location of input cast vote record file
+  // param: firstVoteColumnIndex the 0-based index where rankings begin for this ballot style
+  // param: idColumnIndex the column containing the CVR ID (possibly null)
+  // param: precinctColumnIndex the column containing precinct names (possibly null)
+  CVRReader(
+      ElectionConfig config,
+      String excelFilePath,
+      int firstVoteColumnIndex,
+      Integer idColumnIndex,
+      Integer precinctColumnIndex
+  ) {
+    this.config = config;
+    this.excelFilePath = excelFilePath;
+    this.firstVoteColumnIndex = firstVoteColumnIndex;
+    this.idColumnIndex = idColumnIndex;
+    this.precinctColumnIndex = precinctColumnIndex;
+  }
+
+  // function: parseCVRFile
+  // purpose: parse the given file path into a List of CastVoteRecords suitable for tabulation
+  // returns: list of parsed CVRs
+  List<CastVoteRecord> parseCVRFile() {
+    // contestSheet contains all the CVR data we will be parsing
+    Sheet contestSheet = getFirstSheet(excelFilePath);
+    // container for all CastVoteRecords parsed from the input file
+    List<CastVoteRecord> castVoteRecords = new LinkedList<>();
+
+    if (contestSheet != null) {
+      // validate header
+      // Row iterator is used to iterate through a row of data from the sheet object
+      Iterator<Row> iterator = contestSheet.iterator();
+      // headerRow contains the first row
+      Row headerRow = iterator.next();
+      // require at least one non-header row
+      if (headerRow == null || contestSheet.getLastRowNum() < 2) {
+        Logger.severe(
+            "Invalid CVR source file %s: not enough rows (%d)",
+            this.excelFilePath,
+            contestSheet.getLastRowNum()
+        );
+      }
+
+      // cvrFileName for generating cvrIDs
+      String cvrFileName = new File(excelFilePath).getName();
+      // cvrIndex for generating cvrIDs
+      int cvrIndex = 1;
+
+      // Iterate through all rows and create a CastVoteRecord for each row
+      while (iterator.hasNext()) {
+        // cvr is the object parsed from the row
+        CastVoteRecord cvr = parseRow(iterator.next(), cvrFileName, cvrIndex++);
+        castVoteRecords.add(cvr);
+      }
+    }
+    return castVoteRecords;
+  }
+
+  // function: parseRow
+  // purpose: parse a single row into a CastVoteRecord
+  // returns: a CastVoteRecord object
+  private CastVoteRecord parseRow(
+      Row castVoteRecordRow,
+      String cvrFileName,
+      int cvrIndex
+  ) {
+    // row object is used to iterate CVR file data for this CVR
+    // computed unique ID for this CVR
+    String computedCastVoteRecordID = String.format("%s(%d)", cvrFileName, cvrIndex);
+    // supplied (by input file) unique ID for this CVR
+    String suppliedCastVoteRecordID = null;
+    // list of rankings read from this row
+    LinkedList<Pair<Integer, String>> rankings = new LinkedList<>();
+    // list of raw strings read from this row, for the audit log
+    LinkedList<String> fullCVRData = new LinkedList<>();
+    // the precinct for this ballot
+    String precinct = null;
+
+    // Iterate over all expected cells in this row storing cvrData and rankings as we go.
+    // cellIndex ranges from 0 to the last expected rank column index
+    for (
+        int cellIndex = 0;
+        cellIndex < firstVoteColumnIndex + config.getMaxRankingsAllowed();
+        cellIndex++
+    ) {
+      // cell object contains data for the current cell
+      Cell cvrDataCell = castVoteRecordRow.getCell(cellIndex);
+      String cellString = getStringFromCell(cvrDataCell);
+
+      if (cellString == null) {
+        fullCVRData.add("empty cell");
+      } else {
+        fullCVRData.add(cellString);
+      }
+
+      if (cellString != null) {
+        if (precinctColumnIndex != null && cellIndex == precinctColumnIndex) {
+          precinct = cellString;
+        } else if (idColumnIndex != null && cellIndex == idColumnIndex) {
+          suppliedCastVoteRecordID = cellString;
+        }
+      }
+
+      // if we haven't reached a vote cell continue to the next cell
+      if (cellIndex < firstVoteColumnIndex) {
+        continue;
+      }
+
+      // rank for this cell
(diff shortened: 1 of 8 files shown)


src/com/rcv/ElectionConfig.java
```diff
@@ -408,24 +424,27 @@ String getNameForCandidateID(String candidateID) {
...
-    // candidate is used to index through all candidates for this election
-    for (RawElectionConfig.Candidate candidate : rawConfig.candidates) {
-      if (candidate.code != null) {
-        candidateCodeToNameMap.put(candidate.code, candidate.name);
-        candidatePermutation.add(candidate.code);
-      } else {
-        candidateCodeToNameMap.put(candidate.name, candidate.name);
-        candidatePermutation.add(candidate.name);
+
+    if (rawConfig.candidates != null) {
```
moldover: we validate that candidates may not be null, so I prefer to remove this check or replace with an assertion

src/com/rcv/Main.java
```diff
@@ -45,53 +45,63 @@ public static void main(String[] args) {
...
         );
+        encounteredFileError = true;
       }
     }
 
-    // TODO: confiog.validate() should log specific errors
-    if (encounteredError || !config.validate()) {
-      Logger.severe("There was a problem loading or validating the election configuration.");
-      Logger.severe("Please see the README.txt for details.");
+    if (config != null) {
```
moldover: same thing here

src/com/rcv/Main.java
```diff
@@ -154,9 +168,13 @@ static String executeTabulation(ElectionConfig config) {
...
   // returns: list of all CastVoteRecord objects parsed from CVR files
   private static List<CastVoteRecord> parseCastVoteRecords(ElectionConfig config) throws Exception {
+    if (config.rawConfig.cvrFileSources == null || config.rawConfig.cvrFileSources.isEmpty()) {
+      Logger.severe("Config doesn't contain any CVR input files.");
+    }
+
     // castVoteRecords will contain all cast vote records parsed by the reader
     List<CastVoteRecord> castVoteRecords = new ArrayList<>();
-    // at each iteration of the following loop we add records from another source file
+    // At each iteration of the following loop, we add records from another source file.
```
moldover: bit of a deviation in style here, we don't seem to capitalize or punctuate inline comments anywhere else.

src/com/rcv/ElectionConfig.java
tarheel: If you leave the "candidates" field out of the config file entirely, rawConfig.candidates will be null -- so we do need to check for this. Or did I misunderstand your comment?

src/com/rcv/Main.java
tarheel: The way this method is written, we do need this check... but I'll restructure the logic in the method slightly so it's not necessary.

src/com/rcv/Main.java
tarheel: True. I changed this one because I was having a little trouble parsing the comment. I'm OK with being lax on such things in general, but I think it's fine to occasionally make comments more clear when one of us is sufficiently motivated.

tarheel:
@moldover: I just added a second commit that has a bunch of new validation for source files. Could you take another look? In retrospect, this might have been better as a separate PR, but oh well.

src/com/rcv/Main.java
```diff
@@ -150,9 +150,8 @@ static String executeTabulation(ElectionConfig config) {
         } else {
           Logger.severe("No cast vote records found.");
         }
-      } catch (Exception exception) {
-        encounteredError = true;
-        Logger.severe("ERROR during tabulation: %s", exception.toString());
+      } else {
+        Logger.severe("Skipping tabulation due to source file errors.");
```
HEdingfield: Why move to this from the previous try / catch logic? The former seemed to be able to handle a broader array of situations and provide a more informative error message.

src/com/rcv/Main.java
```diff
@@ -164,31 +163,84 @@ static String executeTabulation(ElectionConfig config) {
...
+              config,
+              source.filePath,
+              source.firstVoteColumnIndex,
+              source.idColumnIndex,
+              source.precinctColumnIndex
+          );
+          // the CVRs parsed from this source
+          List<CastVoteRecord> cvrs = reader.parseCVRFile();
+          if (cvrs.isEmpty()) {
+            Logger.severe("Source file contains no CVRS: %s", source.filePath);
```
HEdingfield: "CVRS" => "CVRs"

src/com/rcv/Main.java
```diff
@@ -164,31 +163,84 @@ static String executeTabulation(ElectionConfig config) {
...
+          // add records to the master list
+          castVoteRecords.addAll(cvrs);
+        }
+
+        if (encounteredProblemForThisSource) {
+          numInvalidSources++;
+        }
+      }
+      if (numInvalidSources > 0) {
+        Logger.severe("Encountered %d invalid source(s)", numInvalidSources);
```
HEdingfield: Should we add something here about the process terminating entirely as a result of invalid sources?

src/com/rcv/Main.java
```diff
@@ -164,31 +163,84 @@ static String executeTabulation(ElectionConfig config) {
...
-      reader.parseCVRFile(
-          source.filePath,
-          source.firstVoteColumnIndex,
-          source.idColumnIndex,
-          source.precinctColumnIndex,
-          config
-      );
-      // add records to the master list
-      castVoteRecords.addAll(reader.castVoteRecords);
+    if (numInvalidSources > 0) {
```
HEdingfield: Merge this into the above check for `numInvalidSources > 0`.

src/com/rcv/Main.java
```diff
@@ -164,31 +163,84 @@ static String executeTabulation(ElectionConfig config) {
   // function: parseCastVoteRecords
   // purpose: parse CVR files referenced in the ElectionConfig object into a list of CastVoteRecords
   // param: config object containing CVR file paths to parse
-  // returns: list of all CastVoteRecord objects parsed from CVR files
-  private static List<CastVoteRecord> parseCastVoteRecords(ElectionConfig config) throws Exception {
+  // returns: list of all CastVoteRecord objects parsed from CVR files (or null if there's an error)
+  private static List<CastVoteRecord> parseCastVoteRecords(ElectionConfig config) {
```
HEdingfield: Are there not other exceptions that could happen that aren't accounted for below? Everything works fine when we remove the `throw`?

src/com/rcv/ElectionConfig.java
```diff
@@ -122,53 +121,70 @@
...
+      if (willContinueUntilTwoCandidatesRemain()) {
+        errors.add("continueUntilTwoCandidatesRemain can't be true in a multi-winner election.");
+      }
+
+      if (isBatchEliminationEnabled()) {
+        errors.add("batchElimination can't be true in a multi-winner election.");
       }
+
+      if (getDecimalPlacesForVoteArithmetic() < 0 || getDecimalPlacesForVoteArithmetic() > 20) {
+        errors.add("decimalPlacesForVoteArithmetic must be between 0 and 20.");
```
HEdingfield: "... 0 and 20, inclusive."

src/com/rcv/Main.java
```diff
@@ -45,53 +45,61 @@ public static void main(String[] args) {
...
-  // returns: the new ElectionConfig object or null if there was a problem
+  // returns: the new ElectionConfig object, or null if there was a problem
   static ElectionConfig loadElectionConfig(String configPath) {
     // config: the new object
     ElectionConfig config = null;
 
-    // encounteredError tracks whether anything went wrong during loading
-    boolean encounteredError = false;
+    // tracks whether we encountered any critical file system errors
+    boolean encounteredFileError = false;
```
HEdingfield: This variable is only set to true one time below, and it could just as easily be eliminated entirely and have `config = null;` replace the line where it's set to true. I'd recommend either doing that or else inserting `encounteredFileError = true;` in the other place where `config = null;` occurs, and also in the check for `rawConfig == null`. I think it's simpler just to get rid of it though.

src/com/rcv/CVRReader.java
```diff
@@ -23,162 +23,197 @@
...
-  // purpose: parse the given file path into a CastVoteRecordList suitable for tabulation
-  // Note: this is specific for the Maine example file we were provided
+  // config for the election
+  private ElectionConfig config;
+  // path of the source file
+  private String excelFilePath;
+  // column index of first ranking
+  private int firstVoteColumnIndex;
+  // column index of CVR ID (if present)
+  private Integer idColumnIndex;
```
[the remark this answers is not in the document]
tarheel: I use the wrapper class instead of the primitive when I want the variable to be nullable. Since Java does auto-unboxing of these wrapper objects, it's safe to treat them as if they were primitives when you know that they're not null.

src/com/rcv/CVRReader.java
```diff
@@ -23,162 +23,197 @@
...
+  List<CastVoteRecord> parseCVRFile() {
     // contestSheet contains all the CVR data we will be parsing
     Sheet contestSheet = getFirstSheet(excelFilePath);
+    // container for all CastVoteRecords parsed from the input file
+    List<CastVoteRecord> castVoteRecords = new LinkedList<>();
+
+    if (contestSheet != null) {
+      // validate header
+      // Row iterator is used to iterate through a row of data from the sheet object
+      Iterator<org.apache.poi.ss.usermodel.Row> iterator = contestSheet.iterator();
```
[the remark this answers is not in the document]
tarheel: good point

src/com/rcv/Main.java
tarheel: I actually need it, because when I encounter a file error, I don't want to set `config = null` immediately; first I want to call `config.getValidationErrors()`.

src/com/rcv/Main.java
tarheel: There's probably other stuff that can go wrong in the parsing. I think I'll test that more in a follow-up.

src/com/rcv/Main.java
tarheel: When this happens, the next log line will say, "Skipping tabulation due to source file errors."

src/com/rcv/Main.java
tarheel: The try/catch logic was just catching everything indiscriminately, which means at best we'd show a generic message and the raw exception -- which isn't a great user experience. Perhaps we'll need to add back a catch-all handler eventually, but I'd like to try to cover as many specific error conditions as possible in the code so that we can optimize the error messages that we show.

src/com/rcv/Main.java
tarheel: oops

src/com/rcv/Main.java
HEdingfield: Agreed, but at the moment if anything falls outside of the conditions you're checking for, the whole thing just dies, which is an even worse experience, right? I'd recommend adding back the sub-optimal UX of seeing the raw exception just in case something does happen to fall outside.

tarheel: Hmm, actually, I'm having trouble finding any way to make it die now. Let's see how it goes...

src/com/rcv/ElectionConfig.java
moldover: Damn, I forgot that was an option.  In that case never mind@

PR #90, opened 2018-07-29
unrecognized candidate in CVR source should cause tabulation to abort

tarheel:
Right now, if you have an unrecognized candidate and you don't define undeclaredWriteInLabel, the program actually crashes with a NullPointerException during tabulation.


src/com/rcv/CVRReader.java
```diff
@@ -220,4 +236,31 @@ private String getStringFromCell(Cell cvrDataCell) {
 
     return cellString;
   }
+
+  static class UnrecognizedCandidateException extends Exception {
```
HEdingfield: Do we need to add the function / purpose / param / returns comments for these two exception classes?

src/com/rcv/Main.java
```diff
@@ -208,13 +211,24 @@ static String executeTabulation(ElectionConfig config) {
...
+          try {
+            List<CastVoteRecord> cvrs = reader.parseCVRFile();
+            if (cvrs.isEmpty()) {
+              Logger.severe("Source file contains no CVRs: %s", source.filePath);
+              encounteredProblemForThisSource = true;
+            }
+            // add records to the master list
+            castVoteRecords.addAll(cvrs);
+          } catch (SourceWithUnrecognizedCandidatesException e) {
+            Logger.severe("Source file contains unrecognized candidates: %s", source.filePath);
```
HEdingfield: "... candidate(s): "

src/com/rcv/CVRReader.java
```diff
@@ -179,11 +196,10 @@ private CastVoteRecord parseRow(Row castVoteRecordRow, String cvrFileName, int c
...
         } else if (candidate.equals(config.getOvervoteLabel())) {
           candidate = Tabulator.explicitOvervoteLabel;
-        } else if (!config.getCandidateCodeList().contains(candidate)) {
-          if (!candidate.equals(config.getUndeclaredWriteInLabel())) {
-            Logger.warn("no match for candidate: %s", candidate);
-          }
-          candidate = config.getUndeclaredWriteInLabel();
+        } else if (!config.getCandidateCodeList().contains(candidate)
+            && !candidate.equals(config.getUndeclaredWriteInLabel())) {
+          Logger.severe("No match for candidate: %s", candidate);
```
HEdingfield: In practice, this actually gets really noisy, and is unnecessary in light of the super cool summary error with counts you print out at the end. Recommend removing.

src/com/rcv/CVRReader.java
```diff
@@ -103,21 +105,36 @@ private static Sheet getFirstSheet(String excelFilePath) {
...
       while (iterator.hasNext()) {
         // cvr is the object parsed from the row
-        CastVoteRecord cvr = parseRow(iterator.next(), cvrFileName, cvrIndex++);
-        castVoteRecords.add(cvr);
+        try {
+          CastVoteRecord cvr = parseRow(iterator.next(), cvrFileName, cvrIndex++);
+          castVoteRecords.add(cvr);
+        } catch (UnrecognizedCandidateException e) {
+          // increment the count
+          unrecognizedCandidateCounts.merge(e.getName(), 1, Integer::sum);
```
HEdingfield: Smooth! Can we add a test config to trigger this error? I was testing it by slightly modifying an existing config (i.e. just deleting out a few candidates).

src/com/rcv/CVRReader.java
```diff
@@ -103,21 +105,36 @@ private static Sheet getFirstSheet(String excelFilePath) {
...
+        }
+      }
+
+      if (unrecognizedCandidateCounts.size() > 0) {
```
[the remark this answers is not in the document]
HEdingfield: I see it too, that's a bummer. We can just remember to ignore this one.

src/com/rcv/CVRReader.java
```diff
@@ -220,4 +236,31 @@ private String getStringFromCell(Cell cvrDataCell) {
 
     return cellString;
   }
+
+  static class UnrecognizedCandidateException extends Exception {
+
+    private String name;
```
HEdingfield: Code analysis points out that this can be `final`.

src/com/rcv/CVRReader.java
```diff
@@ -220,4 +236,31 @@ private String getStringFromCell(Cell cvrDataCell) {
...
+    }
+
+    String getName() {
+      return name;
+    }
+  }
+
+  static class SourceWithUnrecognizedCandidatesException extends Exception {
+
+    private Map<String, Integer> candidateCounts;
```
HEdingfield: Code analysis points out that this can be `final`.

src/com/rcv/CVRReader.java
```diff
@@ -103,21 +105,36 @@ private static Sheet getFirstSheet(String excelFilePath) {
...
   }
 
   // function: parseRow
   // purpose: parse a single row into a CastVoteRecord
   // returns: a CastVoteRecord object
-  private CastVoteRecord parseRow(Row castVoteRecordRow, String cvrFileName, int cvrIndex) {
+  private CastVoteRecord parseRow(
+      Row castVoteRecordRow,
+      String cvrFileName,
+      int cvrIndex) throws UnrecognizedCandidateException {
```
HEdingfield: I know this looks nicer, but my google-java-format plugin is fighting this when I do a Reformat Code. Do you have it enabled on your end?

src/com/rcv/Main.java
```diff
@@ -13,13 +13,16 @@
 
 package com.rcv;
 
+import com.rcv.CVRReader.SourceWithUnrecognizedCandidatesException;
+import com.rcv.CVRReader.UnrecognizedCandidateException;
```
HEdingfield: This import is unused. Do you have "Optimize Imports" checked in your commit window?

src/com/rcv/CVRReader.java
tarheel: Humble proposal: let's pause our practice of adding useless comments unless/until we get confirmation (from Paul Craft, I guess) that we actually need to do this. It's super tedious, it pollutes the code, and it will be easy to add later if necessary.

src/com/rcv/CVRReader.java
tarheel: I thought I had enabled everything, but clearly it's not working. Let me try again...

src/com/rcv/CVRReader.java
HEdingfield: Totally fine with me. Do you remember where you guys heard it was necessary? Maybe we should ask Paul directly?

tarheel: I forget who it was, but we spoke to someone else who was familiar with the code certification process a long time ago. Yeah, we should confirm all of this stuff with Paul.

Issue #73, opened 2018-05-27
make RawElectionConfig fields private


tarheel:
http://www.baeldung.com/jackson-field-serializable-deserializable-or-not

tarheel:
Once we do this, we may just want to combine RawElectionConfig and ElectionConfig.

HEdingfield:
@tarheel so do you think method 3 in the article is the best way to go (i.e. make public getters for everything)?

HEdingfield:
Cool, out of curiosity, what's the rationale?

tarheel:
If we have public getters for all the raw fields, but then the non-raw version is something else (e.g. the raw config has a string, but in the ElectionConfig object that string gets converted to an enum), it'll be confusing to have both mixed into the same class. So it feels cleaner to keep these separate.

tarheel:
(You'd end up having, say, getOvervoteRuleRaw and getOvervoteRule.)

HEdingfield:
I guess I fundamentally don't understand the difference between the two classes; i.e. why have the "middleman" class of RawElectionConfig in the first place, instead of just sucking it up from the JSON and processing it directly into ElectionConfig?

tarheel:
But how exactly would you implement it as one class?

We have to make the fields public to make the Jackson stuff work. So we'd have publicly accessible raw fields, plus a set of getters for the normalized values. That's confusing.

Alternatively, we could make the fields private and add getters. But then we'd have two getters for some of the fields (raw and normalized). Again, that would be confusing.

CalebKleppner:
Just commenting to get subscribed to this and see if it's still an active issue.

HEdingfield:
@CalebKleppner Yep, still active. FYI, you can just click "Subscribe" under the notifications section in the top right of the issue; no need to comment unless you want to.

HEdingfield:
I was considering getting rid of the getters and setters entirely today, in favor of method 5 in the original article posted in this thread, but realized that, although they seem trivial, public getters and setters are necessary in the GUI for loading candidates and CVR file sources (it's a quirk of the `TableView` object).

Maybe we should just close this issue out at this point and leave `RawContestConfig` as-is?

If so, we'll need to remember to remove:
`// TODO: add getters and setters for everything to address issue #73`

CalebKleppner:
Hyton -- You're a monster -- you never stop.  I'm at Louis and 
Charlotte's house.  We expect Jon in a bit.  Are you on your way?

Cheers,
[...]

PR #94, opened 2018-07-30
fix config file parsing

tarheel:
Unfortunately, any field that's not public in RawElectionConfig just silently stops working -- it remains at its default value regardless of what's in the config file.

I'm going to go ahead and merge this, but I'm creating the PR just to document it.


HEdingfield:
Strange, I wonder if it's because Apache POI needs to work with these values, and it's considered outside of the package? I'll set up IntelliJ to ignore those recommendations for this class in the future.

tarheel:
@HEdingfield: there may be a simple fix for this. Just haven't tried it yet.

See the link I posted in #73.

Issue #85, opened 2018-07-23
continueUntilTwoCandidatesRemain bug


tarheel:
See the benton_comm_pos2_summary.xlsx email thread with James Morales.

tarheel:
Mysteriously, I can't get a repro. Emailed James to ask if it repros for him still.

PR #95, opened 2018-07-30
Fix three minor bugs in Tabulator

tarheel:
1. The mistake that George mentioned. Only undervotes should count as skipped ranks when we're checking whether a ballot has skipped too many ranks.
2. In a couple places, we were improperly counting undervotes and overvotes as continuing candidates. I noticed this because I saw a missing entry in the audit log.
3. The issue that James Morales noticed (#85). The condition in `shouldContinueTabulating` when `willContinueUntilTwoCandidatesRemain` was true wasn't quite right. I was unable to repro this earlier in the weekend because I wasn't actually enabling that setting (which I fixed with the RawElectionConfig commit earlier tonight).


src/com/rcv/Tabulator.java
```diff
@@ -327,6 +327,9 @@ private CandidateStatus getCandidateStatus(String candidate) {
       status = CandidateStatus.WINNER;
     } else if (candidateToRoundEliminated.containsKey(candidate)) {
       status = CandidateStatus.ELIMINATED;
+    } else if (candidate.equals(config.getOvervoteLabel()) || candidate
```
HEdingfield: Code reformatter doesn't like breaking lines on dots, recommends the following on my end:

    } else if (candidate.equals(config.getOvervoteLabel())
        || candidate.equals(config.getUndervoteLabel())) {

tarheel: Sigh. The reformatter actually made it this way on my end to resolve a line of excessive length. I guess it's not totally consistent in its choices. :/

HEdingfield: Yikes! If you go to Settings > Editor > Code Style > Java > Wrapping and Braces, do you have Hard wrap at 100? That should be the Google style default.

tarheel: Yes, I have that enabled.

So I just tested this and discovered something super annoying: the commit hook reformatting doesn't behave the same way that the manual Code > Reformat Code command does. If I start with the excessively long line and use the commit hook, I end up with what you saw in the initial version of this PR. If I instead do the manual command (either before or after the commit hook), it updates it to what you suggested.

tarheel: Another minor thing I've encountered: when I open the push dialog, it always defaults to origin:master, which is almost never what I want. Is there a way to get it to default to a branch of the same name as my local branch (and create that if it doesn't exist yet)? Isn't that a much more common workflow?

HEdingfield: Emailed response!

PR #100, opened 2018-07-31
Rewrote existing Swing GUI in JavaFX.

HEdingfield:
JavaFX is SO MUCH BETTER!


HEdingfield:
Merged with master, try running it now! It should look and act a lot cleaner.

Issue #93, opened 2018-07-29
rcv.log silently becomes huge


HEdingfield:
Realized I had a 200 MB (and growing) rcv.log file sitting in the main directory. Can we address this somehow? Maybe at least break them up into a separate "logs" folder with filenames like rcv_X.log where X is tabulator start date and time? How do other programs handle ever-expanding log files?

tarheel:
Typically you would expect the system to have some process for keeping it under control. I like the idea of separating the output by timestamp. We could also make the app automatically find and erase any logs that are older than X days.

moldover:
Yep, the logging utility has built-in features for rolling over or splitting logs based on time or size.  I prefer a size split (or daily) because it facilitates debugging across application runs (e.g. batch execution, or a user can't remember exactly when something glitched...)

CalebKleppner:
It looks to me like rcv.log is a file that contains an ever-growing log of every tabulation that is run.  I've deleted it a few times when it got too big.  Depending on how we format our logging output, that log can get pretty big, so I'd favor tagging the logfile with tabulation start date and time (rcv_2018-07-25_11-40-19.log), since even a single tabulation might exceed whatever cap size we choose.  Am I correct that the entries in rcv.log are duplicates of the entries in the  audit.log file?  So the main purposes is the simultaneous logging of events in cases where the audit.log file doesn't get created?  If so, we might save some time and performance by eliminating one copy.

moldover:
@CalebKleppner the main point of a separate "rcv.log" is for capturing actions which are outside a tabulation run, for example launching the gui, editing config files, etc.  This is helpful for internal debugging and perhaps record-keeping for running batch executions and unit tests, but on reflection I don't think rcv.log should be part of "official" tabulation outputs nor should we spend much time massaging it's output.  Certainly we should not be duplicating the _audit_ logging in the rcv.log file since that is redundant.  Fixing this will greatly reduce the rcv.log file size, and combined with a reasonable rollover strategy, should (I hope) mean we don't have to worry about this anymore.

HEdingfield:
Agreed, this sounds good to me. Let's keep this issue open a little while after making this change though to get a sense of how big it gets (since we may want to create some overflow option if it's still necessary... hopefully it won't be though).

moldover:
Thinking still further, what about splitting the execution and tabulation log outputs entirely (no duplication).  Something like this:

Execution log:  gui actions, config validation errors, ...
[...]

PR #101, opened 2018-08-01
Logging improvements: different tabulation, execution, gui, and audit logging paths.  Cap file size.

moldover:


src/com/rcv/Logger.java
```diff
@@ -37,6 +37,8 @@
   private static final String TABULATION_LOGGER_NAME = ".tabulation";
   // execution log file name
   private static final String DEFAULT_FILE_NAME = "rcv.log";
+  // first value here is bytes per MB and the second is max MB for the log file
+  private static final Integer TABULATION_LOG_FILE_MAX_SIZE_BYTES = 1000000 * 100;
```
tarheel: nit: if you actually want 100 MB, do 1024 * 1024 * 100

src/com/rcv/Logger.java
```diff
@@ -55,7 +57,10 @@ static void setup() throws IOException {
     // formatter specifies how logging output lines should appear
     LogFormatter formatter = new LogFormatter();
     // fileHandler writes formatted strings to file
-    FileHandler fileHandler = new FileHandler(logPath, true);
+    FileHandler fileHandler = new FileHandler(logPath,
+        TABULATION_LOG_FILE_MAX_SIZE_BYTES,
+        1,
```
tarheel: Are you sure this does what you want? I'm wondering what will happen if you start the app when the file is slightly under the limit. It would be bad if it logged to the file until it hit the limit, then started over on the same file and continued logging. The user would end up not being able to access the lines that were logged before the roll happened.

Maybe a safer approach is to set a limit of 2 and make the per-file size half as big?

src/com/rcv/Logger.java
moldover: https://en.wikipedia.org/wiki/Megabyte

src/com/rcv/Logger.java
moldover: Yeah that's a better idea.

src/com/rcv/Logger.java
HEdingfield: That kind of blows my mind too.

HEdingfield:
Caleb posted some good comments to think about in #93 relevant to this.

moldover:
@HEdingfield @tarheel I think this is ready to be reviewed.

src/main/java/com/rcv/Logger.java
```diff
@@ -35,72 +35,117 @@
...
   }
 
-  // function: info
-  // purpose: log formatted output to console and audit file at INFO level
-  // format: format string into which object params will be formatted
-  // param: obj object to be parsed into format string
-  static void info(String format, Object... obj) {
-    tabulationLogger.info(String.format(format, obj));
+  // logs text to all output loggers
+  static void allLogs(Level level, String format, Object... obj) {
```
HEdingfield: My intuition is telling me that there's a more elegant way to implement the loggers so we can just make calls like `executionLogger.warn("Blah!")` instead of `Logger.executionLog(Level.WARN, "Blah!")`. In other words, maybe instead of having these four separate functions for logging in various permutations, it makes more sense to maintain four separate loggers with different permutations of handlers?

src/main/java/com/rcv/GuiMainController.java
```diff
@@ -33,6 +33,7 @@
 import javafx.stage.FileChooser;
 import javafx.stage.FileChooser.ExtensionFilter;
 import javafx.stage.Stage;
+import java.util.logging.Level;
```
HEdingfield: In this class (`GuiMainController`), I think you're going to want to replace all usages of `printToTextStatus()` with logging and delete that method entirely. You may want to recycle some of the date / formatting logic in `printToTextStatus()` into the logger.

src/main/java/com/rcv/Main.java
```diff
@@ -70,41 +70,18 @@ public static void main(String[] args) {
...
+    ElectionConfig config;
 
     // rawConfig holds the basic election config data parsed from json
     RawElectionConfig rawConfig =
         JsonParser.parseObjectFromFile(configPath, RawElectionConfig.class);
 
     if (rawConfig == null) {
-      System.err.println(String.format("Failed to load config file: %s", configPath));
+      Logger.executionLog(Level.SEVERE, String.format("Failed to load config file: %s", configPath));
+      return null;
```
HEdingfield: We're not supposed to have multiple returns, ya jabroni!

src/main/java/com/rcv/Main.java
```diff
@@ -70,41 +70,18 @@ public static void main(String[] args) {
...
-        for (String error : validationErrors) {
-          Logger.severe("Invalid config: %s", error);
-        }
-        config = null;
-      }
-    }
-
-    if (encounteredFileError) {
-      config = null;
+      config.validate();
```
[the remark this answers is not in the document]
moldover: Yeah, I realized after the commit I had failed to check the boxes.  Will fix this and the others.  Clearly I have not been making enough commits lately!

As far as the return value, the user should still be able to load into GUI, edit, and save an invalid config file, hence we return the config object here.

src/main/java/com/rcv/Main.java
moldover: 28452_the-rock_l

src/main/java/com/rcv/Main.java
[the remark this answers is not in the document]
HEdingfield: Hmm, the problem I ran into is that I could still click Tabulate when there was a bad output path, and it would fail and write an audit log to the CWD, which I don't think is desirable behavior. What if, instead of doing validation on load, we do it on save and tabulate actions? I think that might make more sense. This is also making me think that maybe the validation shouldn't write to the audit log until the output path is verified as valid.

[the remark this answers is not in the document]
moldover: Yep, Validate should NOT be part of Load or Save.

Validate MUST be the first step of Tabulate (tabulation fails if validation fails).

We should have a "Validate" button on the editor UI, that the user can push at any time to check their work.  (to this end, I think the editor UI needs a console window to show their errors).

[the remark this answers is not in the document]
HEdingfield: Agreed, I'll plan to bring a separate validate button in. In this PR, it'd be great if you could move validate from the load method over to the tabulate one.

src/main/java/com/rcv/Logger.java
HEdingfield: No problem, I'll think about it more and reassess once we get this merged in.

src/main/java/com/rcv/Main.java
HEdingfield: This man could easily be our next president if we get more RCV laws passed.

src/main/java/com/rcv/Main.java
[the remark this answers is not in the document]
moldover: my pleasure :)

src/main/java/com/rcv/Logger.java
moldover: I'm gonna keep thinking about it too.  I think at least "auditLog" can simply be a log call to tabulationLogger with level = verbose, and we use level filter to keep that out of the GUI.  This stuff isn't critical, but it's nice to keep it tight :)

Issue #45, opened 2018-04-12
Enable config file creation / modification in GUI


HEdingfield:
* GUI should have ability to create, save, and modify config files
* Dropdowns for overvotes, ties, etc.
* Possible flow: first screen has load config or create config, next lets user review / modify and then saves config to output, last screen runs tabulator

HEdingfield:
Specific user feedback:

"Is there any chance the setup can be streamlined somewhat.

It would be great if there were an interactive questionnaire that could populate the JSON file parameter fields for each RCV contest without having to wade through the code to find each instruction.

Here’s a brief example of what I comes to mind.

1.       How many RCV contest are in this election? Answer = 3

2.       What is the name of the first contest? Answer = County Commissioner Pos 2

3.       Number to be Elected? Answer = 1

4.       Number of declared candidates? Answer = 5

5.       Is there a Write-In Position? Answer = YES

6.       Number of Rankings for this contest = 3

7.       Number of Precincts for this Contest = 24

8.       What is the name of the first contest? Answer = County Commissioner Pos 3

9.       Number to be Elected? Answer = 1

10.   Number of declared candidates? Answer = 6

11.   Is there a Write-In Position? Answer = YES

12.   Number of Rankings for this contest = 4

13.   Number of Precincts for this Contest = 24

14.   What is the name of the first contest? Answer = Advisory Measure_OR

15.   Number to be Elected? Answer = 1

16.   Number of declared candidates? Answer = 5

17.   Is there a Write-In Position? Answer = NO

18.   Number of Rankings for this contest = 5

19.   Number of Precincts for this Contest = 24"

HEdingfield:
Potential direction for UI documented in #87.

HEdingfield:
Can also explore "wizard" route outlined in Oracle's docs [here](http://www.oracle.com/technetwork/articles/javaee/wizard-136789.html).

HEdingfield:
Design discussion and decisions made in #87 doc:
https://docs.google.com/document/d/1jQmhIoHNo0cNxD8h3xDHmiGyvxaUwbFi0Ieh3-W-2_U/

HEdingfield:
Next (and last, I think) step is to develop modification of existing config files.

PR #106, opened 2018-08-04
GUI save config

HEdingfield:
Helps address #45.

### DIFF
--- a/src/com/rcv/JsonParser.java
+++ b/src/com/rcv/JsonParser.java
@@ -10,6 +10,7 @@ package com.rcv;
 import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 
@@ -22,6 +23,7 @@ class JsonParser {
   // file access: read
   // returns: instance of the object parsed from json or null if there was a problem
   static <T> T parseObjectFromFile(String jsonFilePath, Class<T> valueType) {
+    // TODO: consolidate two return statements in this method into one?
     try {
       // fileReader will read the json file from disk
       FileReader fileReader = new FileReader(jsonFilePath);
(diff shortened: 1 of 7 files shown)


HEdingfield:
One thing I'm wondering: currently I have it set up so the first option in an enum is the default value, and will be automatically selected in the choice boxes. In light of this, should we rearrange any of the enums?

src/com/rcv/JsonParser.java
```diff
@@ -22,6 +23,7 @@
   // file access: read
   // returns: instance of the object parsed from json or null if there was a problem
   static <T> T parseObjectFromFile(String jsonFilePath, Class<T> valueType) {
+    // TODO: consolidate two return statements in this method into one?
```
tarheel: oh, yeah, please fix that while you're here, if you don't mind

src/com/rcv/Tabulator.java
```diff
@@ -847,20 +848,56 @@ void doAudit(List<CastVoteRecord> castVoteRecords) {
...
+    TRANSFER_WHOLE_SURPLUS("transferWholeSurplus"),
+    TRANSFER_RULE_UNKNOWN("transferRuleUnknown");
+
+    private final String label;
+
+    MultiSeatTransferRule(String label) {
+      this.label = label;
+    }
+
+    static MultiSeatTransferRule getByLabel(String labelLookup) {
```
tarheel: Darn. When I saw that you were using this above, I was hoping that it was a built-in thing that we had overlooked. But anyway, yeah, this is better encapsulation.

src/com/rcv/JsonParser.java
HEdingfield: Yep, done in the next PR.

HEdingfield:
> Looks good. As before, I didn't look closely at the UI stuff.

Still definitely appreciate you checking for general good Java coding practices!

> Is it possible to have these selectors default to nothing or a dummy "Select one" value? If you don't want to go that route, I don't really know what the right default values would be. I'd rather force the user to consciously select each setting.

Yep, I can do this. Only inconsistency is that the optional choices all have default values, so those will be auto-filled even though they aren't needed...

tarheel:
> Only inconsistency is that the optional choices all have default values, so those will be auto-filled even though they aren't needed...

OK. Once we have the whole UI in place, it'll be easier to reason about whether some of those defaults should be removed.

PR #108, opened 2018-08-05
GUI config creator

HEdingfield:
Further addresses #45.

### DIFF
--- a/src/com/rcv/ElectionConfig.java
+++ b/src/com/rcv/ElectionConfig.java
@@ -228,7 +228,7 @@ class ElectionConfig {
   // purpose: getter for electionDate
   // returns: election date
   String getElectionDate() {
-    return rawConfig.date;
+    return rawConfig.contestDate;
   }
 
   // function: isTabulateByPrecinctEnabled
--- a/src/com/rcv/ElectionConfig.java
+++ b/src/com/rcv/ElectionConfig.java
@@ -87,6 +87,8 @@ class ElectionConfig {
     // detected errors
     List<String> errors = new LinkedList<>();
 
+    // TODO: need to add checks that all required String fields !.equals("")
+
     if (getNumDeclaredCandidates() == 0) {
       errors.add("Config must contain at least one declared candidate.");
     }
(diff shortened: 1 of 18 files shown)


src/com/rcv/ElectionConfig.java
```diff
@@ -87,6 +87,8 @@
     // detected errors
     List<String> errors = new LinkedList<>();
 
+    // TODO: need to add checks that all required String fields !.equals("")
```
tarheel: #109

src/com/rcv/GuiConfigController.java
```diff
@@ -48,16 +77,41 @@ public void initialize(URL location, ResourceBundle resources) {
...
     choiceOvervoteRule.getItems().remove(Tabulator.OvervoteRule.RULE_UNKNOWN);
-    choiceOvervoteRule.getSelectionModel().select(0);
+
+    datePickerContestDate.setConverter(
+        new StringConverter<>() {
+          final DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+
+          @Override
+          public String toString(LocalDate date) {
+            return (date != null) ? dateFormatter.format(date) : "";
```
tarheel: Style nit: I'm not in the habit of putting the ternary condition in parentheses because it feels superfluous. Does the Google style say that we should? Any particular arguments for or against?

src/com/rcv/JsonParser.java
```diff
@@ -23,24 +23,26 @@
   // file access: read
   // returns: instance of the object parsed from json or null if there was a problem
   static <T> T parseObjectFromFile(String jsonFilePath, Class<T> valueType) {
-    // TODO: consolidate two return statements in this method into one?
```
tarheel: grazie

src/com/rcv/GuiConfigController.java
HEdingfield: That's my own peculiarity (I find it to be more readable). But sure, I'll remove the parens and won't use them in the future.

HEdingfield:
> One general comment: I get the sense that `GuiConfigController` is going to become pretty unwieldy as you add more form fields and their corresponding handlers. Have you thought about how we might break this logic down into more subclasses? Like, is there a way that each param and its handlers could be enclosed within a class, and the main controller would just have a simple reference to that class?

I think I've actually got 30 - 50% of the code written in there actually. I've been taking a broad approach going for different types of widgets first. We'll start to see economy of scale soon I think. I tend to favor larger classes over tons of separate files, but I'm definitely open to splitting things out when it makes sense (e.g. if I use something more than once).

Issue #98, opened 2018-07-31
Create documentation tooltips for each field in config creator


HEdingfield:
First pass at this can be a simple button that opens up the config_file_documentation.txt.

Eventually, it'd be nice to get full on context-sensitive tooltips integrated.

HEdingfield:
The config creator GUI now has a pane on the right that allows viewing of the documentation. In the future, it'd be cool to work this in as interactive tooltips on each field instead.

gngilbert:
Agree with @HEdingfield as a future enhancement

PR #113, opened 2018-08-07
GUI help

HEdingfield:
Adds help pane to config creator GUI, fixing #98.


moldover:
Louis can't be stopped.

src/main/java/com/rcv/GuiMainController.java
```diff
@@ -41,15 +41,15 @@
 
   // text area which communicates the status of the tabulator's operations
   @FXML
-  private TextArea textStatus;
+  private TextArea textAreaStatus;
```
moldover: Is the textAreaStatus scrollable (or otherwise expandable) or more of a single-line status?

HEdingfield: Yep, TextAreas have that functionality out of the box with JavaFX. It's great!

HEdingfield:
I will happily accept Louis's presumably drugged-out post-surgery approvals.

PR #114, opened 2018-08-08
GUI additional config fields

HEdingfield:
More progress on #45. Partially addresses #73 too.

### DIFF
--- a/src/main/java/com/rcv/GuiConfigController.java
+++ b/src/main/java/com/rcv/GuiConfigController.java
@@ -59,11 +59,13 @@ public class GuiConfigController implements Initializable {
   @FXML
   private TextField textFieldOutputDirectory;
   @FXML
-  private ChoiceBox<Tabulator.OvervoteRule> choiceOvervoteRule;
+  private DatePicker datePickerContestDate;
+  @FXML
+  private TextField textFieldContestJurisdiction;
+  @FXML
+  private TextField textFieldContestOffice;
   @FXML
   private ToggleGroup toggleTabulateByPrecinct;
-  @FXML
-  private DatePicker datePickerContestDate;
   @FXML
   private TableView<Candidate> tableViewCandidates;
   @FXML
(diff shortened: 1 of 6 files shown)


src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -93,6 +109,27 @@ public void buttonSaveClicked() {
...
+    // TODO: check if candidate is already in list?
+    if (!textFieldCandidateName.getText().isEmpty()) {
+      candidate.setName(textFieldCandidateName.getText());
+      candidate.setCode(textFieldCandidateCode.getText());
+      tableViewCandidates.getItems().add(candidate);
+      textFieldCandidateName.clear();
+      textFieldCandidateCode.clear();
+    } else {
+      // TODO: Need to convey this in the UI; also consider moving validation to setter
+      Logger.warn("Candidate name field is required!");
```
moldover: How about, just set whatever the user has entered (even if invalid, e.g. a code but no name) and defer all checks till they 'validate'?  Keeps all the validation code in one place instead of split between UI elements.  The one thing which might be a useful exception to this: if there's an easy way to enforce valid characters in the textField itself, some frameworks do this for you.  We could specify no funky characters (emoji, copy-pasted partial unicode strings, etc..)

HEdingfield: I agree that there's a big benefit to keeping the validation code in one place, and I think we should do that as a general rule (certainly we should at the very least replicate any checks we do in the GUI). I'll leave the required field checks off most of the text fields for now, and leave it up to the backend validation. I'll also leave off a check in the frontend for duplicate values.

However, given the nature of this particular widget (a TableView, where entries are submitted and added one at a time), it seems like a big timesaver to have a simple check to prevent entering a null required value. I'm going off of web design principles here, where you'd want to validate on both the backend (e.g. in the setter) and via JS or HTML in the frontend.

Unfortunately I didn't see any quick and easy ways to filter out "weird" characters; I also think it's important to allow for unicode characters in case our tool is used for contests in other languages. Maybe we can add a check down the line so people don't add "(╯°□°）╯︵ ┻━┻" as a candidate, but it doesn't seem like a big risk to me at the moment (feel free to open a new issue if you disagree).

tarheel: FWIW, I would be open to casting a vote for "(╯°□°）╯︵ ┻━┻".

PR #116, opened 2018-08-10
GUI config rules

HEdingfield:
More progress on #45. Reviewing one commit at a time might be easiest.

### DIFF
--- a/src/main/java/com/rcv/GuiConfigController.java
+++ b/src/main/java/com/rcv/GuiConfigController.java
@@ -27,6 +27,8 @@ import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.ResourceBundle;
 import java.util.stream.Collectors;
+import javafx.beans.value.ChangeListener;
+import javafx.beans.value.ObservableValue;
 import javafx.event.ActionEvent;
 import javafx.fxml.FXML;
 import javafx.fxml.FXMLLoader;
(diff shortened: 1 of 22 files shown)


src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -192,10 +239,24 @@ private void saveElectionConfig(File saveFile) {
...
+            : Tabulator.TieBreakMode.MODE_UNKNOWN.toString();
     rules.overvoteRule =
         choiceOvervoteRule.getValue() != null
             ? choiceOvervoteRule.getValue().toString()
             : Tabulator.OvervoteRule.RULE_UNKNOWN.toString();
+    rules.multiSeatTransferRule =
+        choiceMultiSeatTransferRule.getValue() != null
+            ? choiceMultiSeatTransferRule.getValue().toString()
+            : Tabulator.MultiSeatTransferRule.TRANSFER_RULE_UNKNOWN.toString();
+    rules.maxRankingsAllowed = Integer.parseInt(textFieldMaxRankingsAllowed.getText());
```
tarheel: Will these trigger an uncaught exception if the text isn't a valid integer? Do your listeners make that truly impossible, or just unlikely?

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -192,36 +200,56 @@ public LocalDate fromString(String string) {
...
-    textFieldMinimumVoteThreshold.textProperty().addListener((observable, oldValue, newValue) -> {
-      if (!newValue.matches("\\d*")) {
-        textFieldMinimumVoteThreshold.setText(oldValue);
-      }
-    });
+    // Restrict text fields to non-negative integers and set default values where applicable
+    textFieldMaxRankingsAllowed
+        .textProperty()
+        .addListener(
+            (observable, oldValue, newValue) -> {
```
tarheel: If you can access the field itself from the listener args (maybe via observable?), perhaps you can avoid duplicating this same logic for each field? The amount of repetition here is unfortunate.

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -251,12 +279,40 @@ private void saveElectionConfig(File saveFile) {
...
+    rules.decimalPlacesForVoteArithmetic =
+        !textFieldDecimalPlacesForVoteArithmetic.getText().isEmpty()
+            ? Integer.parseInt(textFieldDecimalPlacesForVoteArithmetic.getText())
+            : ElectionConfig.DEFAULT_DECIMAL_PLACES_FOR_VOTE_ARITHMETIC;
+    rules.minimumVoteThreshold =
+        !textFieldMinimumVoteThreshold.getText().isEmpty()
+            ? Integer.parseInt(textFieldMinimumVoteThreshold.getText())
+            : ElectionConfig.DEFAULT_MINIMUM_VOTE_THRESHOLD.intValue();
+    rules.batchElimination =
+        ((RadioButton) toggleBatchElimination.getSelectedToggle()).getText().equals("True");
```
tarheel: Could also factor out this bit of repeated logic.

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: Yes, it's impossible. Anything that doesn't match regex "\\d*" in these text fields is immediately reverted, as if it was never typed at all. Maybe not the best UX (ideally we'd have a pop-up note that says "integers only"), but it'll do the trick in the short-term.

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: Wasn't possible via observable unfortunately, but I made a new inner listener class in the latest commit.

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: Done in latest commit, along with additional refactoring. Please re-approve when you get a chance to review. 💯

PR #117, opened 2018-08-11
Developed CVR Files pane of config creator GUI

HEdingfield:
Final piece needed for #45. Also more progress on #73.


HEdingfield:
So say we all.

PR #122, opened 2018-08-16
Added Modify Existing Config button and functionality to GUI; Electio…

### LINKED ISSUE #45: Enable config file creation / modification in GUI

HEdingfield:
…nConfig: got rid of unnecessary helper functions for enum fields, got rid of default values for required enum fields, added validation checks for required integer fields; fixed office and jurisdiction field names in test config files. Fixes #45. Progress on #109.

### DIFF
--- a/src/main/java/com/rcv/GuiConfigController.java
+++ b/src/main/java/com/rcv/GuiConfigController.java
@@ -31,6 +31,7 @@ import java.util.logging.Level;
 import java.util.stream.Collectors;
 import javafx.beans.value.ChangeListener;
 import javafx.beans.value.ObservableValue;
+import javafx.collections.FXCollections;
 import javafx.event.ActionEvent;
 import javafx.fxml.FXML;
 import javafx.fxml.FXMLLoader;
@@ -56,6 +57,7 @@ import javafx.util.StringConverter;
 
 public class GuiConfigController implements Initializable {
 
+  private static final DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
   @FXML
   private TextArea textAreaHelp;
   @FXML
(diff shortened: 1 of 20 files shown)


src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -159,9 +178,8 @@ public void buttonOutputDirectoryClicked() {
     dc.setTitle("Output Directory");
 
     File outputDirectory = dc.showDialog(null);
-    if (outputDirectory != null) {
```
tarheel: I'd prefer to stick with braces even for single-line cases like this, if you don't mind. We've been doing that everywhere else.

test_data/config_invalid_params.json
```diff
@@ -1,13 +1,13 @@
...
-  "office":"none",
+  "contestJurisdiction": "none",
+  "contestOffice": "nonie",
   "contestDate": "2018-07-19",
   "rules": {
     "rulesDescription": "Testing our validation logic for params",
-    "batchElimination":true,
-    "continueUntilTwoCandidatesRemain":true,
+    "batchElimination": "True",
+    "continueUntilTwoCandidatesRemain": "false",
```
tarheel: Why is the casing inconsistent (True vs. false)?

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: I don't think I've been following that convention in my code, but I'll put these ones back in in my next commit.

Issue #60, opened 2018-04-25
Consider gathering candidates from CVRs rather than specifying them in the config


HEdingfield:
Any compelling reasons not to do this?

tarheel:
I'm not sure. Technically there's no problem, of course. I'd defer to Caleb and George on what's appropriate: a) require listing candidates (current approach), b) never list candidates in the config, or c) optionally list candidates.

HEdingfield:
Yeah, would be good to get their opinion. Here's what I think makes the most sense:

1. User loads CVRs, and relevant data is scraped (candidates, precincts, etc.)
2. User is then brought to a screen where they can configure the tabulation and they're presented with options here that are specific to the behavior of the tabulator alone. They could also load an existing config file here.
3. User executes tabulation.

CalebKleppner:
This would be a nice feature, but it's not needed for testing submission 1.0, so let's keep our focus on that.

gngilbert:
We do this for CDF input files.  How much work would it be to do it for ESS inputs?

tarheel:
The actual code wouldn't be very hard. The part I'm uncertain about is the user experience. How exactly would this work? I can think of at least a couple options. @HEdingfield's suggestion above sounds good, but would be a pain to implement in the UI. A couple alternatives:

1. On the candidates tab, there's a button labeled "Import from CVR sources" that triggers an immediate parsing of the source files, then overwrites the displayed list with whatever it finds. (In other words, the candidates still go in the config file, but we help the user auto-populate them.)
2. On the candidates tab, there's a checkbox to toggle whether the candidates are specified explicitly vs. inferred from the CVR sources. If the latter, the config doesn't contain candidate info, and the tabulator just assumes that any candidate it encounters while parsing the CVR sources is valid.

gngilbert:
I like option 1.

gngilbert:
One contingency we need to consider is a post-adjudication CVR where write-ins have been normalized.  Obviously there could be errors in the normalization process or records missed in the adjudication.  auto-populating the candidate names list would help catch these types of errors.

andyanderson:
[...]

PR #126, opened 2018-08-23
GUI validation

HEdingfield:
After a 15 hour day, I give you the greatest, most bloated PR ever created. You can just scroll and scroll and scroll, Morty, it never ends.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,7 +1,7 @@
                     GNU AFFERO GENERAL PUBLIC LICENSE
                        Version 3, 19 November 2007
 
- Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Copyright (c) 2007 Free Software Foundation, Inc. <http://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 
--- a/src/main/java/com/rcv/ElectionConfig.java
+++ b/src/main/java/com/rcv/ElectionConfig.java
@@ -20,7 +20,6 @@
 
 package com.rcv;
 
-import com.rcv.FileUtils.UnableToCreateDirectoryException;
 import com.rcv.Tabulator.TieBreakMode;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
@@ -60,18 +59,7 @@ class ElectionConfig {
     boolean isValid = true;
 
     Logger.executionLog(Level.INFO, "Validating config file...");
-
-    try {
-      FileUtils.createOutputDirectory(this.getOutputDirectory());
-    } catch (UnableToCreateDirectoryException exception) {
-      isValid = false;
-      Logger.executionLog(
-          Level.SEVERE,
-          String.format(
-              "Failed to create output directory: %s\n%s",
-              this.getOutputDirectory(), exception.toString()));
-    }
-
+    // TODO: reorder all checks so they go in same order as fields in config GUI
     // TODO: need to add checks that all required String fields !.equals("")
 
     if (getNumDeclaredCandidates() == 0) {
@@ -85,7 +73,7 @@ class ElectionConfig {
     }
 
     if (getOvervoteRule() == Tabulator.OvervoteRule.RULE_UNKNOWN) {
-      // TODO: report what the invalid value was?
+      isValid = false;
       Logger.executionLog(Level.SEVERE, "Invalid overvote rule.");
     } else if (getOvervoteLabel() != null
         && getOvervoteRule() != Tabulator.OvervoteRule.EXHAUST_IMMEDIATELY
@@ -119,7 +107,6 @@ class ElectionConfig {
     }
 
     // If this is a multi-seat election, we validate a number of extra parameters.
-    //
     if (getNumberOfWinners() > 1) {
       if (willContinueUntilTwoCandidatesRemain()) {
         isValid = false;
@@ -196,11 +183,9 @@ class ElectionConfig {
   // returns: directory string from config or falls back to working directory
   String getOutputDirectory() {
     // outputDirectory is where output files should be written
-    String outputDirectory = System.getProperty("user.dir");
-    if (rawConfig.outputDirectory != null) {
-      outputDirectory = rawConfig.outputDirectory;
-    }
-    return outputDirectory;
+    return rawConfig.outputDirectory != null
+        ? rawConfig.outputDirectory
+        : System.getProperty("user.dir");
   }
 
   // function: willContinueUntilTwoCandidatesRemain
@@ -384,7 +369,8 @@ class ElectionConfig {
     if (rawConfig.candidates != null) {
       // candidate is used to index through all candidates for this election
       for (RawElectionConfig.Candidate candidate : rawConfig.candidates) {
-        if (candidate.getCode() != null) {
+        // TODO: go through and add checks for isEmpty in all other String != null situations?
+        if (candidate.getCode() != null && !candidate.getCode().isEmpty()) {
           candidateCodeToNameMap.put(candidate.getCode(), candidate.getName());
           candidatePermutation.add(candidate.getCode());
         } else {
@@ -398,7 +384,7 @@ class ElectionConfig {
       }
 
       String uwiLabel = getUndeclaredWriteInLabel();
-      if (uwiLabel != null) {
+      if (uwiLabel != null && !uwiLabel.isEmpty()) {
         candidateCodeToNameMap.put(uwiLabel, uwiLabel);
       }
     }
--- a/src/main/java/com/rcv/ElectionConfig.java
+++ b/src/main/java/com/rcv/ElectionConfig.java
@@ -58,7 +58,7 @@ class ElectionConfig {
     // return value will be false if there are any validation errors
     boolean isValid = true;
 
-    Logger.executionLog(Level.INFO, "Validating config file...");
+    Logger.executionLog(Level.INFO, "Validating config...");
     // TODO: reorder all checks so they go in same order as fields in config GUI
     // TODO: need to add checks that all required String fields !.equals("")
 
@@ -134,9 +134,9 @@ class ElectionConfig {
     }
 
     if (isValid) {
-      Logger.executionLog(Level.INFO, "Validation successful.");
+      Logger.executionLog(Level.INFO, "Config validation successful.");
     } else {
-      Logger.executionLog(Level.SEVERE, "Validation failed!");
+      Logger.executionLog(Level.SEVERE, "Config validation failed!");
     }
 
     return isValid;
--- a/src/main/java/com/rcv/ElectionConfig.java
+++ b/src/main/java/com/rcv/ElectionConfig.java
@@ -1,6 +1,6 @@
 /*
  * Ranked Choice Voting Universal Tabulator
- * Copyright (C) 2018 Jonathan Moldover, Louis Eisenberg, and Hylton Edingfield
+ * Copyright (c) 2018 Jonathan Moldover, Louis Eisenberg, and Hylton Edingfield
  *
  * This program is free software: you can redistribute it and/or modify it under the terms of the
  * GNU Affero General Public License as published by the Free Software Foundation, either version 3
@@ -180,13 +180,6 @@ class ElectionConfig {
   // returns: the quotient
   BigDecimal divide(BigDecimal dividend, BigDecimal divisor) {
     return dividend.divide(divisor, getDecimalPlacesForVoteArithmetic(), RoundingMode.HALF_EVEN);
-  }
-
-  // function: roundDecimal
-  // purpose: round a number according to the config settings
-  // returns: the rounded value
-  BigDecimal roundDecimal(BigDecimal bd) {
-    return bd.setScale(getDecimalPlacesForVoteArithmetic(), RoundingMode.HALF_EVEN);
   }
 
   // function: getMultiSeatTransferRule
--- a/src/main/java/com/rcv/ElectionConfig.java
+++ b/src/main/java/com/rcv/ElectionConfig.java
@@ -20,12 +20,15 @@
 
 package com.rcv;
 
+import com.rcv.RawElectionConfig.CVRSource;
+import com.rcv.RawElectionConfig.Candidate;
 import com.rcv.Tabulator.TieBreakMode;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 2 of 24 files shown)


src/main/java/com/rcv/ElectionConfig.java
```diff
@@ -405,7 +490,7 @@ private void processCandidateData() {
       }
 
       String uwiLabel = getUndeclaredWriteInLabel();
-      if (uwiLabel != null) {
+      if (uwiLabel != null && !uwiLabel.isEmpty()) {
```
HEdingfield: Should we actually be excluding `uwiLabel` from `candidatePermutation`? Doesn't that give an unfair disadvantage to UWIs?

tarheel: If uwiLabel exists, it automatically gets eliminated in the first round.

LICENSE
```diff
@@ -1,7 +1,7 @@
                     GNU AFFERO GENERAL PUBLIC LICENSE
                        Version 3, 19 November 2007
 
- Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Copyright (c) 2007 Free Software Foundation, Inc. <http://fsf.org/>
```
tarheel: These lines are truly the highlight of the PR.

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -243,7 +252,7 @@ public void buttonDeleteCandidateClicked() {
 
   @Override
   public void initialize(URL location, ResourceBundle resources) {
-    Logger.executionLog(Level.INFO, "Opening config creator GUI...");
+    Logger.executionLog(Level.FINE, "Opening config creator GUI...");
```
tarheel: What's the difference?

src/main/java/com/rcv/Logger.java
```diff
@@ -136,7 +141,11 @@ static void addGuiLogging(TextArea textArea) {
 
           @Override
           public void publish(LogRecord record) {
-            Platform.runLater(() -> textArea.appendText(formatter.format(record)));
+            if (Platform.isFxApplicationThread()) {
+              textArea.appendText(formatter.format(record));
+            } else {
```
tarheel: When will we hit this condition?

src/main/java/com/rcv/Logger.java
```diff
@@ -155,13 +164,15 @@ public void close() {
...
   // throws: IOException if unable to open loggerOutputPath
   static void addTabulationFileLogging(String loggerOutputPath) throws IOException {
-    // create new handler for file logging and add it
-    FileHandler fileHandler = new FileHandler(loggerOutputPath);
     // specifies how logging output lines should appear
     LogFormatter formatter = new LogFormatter();
+    // create new handler for file logging and add it
+    FileHandler fileHandler = new FileHandler(loggerOutputPath);
     fileHandler.setFormatter(formatter);
+    fileHandler.setLevel(Level.FINER);
```
tarheel: Ohhh. I thought FINE meant "OK."

src/main/java/com/rcv/Main.java
```diff
@@ -45,24 +45,26 @@ public static void main(String[] args) {
...
       ElectionConfig config = loadElectionConfig(configPath);
       if (config != null) {
-        Logger.executionLog(Level.INFO, "Tabulator is being used via the CLI.");
         executeTabulation(config);
       } else {
         Logger.executionLog(Level.SEVERE, "Aborting because config is invalid.");
       }
     }
+
+    System.exit(0);
```
tarheel: Just curious what this does. Won't we exit with a 0 automatically?

src/main/java/com/rcv/ElectionConfig.java
```diff
@@ -59,23 +62,119 @@ boolean validate() {
...
+
+    if (getOutputDirectory() == null || getOutputDirectory().isEmpty()) {
+      isValid = false;
+      Logger.executionLog(Level.SEVERE, "Output directory is required.");
+    }
+
+    if (rawConfig.cvrFileSources == null || rawConfig.cvrFileSources.isEmpty()) {
+      isValid = false;
+      Logger.executionLog(Level.SEVERE, "Config doesn't contain any CVR files.");
+    } else {
```
tarheel: Could you factor this out into a helper? `validate` is getting really long.

src/main/java/com/rcv/ElectionConfig.java
```diff
@@ -59,23 +62,119 @@ boolean validate() {
...
+      HashSet<String> cvrFilePathSet = new HashSet<>();
+      for (CVRSource source : rawConfig.cvrFileSources) {
+        if (source.getFilePath() == null || source.getFilePath().isEmpty()) {
+          isValid = false;
+          Logger.executionLog(Level.SEVERE, "filePath is required for each CVR file.");
+        } else if (cvrFilePathSet.contains(source.getFilePath())) {
+          isValid = false;
+          Logger.executionLog(
+              Level.SEVERE, "Duplicate CVR filePaths are not allowed: %s", source.getFilePath());
```
tarheel: I agree with this, but FYI, it breaks our scale test.

src/main/java/com/rcv/ElectionConfig.java
```diff
@@ -59,23 +62,119 @@ boolean validate() {
...
+    }
+
+    if (candidateCodeSet.size() > 0 && candidateCodeSet.size() != candidateNameSet.size()) {
+      isValid = false;
+      Logger.executionLog(
+          Level.SEVERE,
+          "If candidate codes are used, a unique code is required for each candidate.");
+    }
+
+    if (getNumDeclaredCandidates() < 1) {
```
tarheel: `== 0` seems more intuitive

src/main/java/com/rcv/Main.java
```diff
@@ -170,102 +170,52 @@ static void executeTabulation(ElectionConfig config) {
...
-          Logger.tabulationLog(Level.SEVERE, "Invalid source file: missing filePath");
-          encounteredProblemForThisSource = true;
-        }
+    // did we encounter a fatal problem for this source?
+    boolean encounteredSourceProblem = false;
+
+    // At each iteration of the following loop, we add records from another source file.
+    // source: index over config sources
+    for (RawElectionConfig.CVRSource source : config.rawConfig.cvrFileSources) {
+      String logMessage = String.format("Reading CVR file: %s", source.getFilePath());
```
tarheel: prefer `StringBuilder`

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -221,7 +224,8 @@ private void generateSummarySpreadsheet(
       }
     }
 
-    if (precinct == null) { // actions don't make sense in individual precinct results
+    // actions don't make sense in individual precinct results
+    if (precinct == null || precinct.isEmpty()) {
```
tarheel: We should never be passing an empty string here...

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -187,7 +187,10 @@ private void generateSummarySpreadsheet(
     // rowCounter contains the next empty row after all the general header rows have been created.
     // This is where we start adding round-by-round reports. For precinct sheets, there are no
     // general header rows, so we just start with the round-by-round reports.
-    int rowCounter = precinct != null ? 0 : addHeaderRows(worksheet, totalActiveVotesPerRound);
+    int rowCounter =
+        precinct != null && !precinct.isEmpty()
```
tarheel: or here

src/main/java/com/rcv/Tabulator.java
```diff
@@ -754,7 +755,7 @@ private boolean hasContinuingCandidates(SortedMap<Integer, Set<String>> rankToCa
             }
             candidatesSeen.add(candidate);
           }
-          if (duplicateCandidate != null) {
+          if (duplicateCandidate != null && !duplicateCandidate.isEmpty()) {
```
tarheel: this shouldn't happen

src/main/java/com/rcv/Tabulator.java
```diff
@@ -805,7 +808,7 @@ private boolean hasContinuingCandidates(SortedMap<Integer, Set<String>> rankToCa
           }
         }
 
-        if (selectedCandidateID != null) {
+        if (selectedCandidateID != null && !selectedCandidateID.isEmpty()) {
```
tarheel: same

src/main/java/com/rcv/TieBreak.java
```diff
@@ -127,7 +127,7 @@ String selectLoser() {
         // loser: will be set if there is a previous round count loser
         // it will be null if candidates were still tied at first round
         String previousRoundsLoser = doPreviousRounds();
-        if (previousRoundsLoser != null) {
+        if (previousRoundsLoser != null && !previousRoundsLoser.isEmpty()) {
```
tarheel: same

src/main/java/com/rcv/TieBreak.java
```diff
@@ -178,7 +178,7 @@ private String doInteractive() {
         "Enter the number corresponding to the candidate who should lose this tiebreaker.");
     // the candidate selected to lose
     String selectedCandidate = null;
-    while (selectedCandidate == null) {
+    while (selectedCandidate == null || selectedCandidate.isEmpty()) {
```
tarheel: same

src/main/java/com/rcv/TieBreak.java
```diff
@@ -189,10 +189,10 @@ private String doInteractive() {
           // Convert from 1-indexed list back to 0-indexed list.
           selectedCandidate = tiedCandidates.get(choice - 1);
         }
-      } catch (NumberFormatException e) {
+      } catch (NumberFormatException exception) {
         // if parseInt failed selectedCandidate will be null and we will retry
       }
-      if (selectedCandidate == null) {
+      if (selectedCandidate == null || selectedCandidate.isEmpty()) {
```
tarheel: same

LICENSE
HEdingfield: Agreed, it was driving me to distraction.

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: FINE isn't logged in the GuiLogger, so there's less noise in there.

src/main/java/com/rcv/ElectionConfig.java
HEdingfield: Maybe, but < 1 covers more bases, e.g. if in future changes some bug slips in that allows getNumDeclaredCandidates() to return a negative number.

src/main/java/com/rcv/Main.java
HEdingfield: Yeah, I looked this up recently, and it does seem like [StringBuilder is better-performing](https://stackoverflow.com/questions/44117788/performance-between-string-format-and-stringbuilder). I saw a precedent in the Logger class for using String.format though and went with that. Might be worth filing an issue to make a pass over everything and converting?

HEdingfield:
Thanks, @tarheel, I'll go through the 7 unnecessary instances of using `isEmpty()` and revert them in a future commit. Is it mainly the slight performance hit you're worried about here? I was moved to do it because I found 1 or 2 genuine bugs caused by not doing the `isEmpty()` check, since it's possible to have `""` in a config file. I like the idea of changing the getters to return null if the string is empty... I think that's similar to what you were suggesting?

LICENSE
moldover: Finally I can sleep again!

src/main/java/com/rcv/Main.java
moldover: I was wondering about that too.  Nice fix.

src/main/java/com/rcv/ElectionConfig.java
moldover: Also agree.  Good move.

src/main/java/com/rcv/ElectionConfig.java
moldover: If the time comes to break this up let's try and keep the UI model and group things by tab (only if it makes sense in the code of course)

src/main/java/com/rcv/Main.java
moldover: Makes sense for the Logger class at least.  I agree with the link there's places where we should keep String.format for readability.  I'll do that for the logger when I fix the other errors.

src/main/java/com/rcv/Main.java
```diff
@@ -151,97 +166,55 @@ static void executeTabulation(ElectionConfig config) {
...
-          encounteredProblemForThisSource = true;
+              Level.SEVERE, "Source file contains no CVRs: %s", source.getFilePath());
+          encounteredSourceProblem = true;
         }
-
-        if (config.isTabulateByPrecinctEnabled()
-            && (source.getPrecinctColumnIndex() == null || source.getPrecinctColumnIndex() < 0)) {
+        // add records to the master list
+        castVoteRecords.addAll(cvrs);
+      } catch (SourceWithUnrecognizedCandidatesException exception) {
```
moldover: This is a cool idea.  It could be used with #60 to flag problems with cvr and candidate configuration if/when we get to that.

moldover:
Nice work!  GUI logging is much cleaner and provides a nice high-level description for what's happening.   Config error feedback makes sense, and _I get a warm fuzzy feeling_ after fixing issues and seeing "Config validation successful."

src/main/java/com/rcv/Main.java
tarheel: Sorry, let me clarify. I'm fine with using `String.format` for the sake of readability. What I want to avoid is multiple statements appending to a String variable, especially in a loop, because that means we're actually creating a new String object each time.

tarheel:
> Thanks, @tarheel, I'll go through the 7 unnecessary instances of using isEmpty() and revert them in a future commit. Is it mainly the slight performance hit you're worried about here? I was moved to do it because I found 1 or 2 genuine bugs caused by not doing the isEmpty() check, since it's possible to have "" in a config file. I like the idea of changing the getters to return null if the string is empty... I think that's similar to what you were suggesting?

Zero concern about perf! It's more readability. I find it confusing when the code checks for a case that we already know should be impossible. If the consequence of our assumption being wrong is really bad, perhaps we add an `assert`, but otherwise I think it just muddles things for someone working on that code in the future.

I would prefer to catch all these issues at validation time. If the empty string isn't valid, disallow it then --  and whenever we reference the value after that point, assume it's not empty.

PR #143, opened 2018-08-30
Election to contest

### LINKED ISSUE #78: Clean up test folder
It would be nice to standardize the file names and clean up the test folder. I like the format date-location-office. Maybe we could consistently prepend cvr_* and config_* to those.

There also might be some test files in there that aren't needed anymore or, if they are, could also have more consistent names (e.g. "minimum-threshold-test-cvr.xlsx", "test_continue_tabulation_config.json", etc.).

### LINKED ISSUE #97: Reorganize / clean up config file format
As suggested in the [UI update Google doc](https://docs.google.com/document/d/1jQmhIoHNo0cNxD8h3xDHmiGyvxaUwbFi0Ieh3-W-2_U/edit?ts=5b5d375c#). @CalebKleppner suggested that we group the currently-ungrouped parameters as follows:

1. Headers and file names
  "contestName" required:  user types in name (UI displays existing data if any)
  "outputDirectory" required:  user types in name or navigates to directory 
  "tabulateByPrecinct":  click true or false.
  "date:" optional:  user types
  "jurisdiction" optional:  user types
  "office" optional:  user types

@moldover suggested we use a name like "outputSettings" or "reportSettings" (instead of HeadersAndFiles).

We'll also need to remember to reorganize and fix the config file documentation.

### LINKED ISSUE #107: Change "election" references to "contest"
Agree with @moldover in the [UI doc](https://docs.google.com/document/d/1jQmhIoHNo0cNxD8h3xDHmiGyvxaUwbFi0Ieh3-W-2_U/edit?ts=5b5d375c#) that "contest" is a better term than "election" for our tool if we want to be as general as possible.

Recommend a superficial refactor of all uses of the word "election" in the code and documentation to "contest".

@tarheel sound good?

HEdingfield:
Fixes #107, fixes #97, fixes #78.


src/main/java/com/rcv/RawContestConfig.java
```diff
@@ -30,78 +30,43 @@
...
-  // contest date
-  public String contestDate;
-  // should we report round-by-round results by precinct also?
-  public boolean tabulateByPrecinct;
-  // rules object
-  public ContestRules rules;
-  // list of all Candidates
-  public List<Candidate> candidates;
+  // contest settings object
+  public ContestSettings settings;
```
tarheel: The distinction between the words "config", "settings", and "rules" is not intuitive. Can we find a term that's more descriptive than "settings"? It seems like everything under "settings" is related to output. How about "outputSettings"?

HEdingfield: Done! Cowboy merging...

PR #153, opened 2018-09-02
Kill multiSeatTransferRule (fixes #150)

### LINKED ISSUE #150: remove multiSeatTransferRule param

tarheel:
Removing this from the UI makes it look a little janky, but you're making other changes there anyway, so I won't worry about it in this commit.


tarheel:
Yes, that's the only logic that's in place.

Issue #152, opened 2018-08-31
GUI redesign


HEdingfield:
* Get rid of the startup menu screen in favor of going straight to the config creator
* Move buttons to top above tabs; add buttons for new, load, tabulate, and quit
* Save button should read "Save..." to indicate that it'll summon a dialog
* A label at the top should communicate what's currently loaded (else say something like <new config>)
* Save listener (see #134) checks need to occur on new, load, tabulate, quit and alert user asking if they want to save first
* Add and Delete buttons need to be placed so that they are more intuitive (for CVR sources and possibly candidates)
* Delete Selected buttons should be disabled if nothing is selected
* Ability to drag and edit rows (e.g. for candidate permutations)

gngilbert:
Can't tell how much of this has been addressed.

HEdingfield:
This has mostly all been addressed and is superseded by #461 anyway. Moving any last unresolved items to that issue.

Issue #140, opened 2018-08-29
Improve the look of the GUI using CSS


HEdingfield:
Tutorials 25-27 in this playlist give some examples:
https://www.youtube.com/playlist?list=PL6gx4Cwl9DGBzfXLWLSYVy8EbTdpGbUIG

chughes297:
@freedomcounts Would this sort of update count as de minimis? Seems really valuable to do but if we can do it after certification and not worry about it, seems best to do after that cert.

freedomcounts:
This would probably be a de minimis.  If anyone disagrees let me know.

Keith Long, PMP
Freedom Counts, LLC
Keith.long@freedomcounts.com
843-540-1491

tarheel:
Yeah -- if anything at all qualifies as de minimis, it's something like this.

PR #157, opened 2018-09-04
GUI redesign

HEdingfield:
Began addressing #152:
* Got rid of the startup menu screen in favor of going straight to the config creator
* Moved buttons to top above tabs; added buttons for new, load, tabulate, and quit
* Buttons labeled with "..." to indicate that they'll summon dialogs
* Added label at the top to communicate what's currently loaded

Added CSS to begin addressing #140, including dark themes for text areas. Improved layout of Rules tab.


HEdingfield:
Are we cool with buttons that read, e.g. "Load..." instead of "Load Config..."? Not sure if it's clear what we're loading and saving with the short names like this.

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -375,20 +500,51 @@ private void loadConfig(ContestConfig config) {
...
+    Integer returnValue;
+    try {
+      if (textField.getText().isEmpty()) {
+        throw new IllegalArgumentException();
+      }
+      returnValue = Integer.valueOf(textField.getText());
+    } catch (Exception exception) {
+      Logger.log(
+          Level.WARNING,
+          "Integer required! Illegal value '%s' was replaced by '%s'",
```
tarheel: If you leave a field blank, you get this: “WARNING: Integer required! Illegal value '' was replaced by 'null’”

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -375,20 +500,51 @@ private void loadConfig(ContestConfig config) {
...
+        radioExhaustOnDuplicateCandidateTrue,
+        radioExhaustOnDuplicateCandidateFalse);
+    setToggleBoolean(
+        config.isTreatBlankAsUndeclaredWriteInEnabled(),
+        radioTreatBlankAsUndeclaredWriteInTrue,
+        radioTreatBlankAsUndeclaredWriteInFalse);
+  }
+
+  // TODO: this function assumes that if a field isn't provided in the config, it'll be false
+  private void setToggleBoolean(boolean condition, RadioButton radioTrue, RadioButton radioFalse) {
```
tarheel: I wonder if it would be easier to represent these boolean params with checkboxes instead of radios.

src/main/java/com/rcv/ContestConfig.java
```diff
@@ -35,6 +35,12 @@
 
 class ContestConfig {
 
+  // TODO: if any booleans are unspecified in config file, they default to false no matter what
```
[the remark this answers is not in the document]
HEdingfield: To make it so there's a way to set booleans to default to true if desired. I'll file it more clearly later as a minor issue.

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: Checkboxes would allow you to select more than one option, so that's a no-go. The radio button toggle group only allows one option to be selected at a time (so will automatically unselect any others).

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: Yeah, I noticed this too. Will file a minor issue to handle it.

HEdingfield:
Nothing I can do about the dialog boxes, unfortunately. That's purely a weird interaction between Java and MacOS native behavior.

tarheel:
I meant a single checkbox per boolean. That feels a little more intuitive. Not a big deal, though.

PR #159, opened 2018-09-10
User guide

moldover:
A user guide first cut.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,3 +26,6 @@ Jackson Databind 2.9.6
 
 ## Configuration
 Config file documentation is available [here](src/main/resources/config_file_documentation.txt).
+
+## User Guide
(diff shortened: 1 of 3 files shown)


tarheel:
Agree on both counts!

HEdingfield:
That's funny, cause I was thinking the other txt would get phased out and
absorbed completely into the UI in the future. Could you at least put a
link to the user guide in the README for this PR like I did for the other
txt?

CalebKleppner:
Quick thoughts:  first draft of user's guide looks very good.  No need 
to belabor it until ... we need to belabor it.  I don't feel strongly 
about whether we should combine user's guide with readme info, so you 
all should do it how you all think it makes sense.

Caleb Kleppner
MK Election Services, LLC
203-376-4080

PR #161, opened 2018-09-15
fixes in multi-seat threshold logic (partially addresses #147)

tarheel:
This gets us closer to matching the official results. There are still some inconsistencies we'll need to sort out regarding the arithmetic and which operations should take place in each round, but the changes here are a step in the right direction.

### DIFF
--- a/src/main/java/com/rcv/Tabulator.java
+++ b/src/main/java/com/rcv/Tabulator.java
@@ -23,6 +23,7 @@ package com.rcv;
 
 import com.rcv.CastVoteRecord.VoteOutcomeType;
 import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


src/main/java/com/rcv/Tabulator.java
```diff
@@ -360,25 +361,30 @@ private CandidateStatus getCandidateStatus(String candidate) {
...
+      for (BigDecimal tally : currentRoundTallyToCandidates.keySet()) {
+        if (tally.compareTo(winningThreshold) > 0) {
+          // we have winner(s)
+          List<String> winningCandidates = currentRoundTallyToCandidates.get(tally);
+          selectedWinners.addAll(winningCandidates);
         }
       }
     }
+
+    for (String winner : selectedWinners) {
```
HEdingfield: Yay, killed a silly nested for loop!

PR #162, opened 2018-09-15
remove the overvote options we don't need to support

### LINKED ISSUE #151: review overvote options
Caleb thinks we only need these three:
- always exhaust
- always skip to next rank
- exhaust if multiple continuing

He'll review with George.

tarheel:
Fixes #151

### DIFF
--- a/src/main/java/com/rcv/Tabulator.java
+++ b/src/main/java/com/rcv/Tabulator.java
@@ -214,8 +214,7 @@ class Tabulator {
     }
 
     if (config.getTiebreakMode() == TieBreakMode.GENERATE_PERMUTATION) {
-      Logger.log(
-          Level.INFO, "Randomly generated candidate permutation for tie-breaking:");
+      Logger.log(Level.INFO, "Randomly generated candidate permutation for tie-breaking:");
       // candidateID indexes over all candidates in ordered list
       for (String candidateID : config.getCandidatePermutation()) {
         Logger.log(Level.INFO, "%s", candidateID);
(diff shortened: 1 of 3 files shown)


src/main/resources/config_file_documentation.txt
```diff
@@ -149,7 +149,7 @@ Config file must be valid JSON format. Examples can be found under the test fold
         default: false
 
       "overvoteLabel" optional
-        label used in the CVR to denote an overvote; if this parameter is present overvoteRule must be either "alwaysSkipToNextRank" or "exhaustImmediately" (because the other options rely on knowing which specific candidates were involved in each overvote)
+        label used in the CVR to denote an overvote; if this parameter is present overvoteRule must be either "alwaysSkipToNextRank" or "exhaustImmediately" (because the other option, exhaustIfMultipleContinuing, relies on knowing which specific candidates were involved in each overvote)
```
HEdingfield: Also need to fix "overvoteRule" section in doc above to eliminate the other options.

src/main/java/com/rcv/Tabulator.java
```diff
@@ -871,10 +849,7 @@ void doAudit(List<CastVoteRecord> castVoteRecords) {
   enum OvervoteRule {
     EXHAUST_IMMEDIATELY("exhaustImmediately"),
     ALWAYS_SKIP_TO_NEXT_RANK("alwaysSkipToNextRank"),
-    EXHAUST_IF_ANY_CONTINUING("exhaustIfAnyContinuing"),
```
HEdingfield: Do a project-wide search for all 3 of these modes and make sure you don't miss any. I see some config files that still use them, so we'll want to replace those.

tarheel: It was just the config that tests validation of invalid params. But I've updated it anyway for clarity.

Issue #155, opened 2018-09-03
tabulation tally optimizations


tarheel:
A couple changes that will make the code run faster:
1. In rounds 2+, if a CVR's selection from last round is still continuing, we can automatically select it again.
2. In multi-seat contests, we don't need to recompute winners' totals in subsequent rounds, because they'll never change.

HEdingfield:
Anything left to do on this, or did #164 resolve it?

tarheel:
#164 did half of it (if the candidate a CVR selected last round is continuing in this round, automatically select them again). Still need to do this: "In multi-seat contests, we don't need to recompute winners' totals in subsequent rounds, because they'll never change."

HEdingfield:
One thought occurred to me on this: even if a candidate is declared a winner, isn't it possible for them to get more votes in subsequent rounds? And, if so, shouldn't these be added to their totals just for the sake of accuracy in the final outcome?

tarheel:
Good question, but that's not how multi-seat tabulation works. Once a candidate wins and their surplus is transferred, they're no longer considering a continuing candidate, so they don't accumulate any additional votes. As you imply, this does make the final outcome a bit unintuitive; you can't draw any conclusions by comparing the final totals of the different winners. (@CalebKleppner and @gngilbert: please correct me if I said anything wrong here.)

tarheel:
How does the logic seem inaccurate? Or, to put it another way: how else could you design the process? Allowing the candidate to continue to accumulate votes would contradict the purpose of transferring their surplus vote when they first exceeded the winning threshold.

I suppose you could do something even more complicated: allow them to accumulate surplus votes, but fully recalculate their surplus transfers each round based on their new total. That strikes me as perhaps more intuitively fair, at least in the abstract.

CalebKleppner:
And even the order in which a candidate reaches the threshold doesn't 
really mean anything.  There are ways of assessing overall level of 
support for candidates but order of election is a pretty arbitrary one.

Caleb Kleppner
MK Election Services, LLC
203-376-4080

Issue #146, opened 2018-08-31
change audit file format


tarheel:
New format:
- First show how each CVR voted in the first round.
- Then, for each subsequent round, show any transfers.

This will allow us to generate the audit file on the fly instead of storing round-by-round data in each CVR object and then writing out the whole file after the tabulation completes.

PR #164, opened 2018-09-16
audit log improvements (#146)

moldover:
only log cvr data when it changes
log during tabulation (instead of after)
remove VoteOutcome cache from cvr
flush and close audit fileHandler

add check in Tabulate main loop to continue if the cvr counts to the
same continuing candidate as the previous round (issue #155)

### DIFF
--- a/src/main/java/com/rcv/CastVoteRecord.java
+++ b/src/main/java/com/rcv/CastVoteRecord.java
@@ -23,12 +23,12 @@ package com.rcv;
 import java.math.BigDecimal;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
+import java.util.logging.Level;
 import javafx.util.Pair;
 
 class CastVoteRecord {
@@ -41,8 +41,6 @@ class CastVoteRecord {
   private final String precinct;
   // container for ALL CVR data parsed from the source CVR file
   private final List<String> fullCVRData;
-  // contains what happened to this CVR in each round
-  private final List<VoteOutcome> roundOutcomes = new LinkedList<>();
   // records winners to whom some fraction of this vote has been allocated
   private final Map<String, BigDecimal> winnerToFractionalValue = new HashMap<>();
   // map of round to all candidates selected for that round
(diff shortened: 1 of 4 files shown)


src/main/java/com/rcv/CastVoteRecord.java
```diff
@@ -72,24 +70,60 @@
...
+
+    // add fractional transfer value of the vote if it is fractional
+    if (fractionalTransferValue != null && !fractionalTransferValue.equals(BigDecimal.ONE)) {
+      logStringBuilder.append(" [value] ").append(fractionalTransferValue.toString());
+    }
+
+    // add complete data for round 1 only
+    if (round == 1) {
+      logStringBuilder.append(" [Raw Data] ");
+      logStringBuilder.append(fullCVRData).append(" ");
```
tarheel: that last " " seems unnecessary, yeah?

src/main/java/com/rcv/Tabulator.java
```diff
@@ -688,6 +688,22 @@ private OvervoteDecision getOvervoteDecision(Set<String> candidateIDSet) {
 
     // CVR indexes over the cast vote records to count votes for continuing candidateIDs
     for (CastVoteRecord cvr : castVoteRecords) {
+      // see if this cvr recipient is a continuing candidate
+      // in this case we can skip logic to determine who the cvr will count for
+      if (!cvr.isExhausted() &&
+          !(cvr.getCurrentRecipientOfVote() == null) &&
```
tarheel: `cvr.getCurrentRecipientOfVote() != null`

src/main/java/com/rcv/Tabulator.java
```diff
@@ -688,6 +688,22 @@ private OvervoteDecision getOvervoteDecision(Set<String> candidateIDSet) {
 
     // CVR indexes over the cast vote records to count votes for continuing candidateIDs
     for (CastVoteRecord cvr : castVoteRecords) {
+      // see if this cvr recipient is a continuing candidate
```
tarheel: You can make this comment a little clearer. Suggestion: "If this CVR was assigned to a candidate last round and that candidate is still continuing, we can safely assume that the CVR should still be assigned to that candidate in this round."

src/main/java/com/rcv/Tabulator.java
```diff
@@ -688,6 +688,22 @@ private OvervoteDecision getOvervoteDecision(Set<String> candidateIDSet) {
...
+      // in this case we can skip logic to determine who the cvr will count for
+      if (!cvr.isExhausted() &&
+          !(cvr.getCurrentRecipientOfVote() == null) &&
+          isCandidateContinuing(cvr.getCurrentRecipientOfVote())) {
+
+        // this vote stays with it's current recipient
```
tarheel: its

src/main/java/com/rcv/Tabulator.java
```diff
@@ -752,27 +775,29 @@ private OvervoteDecision getOvervoteDecision(Set<String> candidateIDSet) {
           // If this fails, it means the code failed to handle an overvote with multiple
           // continuing candidates.
           assert selectedCandidate == null;
-          // we found a continuing candidate, so increase their tally by 1
```
tarheel: oops, yeah, that comment was inaccurate

src/main/java/com/rcv/Tabulator.java
```diff
@@ -688,6 +688,22 @@ private OvervoteDecision getOvervoteDecision(Set<String> candidateIDSet) {
...
+          !(cvr.getCurrentRecipientOfVote() == null) &&
+          isCandidateContinuing(cvr.getCurrentRecipientOfVote())) {
+
+        // this vote stays with it's current recipient
+        incrementTallies(roundTally,
+            cvr.getFractionalTransferValue(),
+            cvr.getCurrentRecipientOfVote(),
+            roundTallyByPrecinct,
+            cvr.getPrecinct());
+        continue;
```
tarheel: Might be worth adding a comment here explaining why we don't call `logRoundOutcome`.

src/main/java/com/rcv/Tabulator.java
```diff
@@ -843,6 +871,29 @@ private void incrementTally(
...
 
+  // function: incrementTallies
+  // purpose: transfer vote to round tally and (if valid) the precinct round tally
+  // param: tally is a round tally that we're in the process of computing
+  // param: cvr is a single cast vote record
+  // param: selectedCandidateID is the candidate this CVR's vote is going to in this round
+  // param: roundTallyByPrecinct map of precinct IDs to roundTallies
+  // param: precinct ID of precinct for current CVR
+  private void incrementTallies(
+      Map<String, BigDecimal> tally,
```
tarheel: I would rename this to `roundTally` for clarity

src/main/java/com/rcv/Tabulator.java
```diff
@@ -843,6 +871,29 @@ private void incrementTally(
...
+      BigDecimal fractionalTransferValue,
+      String selectedCandidateID,
+      Map<String, Map<String, BigDecimal>> roundTallyByPrecinct,
+      String precinct) {
+    // transfer vote value to round tally
+    incrementTally(tally, fractionalTransferValue, selectedCandidateID);
+    // if enabled and there is a valid precinct string transfer vote valute to precinct tally
```
tarheel: s/valute/value

src/main/java/com/rcv/Tabulator.java
```diff
@@ -843,6 +871,29 @@ private void incrementTally(
...
+  // purpose: transfer vote to round tally and (if valid) the precinct round tally
+  // param: tally is a round tally that we're in the process of computing
+  // param: cvr is a single cast vote record
+  // param: selectedCandidateID is the candidate this CVR's vote is going to in this round
+  // param: roundTallyByPrecinct map of precinct IDs to roundTallies
+  // param: precinct ID of precinct for current CVR
+  private void incrementTallies(
+      Map<String, BigDecimal> tally,
+      BigDecimal fractionalTransferValue,
+      String selectedCandidateID,
```
tarheel: @HEdingfield and I would both prefer to refer to candidates just as candidates (as in selectedCandidate) rather than candidateIDs in the code, since ID kind of implies we're passing around numbers.

src/main/java/com/rcv/Tabulator.java
```diff
@@ -752,27 +775,29 @@ private OvervoteDecision getOvervoteDecision(Set<String> candidateIDSet) {
           // If this fails, it means the code failed to handle an overvote with multiple
           // continuing candidates.
           assert selectedCandidate == null;
```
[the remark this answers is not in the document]
HEdingfield: Yeah, that's what I thought, but it looks like there's no error handling logic in here, so does it just crash the entire process?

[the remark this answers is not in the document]
moldover: https://stackoverflow.com/questions/3018683/what-does-the-assert-keyword-do

assert needs to be enabled at launch by setting a flag  (-ea in the VM options in the run configuration) in which case yes, it would crash the application here.  I think the point with assertion is to catch programming logic errors and fix them, unlike like exceptions which would have error handling logic.  But it does seem like we should be running with the flag for it to be of any use.  I would have thought it was enabled by default.

PR #166, opened 2018-09-16
GUI save listener

### LINKED ISSUE #134: Create save listener for GUI
Modifying config file: after making changes, pressing validate doesn't save it, so it's easy to click menu button and lose all changes; probably need to make save listener after all.

HEdingfield:
Fixes #134. Progress on #152. Minor bug fixes.

### DIFF
--- a/src/main/java/com/rcv/GuiConfigController.java
+++ b/src/main/java/com/rcv/GuiConfigController.java
@@ -184,11 +184,11 @@ public class GuiConfigController implements Initializable {
     }
   }
 
-  private void loadFile(File loadFile) {
-    GuiContext.getInstance().setConfig(Main.loadContestConfig(loadFile.getAbsolutePath()));
+  private void loadFile(File fileToLoad) {
+    GuiContext.getInstance().setConfig(Main.loadContestConfig(fileToLoad.getAbsolutePath()));
     if (GuiContext.getInstance().getConfig() != null) {
       loadConfig(GuiContext.getInstance().getConfig());
-      labelCurrentlyLoaded.setText("Currently loaded: " + loadFile.getAbsolutePath());
+      labelCurrentlyLoaded.setText("Currently loaded: " + fileToLoad.getAbsolutePath());
     }
   }
 
@@ -203,10 +203,10 @@ public class GuiConfigController implements Initializable {
       fc.getExtensionFilters().add(new ExtensionFilter("JSON files", "*.json"));
       fc.setTitle("Load Config");
 
-      File loadFile = fc.showOpenDialog(null);
-      if (loadFile != null) {
-        selectedFile = loadFile;
-        loadFile(loadFile);
+      File fileToLoad = fc.showOpenDialog(null);
+      if (fileToLoad != null) {
+        selectedFile = fileToLoad;
+        loadFile(fileToLoad);
       }
     }
   }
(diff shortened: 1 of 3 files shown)


src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -168,33 +176,42 @@
...
+    }
   }
 
-  public void buttonLoadConfigClicked() {
-    FileChooser fc = new FileChooser();
-    if (selectedFile == null) {
-      fc.setInitialDirectory(new File(System.getProperty("user.dir")));
-    } else {
-      fc.setInitialDirectory(new File(selectedFile.getParent()));
+  private void loadFile(File loadFile) {
```
tarheel: Feels a little confusing for the function and its only param to share the same name.

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -204,10 +221,20 @@ public void buttonSaveClicked() {
     }
     fc.getExtensionFilters().add(new ExtensionFilter("JSON files", "*.json"));
     fc.setTitle("Save Config");
+    return fc.showSaveDialog(null);
+  }
 
-    File saveFile = fc.showSaveDialog(null);
+  private void saveFile(File saveFile) {
+    JsonParser.createFileFromRawContestConfig(saveFile, createRawContestConfig());
+    // Necessary to keep GUI aligned when default values subbed in if invalid during save process
```
tarheel: Not quite sure what this means.

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -386,6 +417,19 @@ public LocalDate fromString(String string) {
...
+    try {
+      emptyConfigString =
+          new ObjectMapper()
+              .writer()
+              .withDefaultPrettyPrinter()
+              .writeValueAsString(createRawContestConfig());
+    } catch (JsonProcessingException exception) {
+      Logger.log(
+          Level.WARNING,
+          "Unable to set emptyConfigString, but everything should work fine anyway!\n%s",
```
tarheel: If that's the case, then what's the purpose of doing it? I guess if this fails, the worst that can happen is we ask if the user wants to save even though the config is empty?

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -463,6 +507,88 @@ private void clearConfig() {
...
+    } catch (JsonProcessingException exception) {
+      Logger.log(
+          Level.WARNING,
+          "Unable tell if saving is necessary, but everything should work fine anyway! Prompting for save just in case...\n%s",
+          exception.toString());
+    }
+    return needsSaving;
+  }
+
+  private boolean checkForSaveAndContinue() {
```
tarheel: If I'm reading this method correctly... the only way you can end up with willContinue = false is if the user clicks "Save" but then doesn't actually save?

src/main/resources/config_file_documentation.txt
```diff
@@ -58,7 +58,7 @@ Config file must be valid JSON format. Examples can be found under the test fold
 
       "idColumnIndex" optional
         index of the column that contains the unique ID for each CVR
-        example: 0
+        example: 1
```
tarheel: thanks

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: Renaming this to `fileToLoad` and `saveFile` to `fileToSave`.

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: Clarifying comment. Basically if you save a file with an invalid value (like null for number of decimal places), the save procedure replaces it with the default value. Unless we reload the file, the GUI will still have a null value in that field. The reload makes sure the default value populates the text field in this situation.

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: Yes, that's the worst that could happen at the moment. Kind of wanted to keep these warnings in here so we could tell if something is breaking. I don't anticipate these will get triggered much at all anyway, so it shouldn't be too confusing to the user.

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: Exactly, if they close the window or press cancel. I think it's easier to check this condition the way I wrote it, rather than trying to capture those other cases.

Issue #160, opened 2018-09-10
in macOS, app doesn't quit properly if file dialog is open


tarheel:
If I have the load or save dialog open and close the main window, the dialog stays open. If I try to quit the app, nothing happens, even after I close the dialog. I have to kill the app by clicking the stop button in IntelliJ.

Similarly, if I have the dialog open and select java > Quit, it closes the main window, but not the dialog or the app as a whole, and I end up having to kill it.

HEdingfield:
Note: this appears to be a MacOS issue, as I'm unable to repro it in
Windows; everything works as expected there. Will test in Linux and report
back.

moldover:
Fixed.

Issue #135, opened 2018-08-29
GUI closing confirmation pop-up


HEdingfield:
Should probably add for both main menu and config creator, e.g.:
https://www.youtube.com/watch?v=ZuHcl5MmRck&index=8&t=0s&list=PL6gx4Cwl9DGBzfXLWLSYVy8EbTdpGbUIG

Probably ties into #134.

HEdingfield:
Closing the GUI when in the middle of tabulating needs to kill all threads.

chughes297:
Added deliverable because I expect it's likely people will accidentally attempt to close the app frequently, would like to keep that from happening.

HEdingfield:
Closed with #320.

PR #172, opened 2018-09-20
File chooser

moldover:
@HEdingfield I just pushed a pretty hacky fix for #160.  Apparently it's "only" an issue for Mac?
https://bugs.openjdk.java.net/browse/JDK-8088859

I made 2 changes:
make file chooser modal -- I think this is better UX and will simplify logic / other potential issues
quit app when filechooser is open -- addresses #160

Was wondering if you could pull file_chooser branch and run it on your PC and see if anything untoward happens.  If it seems ok I will clean things up (right now is just prototype).  Thanks!

### DIFF
--- a/src/main/java/com/rcv/GuiApplication.java
+++ b/src/main/java/com/rcv/GuiApplication.java
@@ -16,14 +16,12 @@
 
 package com.rcv;
 
-import java.util.logging.Level;
 import javafx.application.Application;
 import javafx.application.Platform;
 import javafx.event.EventHandler;
 import javafx.fxml.FXMLLoader;
 import javafx.scene.Parent;
 import javafx.scene.Scene;
-import javafx.stage.FileChooser;
 import javafx.stage.Stage;
 import javafx.stage.WindowEvent;
 
@@ -34,22 +32,12 @@ public class GuiApplication extends Application {
     Parent root = FXMLLoader.load(getClass().getResource("/GuiConfigLayout.fxml"));
     window.setTitle("Universal RCV Tabulator");
     window.setScene(new Scene(root));
+    // cache main window so we can parent file choosers to it
     GuiContext context = GuiContext.getInstance();
-    // cache this so we can parent file chooser to it
-    context.mainWindow = window;
-    EventHandler<WindowEvent> windowEventEventHandler = event -> {
-      // do something
-      Logger.log(Level.INFO, "bla");
-      FileChooser fileChooser = GuiContext.getInstance().fileChooser;
-      if (fileChooser != null) {
-        try {
-          Platform.exit();
-        } catch (Exception e) {
-          e.printStackTrace();
-        }
-      }
-    };
-    window.setOnCloseRequest(windowEventEventHandler);
+    context.setMainWindow(window);
+    // workaround for https://bugs.openjdk.java.net/browse/JDK-8088859
+    EventHandler<WindowEvent> onCloseHandler = event -> Platform.exit();
+    window.setOnCloseRequest(onCloseHandler);
     window.show();
   }
 
--- a/src/main/java/com/rcv/GuiApplication.java
+++ b/src/main/java/com/rcv/GuiApplication.java
@@ -16,11 +16,16 @@
 
 package com.rcv;
 
+import java.util.logging.Level;
 import javafx.application.Application;
+import javafx.application.Platform;
+import javafx.event.EventHandler;
 import javafx.fxml.FXMLLoader;
 import javafx.scene.Parent;
 import javafx.scene.Scene;
+import javafx.stage.FileChooser;
 import javafx.stage.Stage;
+import javafx.stage.WindowEvent;
 
 public class GuiApplication extends Application {
 
@@ -29,6 +34,23 @@ public class GuiApplication extends Application {
     Parent root = FXMLLoader.load(getClass().getResource("/GuiConfigLayout.fxml"));
     window.setTitle("Universal RCV Tabulator");
     window.setScene(new Scene(root));
+    GuiContext context = GuiContext.getInstance();
+    // cache this so we can parent file chooser to it
+    context.mainWindow = window;
+    EventHandler<WindowEvent> windowEventEventHandler = event -> {
+      // do something
+      Logger.log(Level.INFO, "bla");
+      FileChooser fileChooser = GuiContext.getInstance().fileChooser;
+      if (fileChooser != null) {
+        try {
+          Platform.exit();
+        } catch (Exception e) {
+          e.printStackTrace();
+        }
+      }
+    };
+    window.setOnCloseRequest(windowEventEventHandler);
     window.show();
   }
+
 }
--- a/src/main/java/com/rcv/GuiConfigController.java
+++ b/src/main/java/com/rcv/GuiConfigController.java
@@ -174,6 +174,9 @@ public class GuiConfigController implements Initializable {
   public void buttonLoadConfigClicked() {
     if (checkForSaveAndContinue()) {
       FileChooser fc = new FileChooser();
+
+      GuiContext.getInstance().fileChooser = fc;
+
       if (selectedFile == null) {
         fc.setInitialDirectory(new File(System.getProperty("user.dir")));
       } else {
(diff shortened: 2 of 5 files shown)


src/main/java/com/rcv/GuiContext.java
```diff
@@ -16,13 +16,19 @@
...
+
 class GuiContext {
 
   // context instance
   private static final GuiContext INSTANCE = new GuiContext();
   // currently-loaded tabulator config
   private ContestConfig config;
 
+  public Stage mainWindow;
+  public FileChooser fileChooser;
```
HEdingfield: You're going to want to put these two inside the GuiContext; basically replicate how config works currently in here.

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -174,15 +174,20 @@ private void loadFile(File fileToLoad) {
...
+      GuiContext.getInstance().fileChooser = fc;
+
       if (selectedFile == null) {
         fc.setInitialDirectory(new File(System.getProperty("user.dir")));
       } else {
         fc.setInitialDirectory(new File(selectedFile.getParent()));
       }
       fc.getExtensionFilters().add(new ExtensionFilter("JSON files", "*.json"));
       fc.setTitle("Load Config");
+      File fileToLoad = fc.showOpenDialog(GuiContext.getInstance().mainWindow);
```
HEdingfield: Instead of needing to do all that jazz in GuiContext to put the window here, I'm pretty sure we can reference the window that contains the button being clicked. Maybe it's cleaner to do it via GuiContext though? Not sure, would need to play around with the other option and see how it looks.

src/main/java/com/rcv/GuiApplication.java
```diff
@@ -29,6 +34,23 @@ public void start(Stage window) throws Exception {
     Parent root = FXMLLoader.load(getClass().getResource("/GuiConfigLayout.fxml"));
     window.setTitle("Universal RCV Tabulator");
     window.setScene(new Scene(root));
+    GuiContext context = GuiContext.getInstance();
+    // cache this so we can parent file chooser to it
+    context.mainWindow = window;
+    EventHandler<WindowEvent> windowEventEventHandler = event -> {
```
HEdingfield: Is this even necessary now that the window is properly modal?

HEdingfield:
One last thing I'm curious about: the bug you linked says it applies to JDK 8 and 9. I wonder if it's fixed in 10, and maybe Louis was using 9 to run it?

src/main/java/com/rcv/GuiConfigController.java
moldover: dig!

src/main/java/com/rcv/GuiApplication.java
moldover: I'm going to try just always doing this aggressive Platform.exit behavior (regardless if the Chooser is open) and see if that causes any problems.  If not, there is no need to cache the Chooser in the GUIContext, and with your other suggestion below there would be no need to change the GUIContext at all yet.

tarheel:
> One last thing I'm curious about: the bug you linked says it applies to JDK 8 and 9. I wonder if it's fixed in 10, and maybe @tarheel was using 9 to run it?

I just tested in 10 and it still reproed.

src/main/java/com/rcv/GuiConfigController.java
moldover: unfortunately there is no easier path from within the event handler to obtain the root window so I went ahead with plan A.

src/main/java/com/rcv/GuiApplication.java
moldover: this seems to work fine.

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -156,7 +156,7 @@
 
   public void buttonNewConfigClicked() {
     if (checkForSaveAndContinue()) {
-      Logger.log(Level.INFO, "Creating new config.");
+      Logger.log(Level.INFO, "Creating new contest.");
```
HEdingfield: Hmm, don't agree with this change. I could settle for "contest config" though. I suggest reverting these three "config" => "contest" changes and filing an issue so we can treat this more comprehensively if you think it's important. The problem is that now we have these three references changed, but there are cases where "config" is used in messages communicated to the user, title bars, etc.

src/main/java/com/rcv/GuiContext.java
```diff
@@ -43,4 +41,12 @@ ContestConfig getConfig() {
   void setConfig(ContestConfig config) {
     this.config = config;
   }
+
+  Stage getMainWindow() {
+    return this.mainWindow;
```
HEdingfield: `return mainWindow;`

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -423,7 +417,7 @@ private void setTextFieldToInteger(TextField textField, Integer value) {
   }
 
   private void setDefaultValues() {
-    labelCurrentlyLoaded.setText("Currently loaded: <new config>");
+    labelCurrentlyLoaded.setText("Currently loaded: <New Contest>");
```
HEdingfield: Suggest reversion for now.

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -543,7 +537,7 @@ private boolean checkForSaveAndTabulate() {
       Alert alert =
           new Alert(
               AlertType.WARNING,
-              "You must either save your changes before continuing or load a new config!",
+              "You must either save your changes before continuing or load a new contest!",
```
HEdingfield: Suggest reversion for now.

src/main/java/com/rcv/GuiApplication.java
```diff
@@ -29,6 +32,13 @@ public void start(Stage window) throws Exception {
     Parent root = FXMLLoader.load(getClass().getResource("/GuiConfigLayout.fxml"));
     window.setTitle("Universal RCV Tabulator");
     window.setScene(new Scene(root));
+    // cache main window so we can parent file choosers to it
```
HEdingfield: "parent pop-up windows" in general, I'd say.

src/main/java/com/rcv/GuiApplication.java
```diff
@@ -29,6 +32,13 @@ public void start(Stage window) throws Exception {
...
+    GuiContext context = GuiContext.getInstance();
+    context.setMainWindow(window);
+    // workaround for https://bugs.openjdk.java.net/browse/JDK-8088859
+    EventHandler<WindowEvent> onCloseHandler = event -> Platform.exit();
```
HEdingfield: This is great! I think it'll help with #135 as well.

src/main/java/com/rcv/Logger.java
```diff
@@ -107,7 +107,10 @@ static void setup() throws IOException {
   static void addTabulationFileLogging(String outputPath) throws IOException {
     // create file handler at FINE level (we have detailed audit logging we want to capture)
     // and use our custom formatter
-    tabulationHandler = new FileHandler(outputPath);
+    tabulationHandler = new FileHandler(outputPath,
```
HEdingfield: A couple issues here:
1. I'm getting log files with .0 and .1 at the end, so they're unrecognized formats. I think you're going to want to use the %g.log thing like we did with the execution logs to avoid this.
2. For some reason I'm getting logs that start in file 1 and end in file 0 when they produce two files...

src/main/java/com/rcv/GuiConfigController.java
moldover: Sure I can change them back.  Generally I think we should avoid using terms like "config" with the user which are only used by programmers but it doesn't need to happen now.

HEdingfield:
Ping on this. Also, can we kill the `logging_updates` branch now? Seems like those changes were rolled into this one.

src/main/java/com/rcv/Logger.java
moldover: Yeah the %g naming is better, I'll change that one.  for # 2 that may be the expected behavior.  When file 0 gets to 50mb it moves to file 1 and when file 1 gets to 50mb it will move to file 0.

HEdingfield: Agree that item 2 can occur with the execution log because we're limiting it to only 2 files; but the tabulation log is limited to 1000 files, so it definitely shouldn't go from 1 to 0.

Another minor thing that would be nice to fix (though not sure if it's possible: if there's only 1 file, it shouldn't have 0 or 1 in the filename).

moldover: Gotcha.  So turns out that IS the expected behavior.  It makes sense when you consider the typical use case of looking at the most recent log file first, and then working your way back in time (https://www.youtube.com/watch?v=vHxFa08XE3w) to find something.  Our use case is the reverse: we have a specific starting point, and want to consume the data from that point forward in time.  We should roll this conversation into the next iteration on audit log format.

PR #173, opened 2018-09-23
fixes to results spreadsheet percentages (progress on  #147)

tarheel:
A few changes:
- Adjusted the precision setting in the config to match the official process.
- Fixed a rounding bug in our percentage calculation: if you divide first and then multiply by 100, you end up losing some precision. Multiplying first and then dividing fixes it.
- Changed the logic for calculating percentages: in multi-seat contests, the divisor should be the first round vote total, not the current round vote total.
- A bunch of lines that I didn't modify got reformatted in nitpicky ways.


tarheel:
On a separate note: I noticed that when I save a config using the UI, it adds a space before every colon in the JSON. Is that happening to you guys?

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -367,8 +369,11 @@ private void generateSummarySpreadsheet(
 
       // Exhausted votes as percentage of ALL votes (note: this differs from the candidate vote
       // percentages which are percentage of ACTIVE votes for the given round.
```
HEdingfield: "round)."

test_data/config_2013_minneapolis_park.json
```diff
@@ -53,7 +53,7 @@
     "maxRankingsAllowed": 3,
     "maxSkippedRanksAllowed": 2,
     "numberOfWinners": 3,
-    "decimalPlacesForVoteArithmetic": 5,
+    "decimalPlacesForVoteArithmetic": 4,
```
HEdingfield: Also fix for "config_2015_portland_mayor_multi.json" and "config_test_continue_tabulation.json"?

HEdingfield:
@tarheel yes, that occurs because we use `.withDefaultPrettyPrinter()` in `JsonParser`. Seemed weird at first, but it really just makes things a lot nicer and more consistent overall.

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -264,6 +264,8 @@ private void generateSummarySpreadsheet(
       roundPercentageCell.setCellValue("% of vote");
     }
 
+    BigDecimal totalActiveVotesFirstRound = totalActiveVotesPerRound.get(1);
```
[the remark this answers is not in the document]
tarheel: This map is indexed by actual round number, so 1 is correct.

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -309,13 +311,15 @@ private void generateSummarySpreadsheet(
...
-        BigDecimal fraction = config.divide(thisRoundTally, totalActiveVotes);
-        // percentage of active votes
+        // total active votes for %: use current round for single-winner, first round for multi
+        BigDecimal totalActiveVotes =
+            config.getNumberOfWinners() > 1
+                ? totalActiveVotesFirstRound
+                : totalActiveVotesPerRound.get(dataUseRound);
         BigDecimal percentage =
-            fraction.signum() == 1 ? fraction.multiply(new BigDecimal(100)) : BigDecimal.ZERO;
+            thisRoundTally.signum() == 1
```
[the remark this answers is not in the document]
tarheel: Agreed.

test_data/config_2013_minneapolis_park.json
tarheel: This is only relevant for multi-seat contests, and I think it varies by jurisdiction. We can update other configs if/when we determine it's appropriate.

Issue #169, opened 2018-09-17
round number confusion


tarheel:
It looks like there's a bit of disagreement about which "round" an event happened in.

If you look at the [2013 Minneapolis Park results](http://vote.minneapolismn.gov/results/2013/2013-park-board-at-large-tab), the first eliminations and transfers are shown happening in round 2.

In the [2018 Maine results](https://www.maine.gov/sos/cec/elec/results/2018/democraticgovfinal.xls), the first eliminations and transfers are instead associated with round 1.

The vote totals shown are the same either way -- round 1 shows first preferences, and round 2 shows updated totals after the first set of eliminations -- but I'm not sure which round number we should use when we report when an elimination happened. Is it the round before the totals are updated with the transfers, or the round in which they're updated?

We're actually not even being consistent ourselves: in our audit log, we say that the first eliminations happen in round 1, but in our XLSX output, we associate these eliminations and transfers with round 2.

moldover:
I just "fixed" the summary spreadsheets so they are consistent with Minneapolis and our audit log.  I agree this is more sensible, although I think a better format would not combine actions with counts in a single column:

Counts 1   Actions 1  Counts 2  Actions 2
 10                                15            winner
 5                eliminated  0 
 9                                   9            eliminated

moldover:
Given that we don't display transfers on the summary report anymore, and the actions are consistent, can we close this issue, or is there something more to do here?

tarheel:
I'll take this and do one more pass over everything to confirm that it's consistent.

tarheel:
@moldover: OK, yeah, it seems like we're being consistent now.

PR #175, opened 2018-09-24
Optimize past winners (fixes #155)

### LINKED ISSUE #155: tabulation tally optimizations

tarheel:
I tried a few different approaches, and this was the cleanest solution I could find. I was hoping to avoid counting CVRs in this method at all and just rely on the logic in the regular tallying, but that approach ends up complicating the regular tallying code more than I'm comfortable. The amount of duplicated logic here seems like a decent compromise.

I also fixed the precinct logic in this method. It was definitely not correct before.

Also, man, I really need to address #169 soon to preserve my own sanity.

### DIFF
--- a/src/main/java/com/rcv/Tabulator.java
+++ b/src/main/java/com/rcv/Tabulator.java
@@ -287,19 +287,16 @@ class Tabulator {
         Map<String, BigDecimal> winnerToFractionalValue = cvr.getWinnerToFractionalValue();
         for (String winner : winnerToFractionalValue.keySet()) {
           if (!winnersRequiringComputation.contains(winner)) {
-            continue; // this is someone who just won this round, so we can skip them
+            continue;
           }
-          // the fractional value we should use when incrementing
           BigDecimal fractionalTransferValue = winnerToFractionalValue.get(winner);
+
           incrementTally(roundTally, fractionalTransferValue, winner);
-          for (String precinct : precinctRoundTallies.keySet()) {
-            // all the tallies for the given precinct
-            Map<Integer, Map<String, BigDecimal>> roundTalliesForPrecinct =
-                precinctRoundTallies.get(precinct);
-            // the tally for the current round for the precinct
-            Map<String, BigDecimal> roundTallyForPrecinct =
-                roundTalliesForPrecinct.get(currentRound);
-            incrementTally(roundTallyForPrecinct, fractionalTransferValue, winner);
+          if (config.isTabulateByPrecinctEnabled() && cvr.getPrecinct() != null) {
+            incrementTally(
+                precinctRoundTallies.get(cvr.getPrecinct()).get(currentRound),
+                fractionalTransferValue,
+                winner);
           }
         }
       }
--- a/src/main/java/com/rcv/Tabulator.java
+++ b/src/main/java/com/rcv/Tabulator.java
@@ -234,27 +234,35 @@ class Tabulator {
   // logic only considers continuing candidates, so it won't assign any votes to past winners -- but
   // in reality they continue to hold their winning margins for the rest of the rounds, so we need
   // to fill in those values here.
-  // TODO: instead of recomputing these winning tallies each round, we could just read the values
-  // from the previous round. Once someone wins and we redistribute their extra votes, their tally
-  // should be frozen for the rest of the rounds.
+  // We need to do the computation once for each winner to account for transfers. In subsequent
+  // rounds, we can just copy the number from the previous round, since it won't change.
   private void updatePastWinnerTallies() {
-    // pastWinners contains winners from rounds that preceded the current round
-    Set<String> pastWinners = new HashSet<>();
+    Map<String, BigDecimal> roundTally = roundTallies.get(currentRound);
+    Map<String, BigDecimal> previousRoundTally = roundTallies.get(currentRound - 1);
+    List<String> winnersToProcess = new LinkedList<>();
+    Set<String> winnersRequiringComputation = new HashSet<>();
     for (String winner : winnerToRound.keySet()) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/rcv/Tabulator.java
```diff
@@ -234,58 +234,70 @@ private void logSummaryInfo() {
...
-  // TODO: instead of recomputing these winning tallies each round, we could just read the values
-  // from the previous round. Once someone wins and we redistribute their extra votes, their tally
-  // should be frozen for the rest of the rounds.
+  // We need to do the computation once for each winner to account for transfers. In subsequent
+  // rounds, we can just copy the number from the previous round, since it won't change.
   private void updatePastWinnerTallies() {
-    // pastWinners contains winners from rounds that preceded the current round
-    Set<String> pastWinners = new HashSet<>();
+    Map<String, BigDecimal> roundTally = roundTallies.get(currentRound);
+    Map<String, BigDecimal> previousRoundTally = roundTallies.get(currentRound - 1);
```
HEdingfield: What prevents this from bugging out in the first round, since it presumably does a .get(0)?

src/main/java/com/rcv/Tabulator.java
```diff
@@ -234,58 +234,70 @@ private void logSummaryInfo() {
...
     for (String winner : winnerToRound.keySet()) {
       // skip someone who won in the current round, because we already have that tally filled in
-      if (winnerToRound.get(winner) < currentRound) {
-        pastWinners.add(winner);
+      int winningRound = winnerToRound.get(winner);
+      if (winningRound == currentRound) {
+        continue;
+      }
+      winnersToProcess.add(winner);
+      if (winningRound == currentRound - 1) {
```
HEdingfield: I'm a little confused here: why do all this checking for the previous round, etc., instead of just keeping a _global_ hash table of winners, and skipping calcs for them if they're in that list.  I realize this is similar to what you had before (except that was done locally within the function); I don't quite understand what this here buys us that a global table wouldn't, and it seems that the latter would be much simpler.

src/main/java/com/rcv/Tabulator.java
```diff
@@ -234,58 +234,70 @@ private void logSummaryInfo() {
   // logic only considers continuing candidates, so it won't assign any votes to past winners -- but
   // in reality they continue to hold their winning margins for the rest of the rounds, so we need
   // to fill in those values here.
```
moldover: Just to clarify: "in reality they continue to hold _and potentially increase_ their winning margins"?

src/main/java/com/rcv/Tabulator.java
tarheel: `.get(0)` is fine in itself -- it just returns null. And in that case, we're not going to end up trying to use the null later in the method, because `winnersToProcess` is guaranteed to be empty.

src/main/java/com/rcv/Tabulator.java
tarheel: Perhaps I'm misunderstanding your question... but `winnerToRound` is an instance variable (which I think is what you meant by global). I'm just rearranging that data here for convenience. If I'm missing your point, could you elaborate?

src/main/java/com/rcv/Tabulator.java
tarheel: No, just hold. They can't increase their totals.

Not sure if this is part or all of what you had in mind, but this logic shouldn't apply to the single-winner, keep-going-until-only-two-candidates-remain scenario. In that special case, yes, a winner can keep accumulating votes. I'll update the code to be explicit about only calling `updatePastWinnerTallies` when it's a multi-seat contest.

src/main/java/com/rcv/Tabulator.java
HEdingfield: Ok, got it. I still don't feel like I've got a good enough grasp on everything here to explicitly approve it (I should probably stop trying to review code late at night), but no other objections for me.

src/main/java/com/rcv/Tabulator.java
moldover: That's helpful, thanks.  I think I'm going to just edit the updatePastWinnerTallies main comment if you don't mind.  Having stepped through in the debugger with the portland_mayor_multi race it makes a lot more sense now.

tarheel: Sure. Do you mean that you want to update this PR yourself and then merge it? Or should I merge it?

moldover: Actually, I take it back.  On re-re-reading, isn't the end result of calling updatePastWinnerTallies that: previous winners have their round tallies added to the roundTallies object at the winning threshold?  If so, can't we just add a roundTally for past winners of the winningThreshold and skip all the cvr processing?  If I'm wrong, let's discuss in person.  I think you're fine to merge either way, seems to work fine.

moldover: Ok cool, that makes sense.

moldover: Do you think this could be more clearly calculated as the winner's surplus is transferred?

tarheel: Good question. I also considered that possibility, and yeah, I think it might be a bit clearer, but it's still a little awkward because we'd have to stash the value somewhere (since it would be happening one round before we were ready to put it in roundTallies) and then look it up later. If I hadn't already gotten the logic in this PR working, though, I would probably do what you're suggesting instead. I can create another low-pri task for this.

PR #183, opened 2018-11-11
Fix rounding of fractional shares for ballots that have multiple surplus redistributions

tarheel:
Previously we weren't rounding this operation, so if a ballot had a second surplus redistribution, we could end up with more than the permitted number of digits after the decimal.


tarheel:
> There are two other rogue multiply and divide call sites, both in Tabulator.java. Should they get this treatment too?

Good point. I was thinking that they didn't need it, but I reviewed them again and realized that you're right.

I also realized that I wasn't doing the floor operation for the threshold properly. `BigDecimal` has a remarkably confusing API.

PR #184, opened 2018-11-12
add ability to exclude a candidate from tabulation

### LINKED ISSUE #177: New feature requested:  ability to exclude a candidate from a tabulation
The concept is pretty simple:  sometimes a candidate withdraws from a race (or dies), and the candidate receives votes in the CVRs.  When you run the tabulation, you want to skip to the next ranking anytime a CVR would count for the withdrawn candidates.  (This is also the way Cambridge fills a vacancy when a council member resigns, gets elected to another office, or dies:  they re-do the tabulation excluding that candidate from the process.)

I would imagine that we could add a check box on the candidate tab of the GUI that would be labeled, "Exclude".  You could exclude any number of candidates and then run your tabulation.  The candidate could show up in the results but would have zero votes in every round.

I have no idea how tricky this will be to implement, but I imagine that you would create another state for candidates:  continuing, elected, eliminated and now "excluded".  If a CVR would count for an excluded candidate, you should skip to the next ranking, with no effect on skipped rankings or anything else.

Why don't you all think about this and see what milestone we should add it to. I put it in v 1.0, but that could change.

tarheel:
fixes #177

### DIFF
--- a/src/main/java/com/rcv/ContestConfig.java
+++ b/src/main/java/com/rcv/ContestConfig.java
@@ -51,6 +51,7 @@ class ContestConfig {
   private final ArrayList<String> candidatePermutation = new ArrayList<>();
   // mapping from candidate code to full name
   private Map<String, String> candidateCodeToNameMap;
+  private Set<String> excludedCandidates = new HashSet<>();
   // whether or not there are any validation errors
   private boolean isValid;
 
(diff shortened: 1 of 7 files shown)


moldover:
Looks good.  Is there supposed to be any indication of excluded candidates in the outputs?

HEdingfield:
Hoping to get responses to my comments (including the ones that aren't inline) before merging.

tarheel:
As discussed in an earlier diff -- I've stopped adding superfluous comments because that practice was based on an earlier requirement that might no longer be relevant. If/when we learn that we really do need them, we can go back and add them easily.

I've addressed all the other comments. For the case in which all of the candidates are excluded, I'm just going to have it fail validation.

Issue #186, opened 2018-11-17
Summary CSVs have redundant column


HEdingfield:
The new format introduced in #181 adds an extra unnecessary column at the far right of the results that just duplicates the info in the last round. E.g.: 

rounds | Round 1 | Round 2 |  
-- | -- | -- | --
DEM Golden, Jared F. | 63 | 66 | 66
REP Poliquin, Bruce | 48 | 48 | 48
Hoar, William R.S. | 5 | 0 | 0
Bond, Tiffany L. | 2 | 0 | 0
Undeclared | 0 | 0 | 0
Exhausted ballots | 0 | 4 |

Issue #185, opened 2018-11-17
Tabulating by precinct doesn't output JSON summaries


HEdingfield:
The JSON summaries aren't generated when using the "tabulate by precinct" setting, just the CSVs (related discussions in #181 and #105).

moldover:
Fixed.

PR #187, opened 2018-11-19
fix for #186 and #185

moldover:
@HEdingfield thanks for catching this


moldover:
Dear lord.  Thanks for catching that (too).  I need to do a better job testing next time.

Just pushed a fix which:
- Matches the header info (including "Precinct" for precinct summaries) in summary json file as the visualizer will want those fields eventually
- Does NOT add "action" rows for precinct summary json

tarheel:
Hmm. I don't see the fix that you pushed today.

moldover:
@tarheel try now

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -235,15 +245,10 @@ private void generateSummarySpreadsheet(
...
 
-      // displayRound indexes over all rounds plus final results round
-      for (int displayRound = 1; displayRound <= numRounds + 1; displayRound++) {
-        // flag for the last round results which are special-cased
-        boolean isFinalResults = displayRound == numRounds + 1;
-        // For the Final Results "round", we're mostly copying the data from the final round.
-        // round from which to display data
-        int dataUseRound = isFinalResults ? numRounds : displayRound;
+      // displayRound indexes over all rounds
+      for (int displayRound = 1; displayRound <= numRounds; displayRound++) {
```
tarheel: now we can just rename this back to `round`

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -428,32 +436,46 @@ private void generateSummaryJson(String outputPath,
...
     // config - global config into
     HashMap<String, Object> outputJson = new HashMap<>();
+    // config will contain contest configuration info
+    HashMap<String, Object> configData = new HashMap<>();
+    // add config header info
+    configData.put("Contest", config.getContestName());
+    configData.put("Jurisdiction", config.getContestJurisdiction());
+    configData.put("Office", config.getContestOffice());
+    configData.put("Date", config.getContestDate());
+    configData.put("Threshold", winningThreshold.toString());
```
tarheel: nit: `toString` not needed here?

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -428,32 +436,46 @@ private void generateSummaryJson(String outputPath,
...
+      if (precinct == null || precinct.isEmpty()) {
       // actions is a list of one or more action objects
-      ArrayList<Object> actions = new ArrayList<>();
-      // add any elected actions
-      addActionObjects("elected", roundToWinningCandidates.get(round), round, actions);
-      // add any elimination actions
-      addActionObjects("eliminated", roundToEliminatedCandidates.get(round), round, actions);
-      // add action objects
-      roundData.put("actions", actions);
+        ArrayList<Object> actions = new ArrayList<>();
```
tarheel: nit: LinkedList is a little cheaper, I think

moldover: you think?

src/main/java/com/rcv/ResultsWriter.java
moldover: yep

src/main/java/com/rcv/ResultsWriter.java
moldover: surprisingly, this works

src/main/java/com/rcv/ResultsWriter.java
tarheel: Well, an ArrayList is a dynamically-sized array, whereas a LinkedList is just a linked list. If you need to be able to access elements by index in constant time (vs. linear), ArrayList is definitely better. But I've always assumed that if you're just building a list that you're going to iterate over later, LinkedList is preferable because it's a simpler data structure. I could be wrong. In any case, it hardly matters at a tiny scale like this.

If you want to read way more on the subject: https://stackoverflow.com/questions/10656471/performance-differences-between-arraylist-and-linkedlist

src/main/java/com/rcv/ResultsWriter.java
HEdingfield: Counter-nit: `winningThreshold.toString()` is less confusing to me as a reader because it explicitly converts the `BigDecimal` to a `String`.

moldover: On second thought, doesn't this need to be a String to correctly preserve BigDecimal representation?  Think of this as serializing an object, not just for display purposes.

HEdingfield:
My complaint wasn't that actions were shown, it was that they were showing
what happened on the global level instead of the precinct level.

moldover:
My understanding is, actions only exist at the global level.  Can you give an example of what you would want to see, and why?

moldover:
Gotcha.  I'm thinking of "actions" as eliminations / elections (not transfers).  But terminology aside, you are correct that transfers are not currently calculated per precinct, though it could be done.  At any rate it is beyond the scope of this commit.  I made a new issue for it, #189.

src/main/java/com/rcv/ResultsWriter.java
tarheel: I assumed all of these were just getting converted to strings when it outputs them anyway, so it makes no difference. Anyway, whatever works.

HEdingfield:
Here's an example of how chaotic the JSON file can look to give context for my feedback (had to rename to .txt to upload):

[2018-11-27_22-34-55_summary.txt](https://github.com/BrightSpots/rcv/files/2623223/2018-11-27_22-34-55_summary.txt)

I anticipate you'll say that the JSON files aren't meant to be human-readable anyway, but I still think it's worth making those a bit more friendly if it's not too much extra work. I personally prefer to look at those over the CSV files because it's great to see the actions clearly outlined.

moldover:
Generally, json dict entries are unordered.  But, there is a nifty SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS setting which does alphabetic ordering, but it is "global".   It works well for 3 and happens to also achieve 1.   2. doesn't work, but overall it is easier to read.
[json_summary.txt](https://github.com/BrightSpots/rcv/files/2637923/json_summary.txt)

HEdingfield:
Good deal, LGTM!

PR #190, opened 2018-11-22
add nonIntegerWinningThreshold option

tarheel:
partially addresses #147


tarheel:
We really ought to start adding unit tests for things like this... but anyway, I tested both options manually and it seems to work OK.

src/main/java/com/rcv/Tabulator.java
```diff
@@ -326,8 +326,16 @@ private void setWinningThreshold(Map<String, BigDecimal> currentRoundCandidateTo
 
     // divisor for threshold is num winners + 1
     BigDecimal divisor = new BigDecimal(config.getNumberOfWinners() + 1);
-    // threshold = floor(votes / (num_winners + 1)) + 1
-    winningThreshold = currentRoundTotalVotes.divideToIntegralValue(divisor).add(BigDecimal.ONE);
+    if (config.isWinningThresholdIntegral()) {
+      // threshold = floor(votes / (num_winners + 1)) + 1
+      winningThreshold = currentRoundTotalVotes.divideToIntegralValue(divisor).add(BigDecimal.ONE);
+    } else {
+      // threshold = (votes / (num_winner + 1)) + 10^(-1 * decimalPlacesForVoteArithmetic)
```
HEdingfield: num_winner => num_winners

src/main/java/com/rcv/Tabulator.java
```diff
@@ -326,8 +326,16 @@ private void setWinningThreshold(Map<String, BigDecimal> currentRoundCandidateTo
...
+      BigDecimal augend =
```
HEdingfield: Not sure what `augend` means. Can we use a better variable name?

src/main/resources/config_file_documentation.txt
```diff
@@ -127,6 +127,14 @@ Config file must be valid JSON format. Examples can be found under the test fold
         value: [0..20]
         default: 4
 
+      "integralWinningThreshold" optional
+        the vote threshold used to determine winners should be an integer
+        if true, threshold = floor(V/(S+1)) + 1
+        if false, threshold = V(S+1) + 10^-d
```
HEdingfield: V(S+1) => V/(S+1)

src/main/java/com/rcv/ContestConfig.java
```diff
@@ -41,6 +41,7 @@
   static final boolean DEFAULT_EXHAUST_ON_DUPLICATE_CANDIDATES = false;
   static final boolean DEFAULT_CONTINUE_UNTIL_TWO_CANDIDATES_REMAIN = false;
   static final boolean DEFAULT_TREAT_BLANK_AS_UNDECLARED_WRITE_IN = false;
+  static final boolean DEFAULT_INTEGRAL_WINNING_THRESHOLD = true;
```
HEdingfield: I like that this defaults to `true`, but the problem is that when the option isn't defined in a config file (e.g. if you load one of the old ones), it actually defaults to false I think because of the way `RawContestConfig` is initialized. I mentioned this in the TODO at the top of this section and brought this up in a discussion we had before:
`// TODO: if any booleans are unspecified in config file, they default to false no matter what`

What I'd suggest is you actually invert this boolean and make it `NON_INTEGER_WINNING_THRESHOLD` and have it default to `false` here. That way, it's consistent with all our other booleans defaulting to false and also, if you load up an old config file, it'll properly default to the desired value.

src/main/resources/config_file_documentation.txt
```diff
@@ -127,6 +127,14 @@ Config file must be valid JSON format. Examples can be found under the test fold
         value: [0..20]
         default: 4
 
+      "integralWinningThreshold" optional
```
HEdingfield: I think it would be less confusing to replace all uses of the word "integral" with "integer". I understand after reading the code that you're using "integral" as the adjective form of "integer", but this word has multiple meanings in math and computer science, so just sticking with "integer" makes it a lot more clear (and yes, I also see that there's a precedent for "integral" in the `BigDecimal` API, but I think that was a poor choice on their part).

src/main/java/com/rcv/ContestConfig.java
tarheel: Yeah, I thought about this. I really don't like the idea that we can't have any boolean variables that default to true... but I guess there's no better alternative for now.

src/main/java/com/rcv/Tabulator.java
tarheel: It's the formal name for the second argument to an addition operation (and it's what BigDecimal calls the argument to the `add` method). I'm fine calling it something else, but I'm not sure what. `fudgeFactor`?

tarheel:
> Fully agree with you on needing unit tests. In the meantime, could you please include a functional test in the form of a constructed data set and two config files that flip this setting (i.e. where the results would change solely based on this setting)?

I don't think this would have incremental value now that I've tested it manually. I'll just add a TODO to add a real unit test once we have a framework.

tarheel:
image

PR #191, opened 2018-12-02
add final redistribution round for multi-seat

tarheel:
This is a little hairy, but I think it works OK. George and Caleb requested that we redistribute the surplus for all winners in a multi-seat contest, even if there are no more seats to fill. Since the way our code works is to declare a winner in round r and then show the consequences of redistributing that winner's surplus in round r+1, we need to add a single extra round of tabulation for multi-seat contests -- but unlike all other rounds, there are no actions (winner declarations or eliminations) in this round.

### DIFF
--- a/src/main/java/com/rcv/Tabulator.java
+++ b/src/main/java/com/rcv/Tabulator.java
@@ -361,6 +361,8 @@ class Tabulator {
     if (config.willContinueUntilTwoCandidatesRemain()) {
       return numEliminatedCandidates + numWinnersDeclared + 1 < config.getNumCandidates();
     } else {
+      // This line just finds the most recent round in which we've declared a winner (or -1 if we
+      // haven't declared any winners yet).
       int lastWinnerRound = winnerToRound.values().stream().mapToInt(v -> v).max().orElse(-1);
       // If there are more seats to fill, we should keep going, of course.
       // But also: if we've selected all the winners in a multi-seat contest, we should tabulate one
--- a/src/main/java/com/rcv/Tabulator.java
+++ b/src/main/java/com/rcv/Tabulator.java
@@ -156,17 +156,23 @@ class Tabulator {
         for (String winner : winners) {
           winnerToRound.put(winner, currentRound);
         }
-        for (String winner : winners) {
-          // are there still more winners to select in future rounds?
-          if (winnerToRound.size() < config.getNumberOfWinners()) {
+        // In multi-seat contests, we always redistribute the surplus (if any).
+        if (config.getNumberOfWinners() > 1) {
+          for (String winner : winners) {
             // number of votes the candidate got this round
             BigDecimal candidateVotes = currentRoundCandidateToTally.get(winner);
             // number that were surplus (beyond the required threshold)
             BigDecimal extraVotes = candidateVotes.subtract(winningThreshold);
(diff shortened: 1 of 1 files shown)


src/main/java/com/rcv/Tabulator.java
```diff
@@ -347,12 +356,17 @@ private boolean shouldContinueTabulating() {
...
     // how many winners have been selected
-    int numWinners = winnerToRound.size();
+    int numWinnersDeclared = winnerToRound.size();
     // apply config setting if specified
     if (config.willContinueUntilTwoCandidatesRemain()) {
-      return numEliminatedCandidates + numWinners + 1 < config.getNumCandidates();
+      return numEliminatedCandidates + numWinnersDeclared + 1 < config.getNumCandidates();
     } else {
-      return numWinners < config.getNumberOfWinners();
+      int lastWinnerRound = winnerToRound.values().stream().mapToInt(v -> v).max().orElse(-1);
```
moldover: please explain yourself

src/main/java/com/rcv/Tabulator.java
```diff
@@ -395,16 +409,20 @@ private CandidateStatus getCandidateStatus(String candidate) {
...
-    } else { // see if anyone has met or exceeded the threshold
-      // tally indexes over all tallies to find any winners
-      for (BigDecimal tally : currentRoundTallyToCandidates.keySet()) {
-        if (tally.compareTo(winningThreshold) >= 0) {
-          // we have winner(s)
-          List<String> winningCandidates = currentRoundTallyToCandidates.get(tally);
-          selectedWinners.addAll(winningCandidates);
+    // We should only look for more winners if we haven't already filled all the seats.
+    if (winnerToRound.size() < config.getNumberOfWinners()) {
+      // If the number of continuing candidates equals the number of seats to fill, everyone wins.
```
moldover: but do they?

src/main/java/com/rcv/Tabulator.java
```diff
@@ -395,16 +409,20 @@ private CandidateStatus getCandidateStatus(String candidate) {
...
+      if (currentRoundCandidateToTally.size()
+          == config.getNumberOfWinners() - winnerToRound.size()) {
+        selectedWinners.addAll(currentRoundCandidateToTally.keySet());
+      } else { // see if anyone has met or exceeded the threshold
```
moldover: I'm not against this, but I'm not sure this formatting is keeping with the style

moldover:
Did you check the json and csv summary outputs for correctness?

src/main/java/com/rcv/Tabulator.java
tarheel: This is just a way of getting the max value, or -1 if there's no max. If I don't add that `orElse`, the result is an `OptionalInt`, which is surprisingly awkward to work with. I'll add a comment.

src/main/java/com/rcv/Tabulator.java
tarheel: image

src/main/java/com/rcv/Tabulator.java
tarheel: This is actually what the auto-formatter chose. :( Not my favorite.

PR #203, opened 2018-12-28
make max rankings and max skipped ranks fields optional

### LINKED ISSUE #199: Make Max Rankings and Max Skipped Rankings Optional
Can these just default to unlimited, or equivalently, the number candidates?

tarheel:
Fixes #199.

I also noticed that the GUI was using the ContestConfig object to populate the form fields. I changed this so that it uses the RawContestConfig object instead in most cases, since the ContestConfig accessors might have additional logic that we don't want here.

One other thing I encountered: the Number of Winners, Decimal Places for Vote Arithmetic, and Minimum Vote Threshold fields are kind of weird in the GUI. They're not labeled as required, but they have default prepopulated values, and if you try to remove one of these values and leave the field blank, it gets silently added back, plus we log a warning (e.g. "WARNING: Integer required! Illegal value '' was replaced by '0'"). I'm not exactly sure what to recommend here, but the current behavior feels slightly off. I think the way the GUI tries to restore the old value when you put in something invalid is part of what makes it feel weird. I'd rather just let the user screw it up and then throw an error in validation.


src/main/resources/config_file_documentation.txt
```diff
@@ -110,15 +110,17 @@ Config file must be valid JSON format. Examples can be found under the test fold
         how the program should handle an overvote when it encounters one
         value: "alwaysSkipToNextRank" | "exhaustImmediately" | "exhaustIfMultipleContinuing"
 
-      "maxRankingsAllowed" required
+      "maxRankingsAllowed" optional
         maximum number of candidates that a ballot is allowed to rank
         example: 15
         values: [1..100]
+        default: the number of declared candidates
```
HEdingfield: Have we actually hard-capped the number of declared candidates at 100 anywhere? If not, this is kind of confusing. Maybe change values to `[1...total number of candidates]`?

src/main/resources/config_file_documentation.txt
```diff
@@ -110,15 +110,17 @@ Config file must be valid JSON format. Examples can be found under the test fold
...
 
-      "maxSkippedRanksAllowed" required
+      "maxSkippedRanksAllowed" optional
         maximum number of skipped ranks (undervotes) on a ballot before the ballot should be considered exhausted
         example: 1
         value: [0..100]
+        default: unlimited
```
HEdingfield: This is confusing because the default value is outside the valid value range. Also, same comment applies about capping number of candidates at 100?

test_data/config_test_no_optional_fields.json
```diff
@@ -0,0 +1,45 @@
+{
```
HEdingfield: Nice, thanks for this test! I wonder though: do we actually have any test CVRs that push the edge case for max skipped ranks allowed being unlimited?

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -585,45 +585,45 @@ private boolean checkForSaveAndTabulate() {
...
   private void loadConfig(ContestConfig config) {
     clearConfig();
-
-    textFieldContestName.setText(config.getContestName());
-    textFieldOutputDirectory.setText(config.getOutputDirectory());
-    if (config.getContestDate() != null && !config.getContestDate().isEmpty()) {
-      datePickerContestDate.setValue(LocalDate.parse(config.getContestDate(), DATE_TIME_FORMATTER));
+    RawContestConfig rawConfig = config.getRawConfig();
+
+    OutputSettings outputSettings = rawConfig.outputSettings;
```
HEdingfield: Something is telling me that it's better to keep the original design here (using `ContestConfig`). Maybe this stems from me thinking that explicit is always better than implicit, even for unspecified "default" settings.

For example, say we have a default value of 2 for a setting in one version of the software. By using `ContestConfig`, it would actually populate this default value in the GUI, then force the user to re-save before running the tabulator and there's a clear record of what all the settings were. Then in a later update to the software we change the default value to 3. If users don't have that explicitly-defined (default) value in the config, I can imagine them being very confused as to why they can't replicate past results unless they've been following the patch notes closely. Does that make sense?

Of course, this leaves us with the problem of how to indicate "unlimited" for `maxSkippedRank`. I think maybe -1 could be a good value for that, and just note the significance in the documentation.

HEdingfield:
Just realized that your comment on default values getting silently added back in is a necessary extension of my "explicit is better than implicit" argument. I agree that it's an unusual UX when you're saving something to have an invalid value overwritten by a default one. I can't really think of a good alternative at the moment though... at least it tells users what's going on, and maybe that's good enough?

Those fields aren't labeled as required precisely because they have solid default values that they can revert to. The actual field is absolutely required for the purpose of running the calculations, but it's not required to be entered by the user. Do we need to distinguish between this sort of field and one which genuinely isn't required at all (i.e. where null is a valid value)?

src/main/java/com/rcv/GuiConfigController.java
tarheel: I don't really like the idea of using -1 to signify unlimited. It's more intuitive to indicate no limit by just not specifying a limit. But I think we can handle this fine if we update the code so that we don't try to add the default value back when the user has cleared a non-required field.

tarheel: See what you think now.

HEdingfield:
After thinking about this more, I'd suggest the following:
- We abandon the use of the term "default" entirely, and rename our `*DEFAULT*` constants to `*SUGGESTED*` or `*RECOMMENDED*`.
- For all optional fields, we list "if blank:" or "if undefined:" and describe the behavior.
- We don't document what the pre-populated values are to keep us from needing to remember to update the documentation if we change them. If you really think there's a value in documenting these, list as "suggested:" or "recommended:" (this can be for both optional and required fields; any that have pre-populated values in the GUI).

Let me know what you think!

tarheel:
Yes, this makes a lot of sense. You're right that we were using the concept of "default" to mean multiple things, so disambiguating those cases is a good solution. I'll update the PR accordingly.

tarheel:
Rebased. @HEdingfield does this look OK now?

HEdingfield:
Sorry, GitHub doesn't reliably send me messages when you push commits that address comments I've raised. I'm wondering if we should follow a convention where, say, you'd start your own review after pushing a commit and then use it to respond to each of the comments I made with "Done" or "Won't address now because..." or whatever. That way I can just scan for any comments that I made that still don't have responses and know if there are any outstanding issues.

Anyway, not necessary now, just something to think about.  I'll review your latest changes tomorrow evening.

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: After reviewing this, it seems like your redesign works well to address this problem with one caveat: we have to be *very* careful not to change "if not supplied" behavior in future versions. Filed #211 to discuss potential safeguards against this.

test_data/config_test_no_optional_fields.json
HEdingfield: Added #212 for this.

HEdingfield:
This looks great. I pushed a commit that does a little more clean up, hopefully it looks good to you. Marking as approved, with a couple last questions:

1. Do we want to offer suggested tiebreakMode and/or overvoteRule?
2. Why do we have suggested maxSkippedRankingsAllowed of 1? Is this better than unlimited? It's the only case where we have a suggested value for a non-required field.

tarheel:
> Do we want to offer suggested tiebreakMode and/or overvoteRule?

Um, not sure. I don't have a strong opinion here.

> Why do we have suggested maxSkippedRankingsAllowed of 1? Is this better than unlimited? It's the only case where we have a suggested value for a non-required field.

Again, no strong opinion here...

Your changes look fine, btw. Going to finally merge this!

PR #204, opened 2019-01-06
Regression test cleanup pt 1

moldover:
The main change here improves how we resolve relative paths in config files.  Previously we used the current working directory (folder containing the application) as the base.  This is inflexible in general, and caused a bug where relative paths in config files would get overwritten with absolute paths when saving / loading via the GUI.  With this update we use the _location of the current config file being processed_ as the base for resolving relative paths.  This enables portable user, debug, and regression test configurations, more intuitive path behavior in general, and fixes the aforementioned bug.

Other changes: 

More concise test folder / file naming conventions (see "2015_portland_mayor" folder):
- each test (inputs and outputs) will be contained in a folder of same name.  This is more scalable organizationally and makes it easier to work with individual tests or tabulation runs
- change "cvr" "config" prefixes to suffixes.  This is more standard, more readable and improves sorting

Check for cvr file existence when validating cvrs

Added 2013_minneapolis_mayor_scale regression test

runTabulationTest takes test "stem" as input (simpler test creation)


src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -274,7 +277,7 @@ public void buttonClearDatePickerContestDateClicked() {
 
   public void buttonCvrFilePathClicked() {
     FileChooser fc = new FileChooser();
-    fc.setInitialDirectory(new File(System.getProperty("user.dir")));
+    fc.setInitialDirectory(new File(FileUtils.getUserFolder()));
```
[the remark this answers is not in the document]
moldover: Damn.  Actually you are correct, but not because of this code here.  The text cells in the GUI are "observable" and whenever the user clicks 'add' to add a new cvr record, the GUI code calls the getter getFilePath() which as written now always returns the user path.  I'm going to fix this by renaming the functions.  

To code you reference here, this will set the initial folder location when you are browsing to select a cvr file to the user folder which is I believe desired behavior.  The file chooser will always return an absolute path, which is also I believe desired behavior.  If a user wants to save a relative path they will need to enter it manually into the text edit box.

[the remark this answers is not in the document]
tarheel: Oh, yeah, sorry, I wasn't trying to reference this particular line. GitHub just doesn't let me comment inline on a line that you didn't modify.

I'm a little confused about the intention of these directory changes, then. I thought part of the point was to facilitate sharing across environments by making it easier to set relative paths in the config. So you're saying that the file picker (for the output directory and the CVR source files) will still return an absolute path, but if the config already has a relative path and I don't modify it, it won't get changed to absolute when I save the config?

moldover:
@tarheel updated code to address all your comments.

src/main/java/com/rcv/GuiConfigController.java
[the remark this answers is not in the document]
moldover: That is correct:  File picker still returns an absolute path.  Existing paths will not be modified.

This point is still "This _enables_ portable user, debug, and regression test configurations, more intuitive path behavior in general"

I will be using this feature as I clean up unit test data, and George, Caleb and our "power users" can use it to share data as well.

src/main/java/com/rcv/ContestConfig.java
```diff
@@ -107,59 +108,73 @@ private void validateCvrFileSources() {
...
           isValid = false;
           Logger.log(
-              Level.SEVERE, "Duplicate CVR filePaths are not allowed: %s", source.getFilePath());
+              Level.SEVERE, "Duplicate CVR filePaths are not allowed: %s",
+              source.getFullFilePath());
         } else {
-          cvrFilePathSet.add(source.getFilePath());
+          cvrFilePathSet.add(source.getFullFilePath());
         }
-
```
tarheel: Nit: for readability, it's nice to have a blank line between checks. Right now you randomly have them in some places and not in others in this method.

src/main/java/com/rcv/FileUtils.java
```diff
@@ -20,9 +20,45 @@
...
+  // function: setUserDirectory
+  // param: userDirectory default folder for finding and creating files and folders
+  static void setUserDirectory(String userDirectory) {
+    FileUtils.userDirectory = userDirectory;
+  }
+
+  // function: getUserDirectory
+  // returns: returns root for loading and saving user files
+  static String getUserDirectory() {
+    // return userDirectory if it exists fallback to current working directory
```
tarheel: delete second half of this line

src/main/java/com/rcv/RawContestConfig.java
```diff
@@ -78,12 +78,16 @@
     private Integer precinctColumnIndex;
     // provider for this source e.g. "ES&S"
     private String provider;
-
+    // get underlying raw path string
```
tarheel: for consistency this should probably be called getFilePathRaw, right?

tests/com/rcv/TabulatorTests.java
```diff
@@ -90,19 +91,25 @@ static boolean fileCompare(String path1, String path2) {
...
-  // param: expectedFile name of expected summary results json file
-  static void runTabulationTest(String configFile, String expectedFile) {
+  // param: stem base name of folder containing config file cvr files and expected result files
+  static void runTabulationTest(String stem) {
     // full path to config file
     String configPath =
-        Paths.get(System.getProperty("user.dir"), TEST_ASSET_FOLDER, configFile)
+        Paths.get(System.getProperty("user.dir"),
+            TEST_ASSET_FOLDER, stem,
+            stem + "_config.json")
```
tarheel: I'm not sure, but this doesn't look like what the automatic reformatting would yield.

tests/com/rcv/TabulatorTests.java
```diff
@@ -128,15 +135,23 @@ public static void setup() {
...
   void testPortlandMayor() {
-    runTabulationTest("config_2015_portland_mayor.json", "2015_portland_mayor_expected.json");
+    runTabulationTest("2015_portland_mayor");
   }
 
+  // function: test2013MinneapolisMayorScale
+  // purpose: test large scale (1,000,000+) cvr contest
+  @Test
+  @DisplayName("2013 Minneapolis Mayor Scale")
+  void test2013MinneapolisMayorScale() {
```
tarheel: How long does this take? It's good to have, but might get a little annoying if it runs every time we want to run the test suite.

tests/com/rcv/TabulatorTests.java
moldover: you know... I have the reformat code option checked in intelliJ but it doesn't seem to pick this up :(

tarheel: I find it works more consistently if you run it manually. The crazy part is that running it manually applies slightly different rules vs. what the commit hook does.

PR #206, opened 2019-01-07
fix bugs uncovered by Armin's tests

### LINKED ISSUE #205: bugs in JSON summary
Armin's tests with 2013_minneapolis_park and 2017_minneapolis_mayor revealed a few bugs in our code:
- 2013 park: the tally says UWI, but then we eliminated Undeclared. These strings should match.
- 2017 mayor: UWI's tally is 137, but when Undeclared is eliminated, only 136 votes are accounted for in the transfers.
- 2017 mayor: Theron Preston Washington should have an empty transfers map when they get eliminated.

tarheel:
Fixes #205.

This PR resolves three issues:
- When we exhausted a ballot due to overvote when that overvote was in the last ranking slot, we weren't properly recording the transfer.
- When we had an elimination or win with no transfers, we weren't adding a transfers map to the JSON. Not technically a bug, but arguably this makes the code easier on the visualization side.
- We weren't translating back to the full candidate names when we printed tallies in the JSON.

### DIFF
--- a/src/main/java/com/rcv/ResultsWriter.java
+++ b/src/main/java/com/rcv/ResultsWriter.java
@@ -44,7 +44,6 @@ import java.util.logging.Level;
 import org.apache.commons.csv.CSVFormat;
 import org.apache.commons.csv.CSVPrinter;
 
-
 class ResultsWriter {
 
   // number of rounds needed to elect winner(s)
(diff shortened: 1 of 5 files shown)


moldover:
One thing to keep in mind now that we are implementing regression tests: any changes to summary json  will "break" regression tests, requiring new test outputs to be verified and updated.  Making sure tests complete will become part of the PR process.

tarheel:
Good point. Ideally we would start using an automated solution like CircleCI for this... but for the moment it's not a big deal to expect the author to run the one test class manually.

Issue #198, opened 2018-12-19
Refactor Main.java into TabulationSession class


moldover:
This has 2 goals: 

1) We need to add state to support testing.  For example, for issue #195 we need some way of storing which precinct ids were discovered during cvr parsing, and then mapping them to output summary files to ensure that code is stable for regression testing.

2) There is a bunch of processing logic which doesn't belong in a Main class (executeTabulation, loadContestConfig, etc.) and also doesn't belong in Tabulator class either.    

I propose a TabulationSession class which encapsulates these tasks and removes them from Main.

tarheel:
Sounds fine to me. Since you have a clear sense of what's needed here, do you want to take it?

Issue #195, opened 2018-12-17
Support precinct summary file comparison for regression testing


moldover:
Regression tests need to compare actual tabulation result file to expected result files.  Our tabulator "discovers" precinct IDs during the course of tabulation.  We would need some way of exposing these IDs to the test harness, and mapping them to a list of known files.

freedomcounts:
Do we need this for certification?  It would seem likely to me.

tarheel:
Note that this also needs to work for the sequential case. I'm fixing that output logic in #291.

HEdingfield:
I think this is the same issue I ran into today: precinct tests like `precinct_example` and `missing_precinct_example` currently only compare the overall .json summary to a single expected file. They should compare each precinct summary as well (including the missing one in the latter case).

tarheel:
Are we still comfortable that this is non-essential? I'm working on #189 and would need this issue to be addressed in order to automate testing of my work.

tarheel:
(Note that we also labeled #189 as non-essential, so we're being consistent.)

PR #209, opened 2019-01-16
Tabulator session

moldover:
This PR adds a TabulatorSession object (issue #198) which manages the broader tabulation process including reading config, validating config, parsing cvr files, tabulating, and writing results.  This cleans up Main, and facilitates further testing and validation work.  For example, precinct IDs are now collected and stored in the TabulatorSession object.  This will allow #195 to be completed. 

I also added a bunch of regression tests based on existing test data.  With these tests we now have a real tool for protecting ourselves from making inadvertent changes. 

@tarheel @HEdingfield when this PR is merged, going forward, it will be the responsibility of anyone merging to master to ensure that all the tests complete successfully.  The current test suite runs in about 80 seconds on my computer.


src/main/java/com/rcv/TabulatorSession.java
```diff
@@ -0,0 +1,260 @@
...
+      } catch (IOException exception) {
+        Logger.log(
+            Level.SEVERE, "Failed to configure tabulation logger!\n%s", exception.toString());
+      }
+
+      if (isTabulationLogSetUp) {
+        Logger.log(Level.INFO, "Logging tabulation to: %s", tabulationLogPath);
+        // Read cast vote records from CVR files
+        // parse the cast vote records
+        Pair<List<CastVoteRecord>, Set<String>> cvrInfo = parseCastVoteRecords(config);
```
tarheel: Not a big deal, but it's unfortunate that we then have to call `getValue` and `getKey` on this when it's not actually a key-value pair...

test_data/2015_portland_mayor_multi/2015_portland_mayor_multi_config.json
```diff
@@ -0,0 +1,83 @@
+{
```
tarheel: I removed this one. It's not a real thing (mayor is not multi-seat) and it doesn't test anything in particular.

tests/com/rcv/TabulatorTests.java
```diff
@@ -1,6 +1,6 @@
 /*
  * Ranked Choice Voting Universal Tabulator
- * Copyright (c) 2018 Jonathan Moldover, Louis Eisenberg, and Hylton Edingfield
+ * Copyright (c) 2018 Jonath  an Moldover, Louis Eisenberg, and Hylton Edingfield
```
tarheel: undo

test_data/2015_portland_mayor_multi/2015_portland_mayor_multi_config.json
moldover: Oh that's true.  I added this because I wanted a short and simple multi seat test for debugging something, but yeah it doesn't need to be a test.

src/main/java/com/rcv/TabulatorSession.java
moldover: I changed it to pass the IDs as a parameter so it's a bit simpler.

src/main/java/com/rcv/TabulatorSession.java
```diff
@@ -0,0 +1,260 @@
...
+    // set config file parent folder as default user folder
+    FileUtils.setUserDirectory(new File(configPath).getParent());
+    // config: the new contestConfig object
+    ContestConfig config = null;
+
+    // rawConfig holds the basic contest config data parsed from json
+    // this will be null if there is a problem loading it
+    RawContestConfig rawConfig = JsonParser.createRawContestConfigFromFile(configPath);
+
+    // if raw config failed alert user
```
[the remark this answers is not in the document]
moldover: I agree this code should be cleaned up (probably moved into ContestConfig), but it's outside the scope of this PR.  I created a new issue for it #210.

moldover:
@tarheel I think I addressed all your comments

src/main/java/com/rcv/StreamingCVRReader.java
```diff
@@ -185,6 +189,10 @@ private void endCVR() {
             currentRankings);
     // add it to overall list
     cvrList.add(newRecord);
+    // add precinct ID if one was found
+    if(currentPrecinct != null) {
```
tarheel: reminder to run Code > Reformat Code

src/main/java/com/rcv/TabulatorSession.java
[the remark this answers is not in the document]
tarheel: But yeah, I see now that you just moved this entire method from elsewhere, so it's reasonable not to modify it in this PR.

tests/com/rcv/TabulatorTests.java
```diff
@@ -89,33 +90,36 @@ static boolean fileCompare(String path1, String path2) {
     return result;
   }
 
+  // function: getConfigPath
+  // purpose: given stem returns path to config file in test asset folder
+  // returns: path to config file
+  static String getConfigPath(String stem) {
```
HEdingfield: You could generalize the `getConfigPath` function to generate the `expectedPath` too (e.g. with a simple string parameter that inserts whichever word after the underscore and before .json). You should rename the function if you do this though.

tests/com/rcv/TabulatorTests.java
```diff
@@ -89,33 +90,36 @@ static boolean fileCompare(String path1, String path2) {
...
+
+    // create a session object and run the tabulation
+    TabulatorSession session = new TabulatorSession(TabulatorTests.getConfigPath(stem));
+    session.tabulate();
     // actualSummaryOutputPath is the summary json we just tabulated
-    String actualSummaryOutputPath = Main.getSummaryOutputPath();
+    String actualSummaryOutputPath = session.summaryOutputPath;
     // compare actual to expected
-    assertTrue(fileCompare(actualSummaryOutputPath, expectedPath));
+    assertTrue(fileCompare(expectedPath, actualSummaryOutputPath));
```
HEdingfield: Since we're actually creating files here, I'd strongly recommend adding logic here that cleans up after the test is complete, e.g. a `teardown` function that removes all created files and directories after each test is run.

It's also really hard to parse out the results of each individual test given all the log / console window spam; I'd think the testing framework should allow us to only show the test results in a single output / console window? I guess it's not really a problem as long as we're running things inside of IntelliJ, but maybe something to add to the wishlist.

I'm not sure about the whole `junit.jupiter` sub-library; why are we using that instead of the basic `junit` suite?

tests/com/rcv/TabulatorTests.java
```diff
@@ -75,9 +75,10 @@ static boolean fileCompare(String path1, String path2) {
         // both files have content so compare it
         if (!line1.equals(line2)) {
           // report inequality and keep processing in case there are more inequalities
-          Logger.log(Level.SEVERE, "files are not equal:\n%s\n%s", line1, line2);
+          Logger.log(Level.SEVERE, "files are not equal (%d)\n%s\n%s", currentLine, line1, line2);
           result = false;
```
HEdingfield: Add a break here to avoid log spam if lots of lines aren't equal?

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -243,7 +245,7 @@ public void buttonTabulateClicked() {
     if (checkForSaveAndTabulate()) {
       if (GuiContext.getInstance().getConfig() != null) {
         buttonBar.setDisable(true);
-        TabulatorService service = new TabulatorService();
+        TabulatorService service = new TabulatorService(selectedFile.getAbsolutePath());
```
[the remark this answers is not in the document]
moldover: I believe this is correct encapsulation.  TabulatorService needs to know about TabulationSession and how to construct one.  It does not need to know about GUIContext.  Why add that dependency?

src/main/java/com/rcv/StreamingCVRReader.java
moldover: :+1:

tests/com/rcv/TabulatorTests.java
```diff
@@ -75,9 +75,10 @@ static boolean fileCompare(String path1, String path2) {
         // both files have content so compare it
         if (!line1.equals(line2)) {
           // report inequality and keep processing in case there are more inequalities
-          Logger.log(Level.SEVERE, "files are not equal:\n%s\n%s", line1, line2);
+          Logger.log(Level.SEVERE, "files are not equal (%d)\n%s\n%s", currentLine, line1, line2);
```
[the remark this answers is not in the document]
moldover: :+1:

tests/com/rcv/TabulatorTests.java
moldover: Added a counter which caps output at 10 lines

tests/com/rcv/TabulatorTests.java
moldover: dig it

tests/com/rcv/TabulatorTests.java
moldover: Yeah, I was thinking about something like that too... the scale test does produce a lot of data.  I think automated cleanup would be fine and just skip cleanup when the test fails so we can look at it.  I will add that.  

As for output, yes there is a visualizer for this.  Did you follow the directions here?  https://github.com/BrightSpots/rcv/pull/194

It is using junit5, and you should see something like this when running tests from IntelliJ:

regressiontests

src/main/java/com/rcv/GuiConfigController.java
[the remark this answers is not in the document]
HEdingfield: It just seemed weird to me that we create a `TabulatorSession`object in line 178, but then we create a new one when we actually execute the tabulation. Maybe pass in the existing `TabulatorSession` instead of the path when initializing `TabulatorService`? Or maybe it's not a big deal? Just seemed like code smell to me.

tests/com/rcv/TabulatorTests.java
HEdingfield: Clean up sounds great!

Yep, sorry I wasn't more clear: I do have that working in IntelliJ, but I was thinking about someone running the tests outside of IntelliJ, like from the CLI. Just my reluctance to rely on third party software to develop effectively. Anyway, this isn't necessary to worry about now since the three of us are all using IntelliJ, but I think it's worth filing an issue to make it more friendly to other devs who might not be in the future.

src/main/java/com/rcv/GuiConfigController.java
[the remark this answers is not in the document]
moldover: Gotcha.  Yes, your instinct is correct.  The code needs another piece of cleanup to refactor how ContestConfig objects are created and validated, so they can be used aside from TabulatorSession.  When that's done, the TabulatorSession reference in line 178 will be replaced with something like:
ContestConfig config = ContestConfig.loadContestConfig(fileToLoad.getAbsolutePath);

tests/com/rcv/TabulatorTests.java
moldover: Gotcha.  It really makes sense to use a thrid party tool here.  I researched this when I began planning the test code and there are various mature options.  I picked Junit5 because it is modern, widely used, well supported, and offers a rich feature set, although I expect we will only use the basics.  It can be integrated with other IDEs, and run standalone as well.  IMO it is pretty developer friendly.  https://stackoverflow.com/questions/52373469/how-to-launch-junit-5-platform-from-the-command-line-without-maven-gradle/52373592#52373592

Issue #210, opened 2019-01-17
Clean up ContestConfig construction and validation


moldover:
Config validation and construction code in TabulatorSession.loadContestConfig should be moved and consolidated with existing code in ContestConfig.java

PR #213, opened 2019-01-31
Cleanup ContestConfig creation and validation logic

moldover:
This PR addresses issue #210  by consolidating ContestConfig creation and validation logic into ContestConfig class.  This improves encapsulation and comprehensibility.

I also added VVSG reference files and renamed the existing ones as additional unrelated cleanup.


src/main/java/com/rcv/ContestConfig.java
```diff
@@ -59,6 +58,65 @@
...
+    // set config file parent folder as default user folder
+    FileUtils.setUserDirectory(new File(configPath).getParent());
+
+    // rawConfig holds the basic contest config data parsed from json
+    // this will be null if there is a problem loading it
+    RawContestConfig rawConfig = JsonParser.parseObjectFromFile(configPath, RawContestConfig.class);
+    if (rawConfig == null) {
+      Logger.log(Level.SEVERE, "Failed to load contest config: %s", configPath);
+    } else {
+      Logger.log(Level.SEVERE, "Loaded: %s", configPath);
```
tarheel: this shouldn't be SEVERE, should it?

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -219,7 +217,8 @@ private File getSaveFile() {
   private void saveFile(File fileToSave) {
     // set save file parent folder as the new default user folder
     FileUtils.setUserDirectory(fileToSave.getParent());
-    JsonParser.createFileFromRawContestConfig(fileToSave, createRawContestConfig());
+    // create a rawConfig object from GUI content and serialize it as json
+    JsonParser.parseObjectToFile(fileToSave, createRawContestConfig());
```
tarheel: This is just a semantic question, but is "parse" the right verb here? It feels like the opposite of what this is doing. "Serialize" would make more sense to me.

src/main/java/com/rcv/JsonParser.java
```diff
@@ -29,37 +29,43 @@
...
+  // function: parseObjectFromFile
+  // purpose: parse input json file into an object of the specified type
   // param: jsonFilePath path to json file to be parsed into java
+  // param: valueType class of the object to be created from parsed json
   // file access: read
-  // returns: instance of the RawContestConfig parsed from json, or null if there was a problem
-  static RawContestConfig createRawContestConfigFromFile(String jsonFilePath) {
-    RawContestConfig rawConfig;
+  // returns: instance of the object parsed from json or null if there was a problem
+  static <T> T parseObjectFromFile(String jsonFilePath, Class<T> valueType) {
```
tarheel: I'm curious why you bothered to make these methods generic. Will that have a practical purpose in the future?

src/main/java/com/rcv/ContestConfig.java
moldover: yep

src/main/java/com/rcv/GuiConfigController.java
moldover: yeah serialize is better

src/main/java/com/rcv/JsonParser.java
moldover: The main motivator here is consolidating the ContestConfig construction code.  Now, originally these functions were generic when I wrote them.  I'm not sure why it was changed to be specific to RawContestConfig, as that conflates json serializing with RawContestConfig construction logic and needlessly removes the generic capability.  Additionally this does have a practical use here.  When I was writing regression testing code, I used JsonParser with a new ContestSummary object to do comparisons.  I did not end up going that route (line by line file comparison proved more useful) but this code helped me explore that option.  I don't have another specific use case in mind atm, but there's no reason not to have this api.

src/main/java/com/rcv/GuiConfigController.java
moldover: agree, maybe I should change this to serialize / deserialize?

src/main/java/com/rcv/ContestConfig.java
```diff
@@ -36,7 +36,6 @@
 
 class ContestConfig {
 
-  // TODO: if any booleans are unspecified in config file, they default to false no matter what
```
HEdingfield: I think we do need to keep some kind of warning / reminder to us or any future coders about this... a TODO probably wasn't the best way to do it, but it's important.

src/main/java/com/rcv/ContestConfig.java
```diff
@@ -59,6 +58,65 @@
...
+    // config will hold the new ContestConfig if construction succeeds
+    ContestConfig config = null;
+    // set config file parent folder as default user folder
+    FileUtils.setUserDirectory(new File(configPath).getParent());
+
+    // rawConfig holds the basic contest config data parsed from json
+    // this will be null if there is a problem loading it
+    RawContestConfig rawConfig = JsonParser.readFromFile(configPath, RawContestConfig.class);
+    if (rawConfig == null) {
+      Logger.log(Level.SEVERE, "Failed to load contest config: %s", configPath);
```
HEdingfield:
```suggestion
      Logger.log(Level.SEVERE, "Failed to load contest config:\n%s", configPath);
```

Easier to read notifications when they're on separate lines from garbagey file paths.

src/main/java/com/rcv/ContestConfig.java
```diff
@@ -59,6 +58,65 @@
...
+    } else {
+      Logger.log(Level.INFO, "Loaded: %s", configPath);
```
HEdingfield:
```suggestion
      Logger.log(Level.INFO, "Successfully loaded contest config:\n%s", configPath);
```

src/main/java/com/rcv/ContestConfig.java
```diff
@@ -59,6 +58,65 @@
...
+      }
+    }
+    return config;
+  }
+
+  // function: validateRawConfigFields
+  // purpose: perform some validation checks
+  // param: RawContestConfig to perform checks against
+  // returns: true if checks pass otherwise false
+  private static boolean validateRawConfigFields(RawContestConfig rawConfig) {
```
HEdingfield: It makes more sense to me to have this live inside `RawContestConfig`, since we've already got a `validate` function living here in `ContestConfig`.

src/main/java/com/rcv/JsonParser.java
```diff
@@ -29,40 +28,47 @@
...
+  // function: readFromFile
+  // purpose: parse input json file into an object of the specified type
   // param: jsonFilePath path to json file to be parsed into java
-  // file access: read
-  // returns: instance of the RawContestConfig parsed from json, or null if there was a problem
-  static RawContestConfig createRawContestConfigFromFile(String jsonFilePath) {
-    RawContestConfig rawConfig;
+  // param: valueType class of the object to be created from parsed json
+  // returns: instance of the object parsed from json or null if there was a problem
+  static <T> T readFromFile(String jsonFilePath, Class<T> valueType) {
```
HEdingfield: Nice, I've had a mind to generalize these for a while now!

src/main/java/com/rcv/JsonParser.java
```diff
@@ -29,40 +28,47 @@
...
+  // param: jsonFile File object to write to
+  // param: objectToSerialize object to be written
+  static void writeToFile(File jsonFile, Object objectToSerialize) {
     try {
-      new ObjectMapper().writer().withDefaultPrettyPrinter().writeValue(jsonFile, config);
-      Logger.log(
-          Level.INFO, "Saved config via the GUI to: %s", jsonFile.getAbsolutePath());
+      new ObjectMapper().writer().withDefaultPrettyPrinter().writeValue(jsonFile,
+          objectToSerialize);
+      Logger.log(Level.INFO, "Saved object to: %s", jsonFile.getAbsolutePath());
```
HEdingfield:
```suggestion
      Logger.log(Level.INFO, "Saved to:\n%s", jsonFile.getAbsolutePath());
```
I think seeing "object" might confuse users.

PR #220, opened 2019-02-07
George's multi-winner test 3

tarheel:
This was a fun one! The issue that @gngilbert found arose from a subtle mistake I had made in the order of operations in the `recordCurrentRecipientAsWinner` method in `CastVoteRecord` (which I think only became relevant once I switched the rounding mode to DOWN). I hope that the inline comment that I added to the code there makes sense.

image


gngilbert:
Thanks. Will the new version be on the GitHub in the morning?

tarheel:
Depends on when Jon or Hylton has a chance to review my changes. But I can send you an updated JAR before then if you want to test it.

tarheel:
Hmm, not sure why that test is failing for you. All the tests are passing for me. I'll double-check it, though.

And yeah, I'll rename the file.

HEdingfield:
I assume they all passed when you checked again? Very strange...

tarheel:
> I assume they all passed when you checked again? Very strange...

Yeah. I think you'll have to debug this on your end. Must be something environment-specific...?

HEdingfield:
Figured out that the date was a day off in my local version of the config file... very strange that pulling the master down didn't overwrite that, but oh well, fixed now.

PR #223, opened 2019-02-11
Cleanup and test fix

HEdingfield:


src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -60,11 +60,12 @@
 import javafx.stage.FileChooser.ExtensionFilter;
 import javafx.util.StringConverter;
 
+@SuppressWarnings("WeakerAccess")
```
tarheel: what is this?

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -350,18 +354,25 @@ public void buttonDeleteCandidateClicked() {
   @Override
   public void initialize(URL location, ResourceBundle resources) {
     Logger.addGuiLogging(this.textAreaStatus);
-    Logger.log(Level.INFO, "Opening tabulator GUI.");
+    Logger.log(Level.INFO, "Opening tabulator GUI...");
 
     String helpText;
     try {
+      //noinspection ConstantConditions
```
tarheel: and what's this?

src/main/java/com/rcv/Tabulator.java
```diff
@@ -461,8 +461,7 @@ private CandidateStatus getCandidateStatus(String candidate) {
       eliminated.add(label);
       Logger.log(
           Level.INFO,
-          "Eliminated %s in round %d because it represents undeclared write-ins. It had "
-              + "%s votes.",
+          "Eliminated candidate \"%s\" in round %d because it represents undeclared write-ins. It had %s votes.",
```
tarheel: Here and elsewhere: I thought we were supposed to limit lines to 100 characters. Are you proposing changing it to a higher value?

src/main/java/com/rcv/TallyTransfers.java
```diff
@@ -15,7 +15,7 @@
  *
  * TallyTransfers is a container class which stores a map for each round showing how many votes were
  * transferred from each candidate to each candidate.
- * The primary purpose for this is generating sankey plots which visually show the flow of votes
+ * The primary purpose for this is generating Sankey plots which visually show the flow of votes
```
tarheel: s/which/that/

src/test/java/com/rcv/TabulatorTests.java
```diff
@@ -126,10 +126,14 @@ private static void runTabulationTest(String stem) {
...
-    File ouputFolder = new File(session.outputPath);
-    for (File file : ouputFolder.listFiles()) {
-      if (!file.isDirectory()) {
-        file.delete();
+    File outputFolder = new File(session.outputPath);
+    if (outputFolder.listFiles() != null) {
+      //noinspection ConstantConditions
+      for (File file : outputFolder.listFiles()) {
+        if (!file.isDirectory()) {
+          //noinspection ResultOfMethodCallIgnored
```
tarheel: Is this some weird IntelliJ thing? Seems kind of overboard to have to add this kind of comment every time you don't want to use a method's return value...

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: It suppresses future incorrect / unnecessary IntelliJ Code Inspection warnings.

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: Ditto.

src/main/java/com/rcv/Tabulator.java
HEdingfield: Very weird that IntelliJ isn't correcting this during file reformats (in fact, it looks like it's doing the opposite). I'm just going to go with it since it's easier to go with the automated recommendation than to fight it.

src/main/java/com/rcv/TallyTransfers.java
HEdingfield: Done in current commit I'm working on in another branch.

src/test/java/com/rcv/TabulatorTests.java
HEdingfield: Yes it is. Luckily it only occurs in this one spot. FWIW our code base at work is rife with these sorts of suppression statements.

src/main/java/com/rcv/Tabulator.java
tarheel: Sigh. This seems like another case in which there's a mysterious inconsistency between your machine, Jon's, and mine in how IntelliJ is applying the style rules. For me, it complains whenever someone exceeds 100 characters.

HEdingfield: Ping pong :(
At least it doesn't affect the functionality.

Issue #207, opened 2019-01-07
enable editing of candidates and CVR source files in GUI


tarheel:
Currently this isn't possible. If you want to edit an item, you have to remove it and re-add it.

HEdingfield:
No out-of-the-box support for this with the `TableView` class unfortunately, but there are a [couple](https://www.youtube.com/watch?v=_FBPukQOLCY) of [videos](https://www.youtube.com/watch?v=z4LVoLg6ch0) that demo this.

gngilbert:
While it would be very useful to me in testing, I agree that it is not critical at this point.  However, at the point where independent users are using it regularly, it would make a big difference to some of them.  We can put it on hold for now if it is problematic but let's not forget it.  thanks.

CalebKleppner:
UI that changes would be nice at some point:  a) an edit button on the candidates tab so that you can change the spelling of a name, add or change a code and switch a candidate to or from excluded.  Generally, we just enter a candidate name and then click “Add”, so the add button should be immediately to the right of Name * (and one tab away).  Then we should have the box for Code, the check box for “Excluded”, then “Edit Selected” and finally “Delete Selected”. So the final order would be:

     Name *, Add, Code, Excluded, Edit Selected, Delete Selected

b) Any way to re-order candidates?  Drag-n-drop would be nice ...

![image](https://user-images.githubusercontent.com/41450076/51409682-8e392280-1b30-11e9-94d2-107cdf5b2e7f.png)

c) On the CVR Files tab, is it worth adding an Edit button and possibly a method of reordering CVR files?

![image](https://user-images.githubusercontent.com/41450076/51409701-97c28a80-1b30-11e9-818d-a2d330e553f2.png)

gngilbert:
All sounds good to me.

George Gilbert
RCV Resource Center
george.gilbert@rankedchoicevoting.org
336-906-0047

gngilbert:
We have been approached by ES&S regarding their interest in partnering with us to offer several Utah jurisdictions a combination of ES&S voting system and the UTab Software to conduct RCV elections this spring.  As I understand it, both Utah jurisdictions with spring elections include multi-winner offices that will likely employ the "sequential tabulation" process.  It would definitely be best if we could offer this capability with the ability to edit a candidate's "excluded" status.  

[...]

PR #224, opened 2019-02-11
GUI list edit

HEdingfield:
Partially addresses #207. This is currently not completed. I had a lot of trouble figuring out how to make the non-String fields (boolean checkboxes and integer fields) editable. Will need to get that sorted before pulling, but wanted to share in the meantime in case anyone else had ideas.


HEdingfield:
This should be all set to review now. Once this is merged in #207 should be fixed.

src/main/java/com/rcv/JsonParser.java
```diff
@@ -33,8 +33,9 @@
   // purpose: parse input json file into an object of the specified type
   // param: jsonFilePath path to json file to be parsed into java
   // param: valueType class of the object to be created from parsed json
+  // param: logsEnabled display log messages
   // returns: instance of the object parsed from json or null if there was a problem
-  static <T> T readFromFile(String jsonFilePath, Class<T> valueType) {
+  static <T> T readFromFile(String jsonFilePath, Class<T> valueType, boolean logsEnabled) {
```
tarheel: Boolean args tend to be confusing in public methods. Would prefer to expose a separate wrapper method `readFromFileWithoutLogging` for clarity.

HEdingfield: Changed, k picante!

PR #228, opened 2019-02-24
handle residual surplus properly

### LINKED ISSUE #215: handle residual surplus
Once #214 is in, we'll have the possibility of "residual surplus": because we round down fractional vote transfers, we may end up subtracting slightly less than the full surplus from a winning candidate. For now we just leave that extra bit in the candidate's vote total, but @gngilbert has suggested that we could remove it and transfer it to a separate "residual surplus" tally.

tarheel:
Fixes #215.

Note on testing: since our test framework only looks at the JSON output and this residual surplus tally is only displayed in the CSV output, we're not fully testing this change... but I did review the CSV output manually for a few cases and verify that it makes sense, and the JSON output does test that we're setting each winner's total to the threshold value (i.e. removing the residual surplus from their total).


src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -261,26 +268,39 @@ private void generateSummarySpreadsheet(
...
       // total votes cell
-      csvPrinter.print(thisRoundExhausted.toString());
+      csvPrinter.print(thisRoundInactive.toString());
     }
     csvPrinter.println();
 
+    // row for residual surplus (if needed)
+    Optional<BigDecimal> maxResidual =
+        roundToResidualSurplus.values().stream().max(BigDecimal::compareTo);
+    if (maxResidual.isPresent() && maxResidual.get().signum() == 1) {
```
moldover: can you clarify what this line means?

moldover:
Just want to make sure I understand the concept: residual surplus is the difference between a winning threshold, and a candidates' surplus fractional vote transferred to other candidates.  The difference may arise because surplus fraction is calculated using rounding, and the surplus with be the amount of rounding "error".

src/main/java/com/rcv/ResultsWriter.java
tarheel: Just testing whether there's a positive value present. This is one of those cases where I know how to write it in an elegant way in Ruby, but in Java it's surprisingly complicated.

https://docs.oracle.com/javase/10/docs/api/java/math/BigDecimal.html#signum()

But upon further consideration... I can just test the final value, because the values increase monotonically. That will be simpler to read.

tarheel:
> Just want to make sure I understand the concept: residual surplus is the difference between a winning threshold, and a candidates' surplus fractional vote transferred to other candidates. The difference may arise because surplus fraction is calculated using rounding, and the surplus with be the amount of rounding "error".

Yeah. To state it another way: if a candidate wins with X votes and the winning threshold is Y, the surplus is (X-Y). But the total amount of surplus we transfer, Z, might not equal X-Y (due to rounding), so the residual surplus is X-Y-Z.

PR #231, opened 2019-02-26
add support for sequential multi-seat tabulation

tarheel:
Fixes #229. I had to do a decent amount of refactoring of `TabulatorSession` and `TabulatorTests` to get everything working properly.

I highly recommend reviewing each commit independently, as the PR breaks down neatly into three incremental steps (add the config option, make it work, and add a test for it).

There's also a small commit that fixes missing whitespace in a couple existing JSON files (which was causing test failures). @HEdingfield, didn't you run into this same issue a few weeks ago? It's the strangest thing.

PR #232, opened 2019-03-01
fix crazy issue with JSON whitespace

tarheel:
There's some bizarre issue arising from the bowels of the Jackson code that's causing it to be inconsistent about whether it uses " : " or ": " as the object/value separator. I'm not 100% sure this change will fix it because I still don't actually understand the source of the problem... but being explicit about what we want it to do seems like a promising approach to try. It seems to work so far.


tarheel:
I don't think this just started with #231 (which I haven't committed yet), but I don't know how it started or why it's happening.

Arhgghghgh. I feel like I'm wasting a lot of time on this, and it's not even solving a real problem in our code; it's just a false alarm in our tests. An alternative would be to give up on trying to understand why this is happening, at least for the moment, and just change our test code so that it doesn't try to do a character-by-character file comparison -- and instead compares the contents of the files when they're read back into Maps. What do you think?

tarheel:
@HEdingfield: putting the curly brace on the same line that the square bracket is on is how all of the JSON files in master are formatted. Are you saying that it doesn't do that if you run the code from this PR, but it does if you run the code from master?

tarheel:
And to answer your other question: all of the other uses of `DefaultPrettyPrinter` are for the config file, not the expected output. I do think we've seen this same issue with the configs at some points (seemingly arbitrary use of " : " vs. ": "), so if we figure out what's happening in the tests, we should propagate that fix to the rest of the code.

tarheel:
Wow. I'm now getting the same failure that you're getting. I ran the exact same code last night and the tests passed. This is... fascinating.

tarheel:
Oh, I have a theory that might explain part of this: the IntelliJ commit hooks might be modifying the JSON formatting. That would explain why the tests are failing for me now, because last night I tested, then committed, but I didn't test again after that.

tarheel:
OK, yeah, I'm pretty confident that IntelliJ is the culprit for all this nonsense. I looked for a bit and couldn't figure out how to either a) tell it to ignore JSON entirely when it reformats, or b) make its rules identical to what DefaultPrettyPrinter is doing. For now, I'm just going to be careful not to let it touch JSON files. I'll open a separate issue for the more robust solution.

tarheel:
bitmoji

HEdingfield:
What a wacky adventure... thanks for getting to the bottom of it!

PR #234, opened 2019-03-02
add Hare quota support

tarheel:
Since I had to do the work for this, might as well include it. We can hide it from the UI or remove it entirely later if we decide it's not worth having.


src/main/resources/config_file_documentation.txt
```diff
@@ -144,6 +144,13 @@ Config file must be valid JSON format. Examples can be found under the test fold
         if true, threshold = V/(S+1) + 10^-d
         if false, threshold = floor(V/(S+1)) + 1
         where V = total number of votes (in the current round for single-seat or in the first round for multi-seat); S = numberOfWinners; and d = decimalPlacesForVoteArithmetic
+        (note that S+1 in the formulas above becomes just S if hareQuota is set to true)
+        value: true | false
+        if not supplied: false
+
+      "hareQuota" optional
+        the winning threshold should be computed using the Hare quota (votes divided by seats) instead of the preferred Droop quota (votes divided by seats+1)
```
HEdingfield:
```suggestion
        the winning threshold should be computed using the Hare quota (votes divided by seats) instead of the preferred Droop quota (votes divided by (seats+1))
```

src/main/java/com/rcv/Tabulator.java
```diff
@@ -358,8 +358,13 @@ private void setWinningThreshold(Map<String, BigDecimal> currentRoundCandidateTo
       currentRoundTotalVotes = currentRoundTotalVotes.add(numVotes);
     }
 
-    // divisor for threshold is num winners + 1
-    BigDecimal divisor = new BigDecimal(config.getNumberOfWinners() + 1);
+    // divisor for threshold is num winners + 1 (unless archaic Hare quota option is enabled, in
```
HEdingfield: > archaic

ooo I hope they got aloe for that burn!

src/main/java/com/rcv/ContestConfig.java
```diff
@@ -386,6 +386,11 @@ private void validateRules() {
         isValid = false;
         Logger.log(Level.SEVERE, "batchElimination can't be true in a multi-winner contest!");
       }
+    } else {
+      if (isHareQuotaEnabled()) {
+        isValid = false;
+        Logger.log(Level.SEVERE, "hareQuota can't be true in a single-seat contest!");
```
HEdingfield: Slight change in wording since this will trigger if `getNumberOfWinners()` is null or <= 1.
```suggestion
        Logger.log(Level.SEVERE, "hareQuota can only be true in a multi-seat contest!");
```

src/main/java/com/rcv/Tabulator.java
tarheel: bitmoji

Issue #242, opened 2019-03-30
Support NIST CDF input format


moldover:
CDF format includes not just votes (like ES&S files) but also the election data: candidates and contests.  Reading CDF format will involve first reading the election data to build the list of candidates / contests.  Then read the ballots, and match up ballot selections with candidates / contests.  Example file is attached.  A config file will of course be needed to specify all the other tabulator options.

I can imagine one solution which adds a flag to the config file to indicate that candidate info will be parsed from the CVR file at runtime.  Then the tabulator will open input CDF file and extract the candidate data and use it to populate the RawContestConfig candidates list.  Then a tabulation is performed.  

[Multi-winner_NIST_CVR_3.txt](https://github.com/BrightSpots/rcv/files/3025788/Multi-winner_NIST_CVR_3.txt)

tarheel:
I agree that the latter would be cleaner.

moldover:
@tarheel does it make sense to split this into the following?  If so, wanna pick one?  
1) parsing election data from CDF and populating rawConfig
2) parsing vote data into CastVoteRecord objects

tarheel:
There's some overlap for reading in the file, yeah? Would it be easiest to do these two sequentially? Anyway, I probably can't make progress until Tuesday or Wednesday, so you can decide how to start.

moldover:
Yeah maybe a bit over overlap.  Tell ya what, I've already made some progress on 2, so how about I wrap that up and push it tonight (Tuesday) and you can take over.  At that point I'll switch to working on the output, unless Hylton or Hemal are interested in doing that.

PR #245, opened 2019-04-04
CDF Reader

moldover:
Addresses #242 CDF input for the example file.  
Added new unit test which tabulates CDF file.
CVR Source must be set to "CDF" to trigger CDF parsing (otherwise defaults to ES&S).
CDF files must be tabulated individually (too many weird edge cases could arise otherwise).
If CDF is specified candidates are not required to be in config file.
On config load, CDF file is processed to extract candidate names and add them to candidates list.


tarheel:
Nice. I'll check this out tomorrow.

src/main/java/com/rcv/CommonDataFormatReader.java
```diff
@@ -0,0 +1,162 @@
...
+                String selectionID = (String) contestSelection.get("ContestSelectionId");
+                // extract all the positions (ranks) which this selection has been assigned
+                ArrayList selectionPositions = (ArrayList) contestSelection
+                    .get("SelectionPosition");
+                for (Object selectionPositionObject : selectionPositions) {
+                  // extract the position object
+                  HashMap selectionPosition = (HashMap) selectionPositionObject;
+                  // and finally the rank
+                  Integer rank = (Integer) selectionPosition.get("Rank");
+                  assert rank != null;
```
tarheel: and >= 1?

src/main/java/com/rcv/CommonDataFormatReader.java
```diff
@@ -0,0 +1,162 @@
...
+                  // create a new ranking object and save it
+                  rankings.add(new Pair<>(rank, selectionID));
+                }
+              }
+            }
+
+            // create new cast vote record
+            CastVoteRecord newRecord = new CastVoteRecord(computedCastVoteRecordID,
+                ballotID, null, null, rankings);
```
tarheel: I guess there's no `fullCVRData` equivalent here? Will our existing audit log just not be compatible with CDF input?

src/main/java/com/rcv/TabulatorSession.java
```diff
@@ -188,8 +188,14 @@ private boolean setUpLogging(ContestConfig config) {
       Logger.log(Level.INFO, "Reading cast vote record file: %s...", cvrPath);
       // the CVRs parsed from this source
       try {
-        List<CastVoteRecord> cvrs =
-            new StreamingCVRReader(config, source).parseCVRFile(castVoteRecords, precinctIDs);
+        List<CastVoteRecord> cvrs;
+        if(source.getProvider().equals("CDF")) {
+          CommonDataFormatReader reader = new CommonDataFormatReader(cvrPath);
+          cvrs = reader.parseCVRFile(castVoteRecords);
```
tarheel: I just noticed that this API is weird. The caller passes in a collection, and then for every CVR that the method parses, it adds it both to the passed-in collection and to a new collection that it returns.

It would be more intuitive if the caller didn't pass in the collection. When the method returns `cvrs`, we can just call `castVoteRecords.addAll(cvrs)` here.

src/main/java/com/rcv/ContestConfig.java
```diff
@@ -84,10 +84,18 @@
...
+  private ContestConfig(RawContestConfig rawConfig, String sourceDirectory) {
     this.rawConfig = rawConfig;
     this.sourceDirectory = sourceDirectory;
-    this.processCandidateData();
+  }
+
+  // function: loadContestConfig
+  // purpose: create ContestConfig from pre-populated rawConfig and default folder
+  // returns: new ContestConfig object if checks pass otherwise null
+  static ContestConfig loadContestConfig(RawContestConfig rawConfig, String sourceDirectory) {
```
tarheel: Thanks. This is better.

src/main/java/com/rcv/ContestConfig.java
```diff
@@ -119,6 +127,7 @@ static ContestConfig loadContestConfig(String configPath) {
         Logger.log(Level.SEVERE, "Failed to create contest config!");
       }
     }
+    config.processCandidateData();
```
tarheel: Instead of adding this here, you should refactor the lines immediately above here (where you call the constructor) to just call the other `loadContestConfig` instead.

src/main/java/com/rcv/ContestConfig.java
```diff
@@ -203,6 +213,21 @@ private void validateCvrFileSources() {
...
 
+        // perform CDF checks
+        if (source.getProvider().equals("CDF")) {
+          if (rawConfig.cvrFileSources.size() != 1) {
+            isValid = false;
+            Logger.log(Level.SEVERE, "CDF files must be tabulated individually.");
+          }
+          if (isTabulateByPrecinctEnabled()) {
+            isValid = false;
+            Logger.log(Level.SEVERE, "Tabulate By Precinct may not be used with CDF files.");
```
tarheel: nit: tabulateByPrecinct

src/main/java/com/rcv/ContestConfig.java
```diff
@@ -203,6 +213,21 @@ private void validateCvrFileSources() {
...
+          }
+          continue;
```
tarheel: Instead of this, why not just put the ES&S checks in an `else` block?

src/main/java/com/rcv/CommonDataFormatReader.java
```diff
@@ -0,0 +1,162 @@
...
+  // purpose: class constructor
+  // param: filePath source file to read
+  CommonDataFormatReader(String filePath) {
+    this.filePath = filePath;
+  }
+
+  // function: getCandidates
+  // purpose: returns list of candidates parsed from CDF election json
+  Set<String> getCandidates()
+  {
```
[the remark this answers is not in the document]
moldover: Yeah, I certainly do.  I don't understand why it doesn't work the same for me.

src/main/java/com/rcv/CommonDataFormatReader.java
```diff
@@ -0,0 +1,162 @@
...
+    // container for results
+    Set<String> candidates = new HashSet<>();
+    try {
+      // try to read in the file
+      HashMap json = JsonParser.readFromFile(filePath, HashMap.class);
```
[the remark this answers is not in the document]
moldover: It could be, but that ends up producing Unckecked assignment warnings.  Without getting deeper into Java generics, this approach removed the warnings and works reasonably.

src/main/java/com/rcv/CommonDataFormatReader.java
```diff
@@ -0,0 +1,162 @@
...
+      // top-level election is a list of election objects:
+      ArrayList electionArray = (ArrayList) json.get("Election");
+      for(Object electionObject : electionArray) {
+        HashMap election = (HashMap) electionObject;
+        // for each election get the contests it contains
+        ArrayList contestArray = (ArrayList) election.get("Contest");
+        for(Object contestObject : contestArray) {
```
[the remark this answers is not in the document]
moldover: good call

src/main/java/com/rcv/TabulatorSession.java
moldover: Yeah, totally agree.  I think it got to the current state incrementally with no real problem, but it should be easy enough to fix.

src/main/java/com/rcv/ContestConfig.java
moldover: dig it

src/main/java/com/rcv/ContestConfig.java
moldover: yeah that's more readable

src/main/java/com/rcv/CommonDataFormatReader.java
```diff
@@ -0,0 +1,162 @@
...
+import java.io.File;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.logging.Level;
+import javafx.util.Pair;
+
+class CommonDataFormatReader {
```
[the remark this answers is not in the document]
moldover: Yeah, I punted on this.  Won't take long but I'd rather get through the additional tests first.

moldover:
Just realized you had some more comments I didn't see (thanks Github!).  Will address them later.

tarheel:
> Will address them later.

Do you mean that I should wait for you to make them here, or that you'll do them in a future PR?

tarheel:
I'm still mystified by GitHub's behavior of seemingly arbitrarily hiding some review comments...

src/main/java/com/rcv/CommonDataFormatReader.java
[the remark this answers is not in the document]
tarheel: It behaves slightly differently for me if I run it manually vs. as a commit hook. But something like this or `if(X)` it should catch either way...

src/main/java/com/rcv/CommonDataFormatReader.java
```diff
@@ -0,0 +1,162 @@
...
+
+  // function: parseCVRFile
+  // purpose: parse the given file into a List of CastVoteRecords for tabulation
+  // param: castVoteRecords existing list to append new CastVoteRecords to
+  // returns: list of parsed CVRs
+  List<CastVoteRecord> parseCVRFile(List<CastVoteRecord> castVoteRecords) {
+
+    List<CastVoteRecord> cvrs = new ArrayList<>();
+    // cvrIndex and fileName are used to generate IDs for cvrs
+    Integer cvrIndex = 0;
```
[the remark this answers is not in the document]
moldover: fixed

src/main/java/com/rcv/CommonDataFormatReader.java
```diff
@@ -0,0 +1,162 @@
...
+      ArrayList CVRs = (ArrayList) json.get("CVR");
+      // for each CVR object extract the current snapshot
+      // it will be used to build the ballot data
+      for (Object CVR : CVRs) {
+        HashMap CVRObject = (HashMap) CVR;
+        String currentSnapshotID = (String) CVRObject.get("CurrentSnapshotId");
+        // get ballotID
+        String ballotID = (String) CVRObject.get("BallotPrePrintedId");
+        // compute internal ballot ID
+        String computedCastVoteRecordID = String.format("%s(%d)", fileName, ++cvrIndex);
```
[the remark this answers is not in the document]
moldover: Yep

src/main/java/com/rcv/CommonDataFormatReader.java
```diff
@@ -0,0 +1,162 @@
...
+
+        ArrayList CVRSnapshots = (ArrayList) CVRObject.get("CVRSnapshot");
+        for (Object snapshotObject : CVRSnapshots) {
+          HashMap snapshot = (HashMap) snapshotObject;
+          if (snapshot.get("@id").equals(currentSnapshotID)) {
```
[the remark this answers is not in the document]
moldover: This is of course hypothetical, but when I wrote that I was imagining a scenario where a scanner produces an "original" snapshot which has just the raw marks (maybe density or whatever that scanner's thing is) THEN it applies some rules like party-line selection and produces a second snapshot which is "modified" to have all a particular parties selected.  THEN it is handed off to a tabulator.

I will break up that function, it's a bit lengthy and yes, indented :(

src/main/java/com/rcv/CommonDataFormatReader.java
moldover: I guess?  Never 0-index anything!

src/main/java/com/rcv/CommonDataFormatReader.java
moldover: The idea is that CDF _is_ the audit log.

moldover:
Ok, I've addressed the comments I missed before and this is ready.  I made a new issue for refactoring the CVR reader classes so that is not included.

Issue #243, opened 2019-03-31
Support NIST CDF tabulation output


moldover:
The CDF format can indicate the results of a tabulation.  We need to support this to be a reference application.  For each round, we output the status of every ballot: who it counts for, and how much.  In the CDF json file, this is stored as a CVRSnapshot: a "snapshot" in time of CVR state, which changes during tabulation.  

Note: a CVRSnapshot is added to every CVR in the file, for every round, even if a particular CVR doesn't change state in a given round.  Yes, this means a LOT of duplication but we're not going to worry about that for now.  There is no live use-case yet, let alone anyone to asking for smaller file sizes.  This may be addressed when the standard itself is updated.

JSON details:

The structure of the CDF output is based directly on the input sample file attached here.  I am still working on creating a sample output file and will attach or commit it once it is ready.
[Multi-winner_NIST_CVR_INPUT.txt](https://github.com/BrightSpots/rcv/files/3026731/Multi-winner_NIST_CVR_INPUT.txt)

As mentioned, each CVRSnapshot represents the state of a particular CVR in a particular round.  The order of the snapshots implies the order of rounds.

Within the CVRSnapshot is a list of CVRContestSelections, one for each ranking a voter selected.  Each CVRContestSelection has a SelectionPosition object which contains the critical fields:
  **Rank: [positive integer rank]**  What is the ranking for this position (won't change)
  **NumberVotes: [non-negative number]** What is the vote value for this selection (could become fractional in multi-seat fractional transfer scenarios)
**IsAllocable: ["yes" | "no" | "unknown"]** Does this selection count in this round?  This will change when a vote is initially counted for a candidate or transfers.  When a vote is exhausted, ALL SelectionPosition objects will have IsAllocable:"no"

tarheel:
When do you set isAllocable to unknown?

moldover:
unknown is used by a ballot marking/creation device (e.g. at a polling place) when the ballot is first created, before any election rules are applied (e.g. no overvote determination has been applied).  Some other device (e.g. tabulator) or person (adjudicator) needs to apply specific election rules and determine how this cvr will actually be counted over the course of an election.

tarheel:
So am I understanding this correctly? In the output, for each round we'll have a full copy of the CVR snapshot for each ballot, where the only difference from round to round is that IsAllocable and NumberVotes may change?

A couple more thoughts:
- As you mentioned, the amount of duplication here is just massive.
[...]

PR #246, opened 2019-04-04
generate CVR CDF JSON

tarheel:
First attempt on #243.

I tested it with a (very small) single-seat contest and a multi-seat contest and they both seemed to work fine. I'm sure we'll need to iterate on some of the details, though. And there are some fields (e.g. the election ID and contest ID) that I'm just using constants for now, but that we should actually be propagating from the input file if it's also using the CDF.

I made some effort to minimize the memory footprint, but there's room to make it more compact if necessary. Right now I imagine it will not scale well at all; even if we can keep the data in CastVoteRecord to a reasonable size, the size of the output JSON (represented as a Map) that we're preparing in memory grows extremely quickly. If we actually want to be able to generate this for an election with many thousands of ballots, I think we'll need to rewrite the JSON output logic to use a streaming solution. Something like this: https://static.javadoc.io/com.google.code.gson/gson/2.6.2/com/google/gson/stream/JsonWriter.html

Anyway, here's what my code generates for the multi-winner redistribution test that's already in our test data: https://jsoneditoronline.org/?id=d6a274e2ae9641b79cbe16d1cc19d81e

I recommend reviewing the three commits one at a time.

Super nit that I noticed: we're not consistent in our code about the casing of initialisms. I'm not even consistent within this single PR. E.g. should it be generateCvrSnapshotID, generateCvrSnapshotId, generateCVRSnapshotID, or g3n3rAt3cVrSnApSh0tId? We should just agree on this and then clean it up in a future PR.

### DIFF
--- a/src/main/java/com/rcv/CastVoteRecord.java
+++ b/src/main/java/com/rcv/CastVoteRecord.java
@@ -23,8 +23,10 @@ package com.rcv;
 import java.math.BigDecimal;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
@@ -50,6 +52,12 @@ class CastVoteRecord {
   private boolean isExhausted;
   // tells us which candidate is currently receiving this CVR's vote (or fractional vote)
   private String currentRecipientOfVote = null;
+  // If CVR CDF output is enabled, we store the necessary info here: for each round, the list of
+  // candidates this ballot is counting toward (0 or 1 in a single-seat contest; 0 to n in a
+  // multi-seat contest because of fractional vote transfers), and how much of the vote each is
+  // getting. As a memory optimization, if the data is unchanged from the previous round, we don't
+  // add a new entry.
+  private Map<Integer, List<Pair<String, BigDecimal>>> cdfSnapshotData = new HashMap<>();
 
   // function: CastVoteRecord
   // purpose: create a new CVR object
@@ -118,6 +126,23 @@ class CastVoteRecord {
 
     // output with level FINE routes to audit log
     Logger.log(Level.FINE, logStringBuilder.toString());
+  }
+
+  Map<Integer, List<Pair<String, BigDecimal>>> getCdfSnapshotData() {
+    return cdfSnapshotData;
+  }
+
+  void logCdfSnapshotData(int round) {
+    List<Pair<String, BigDecimal>> data = new LinkedList<>();
+    for (Entry<String, BigDecimal> entry : winnerToFractionalValue.entrySet()) {
+      // TODO: can we avoid duplicating this in memory?
+      data.add(new Pair(entry.getKey(), entry.getValue()));
+    }
+    if (currentRecipientOfVote != null) {
+      data.add(new Pair(currentRecipientOfVote, getFractionalTransferValue()));
+    }
+
+    cdfSnapshotData.put(round, data);
   }
 
   // function: exhaust
--- a/src/main/java/com/rcv/CastVoteRecord.java
+++ b/src/main/java/com/rcv/CastVoteRecord.java
@@ -132,6 +132,8 @@ class CastVoteRecord {
     return cdfSnapshotData;
   }
 
+  // purpose: store info that we'll need in order to generate the CVR JSON snapshots in the Common
+  // Data Format at the end of the tabulation (if this option is enabled)
   void logCdfSnapshotData(int round) {
     List<Pair<String, BigDecimal>> data = new LinkedList<>();
     for (Entry<String, BigDecimal> entry : winnerToFractionalValue.entrySet()) {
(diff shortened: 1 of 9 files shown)


src/main/java/com/rcv/CastVoteRecord.java
```diff
@@ -50,6 +52,12 @@
   private boolean isExhausted;
   // tells us which candidate is currently receiving this CVR's vote (or fractional vote)
   private String currentRecipientOfVote = null;
+  // If CVR CDF output is enabled, we store the necessary info here: for each round, the list of
+  // candidates this ballot is counting toward (0 or 1 in a single-seat contest; 0 to n in a
+  // multi-seat contest because of fractional vote transfers), and how much of the vote each is
+  // getting. As a memory optimization, if the data is unchanged from the previous round, we don't
+  // add a new entry.
+  private Map<Integer, List<Pair<String, BigDecimal>>> cdfSnapshotData = new HashMap<>();
```
moldover: does it make sense to combine this with winnerToFractionalValue?

src/main/java/com/rcv/CastVoteRecord.java
```diff
@@ -116,6 +128,23 @@ void logRoundOutcome(
     Logger.log(Level.FINE, logStringBuilder.toString());
   }
 
+  Map<Integer, List<Pair<String, BigDecimal>>> getCdfSnapshotData() {
+    return cdfSnapshotData;
+  }
+
+  void logCdfSnapshotData(int round) {
```
moldover: comments please

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -447,6 +470,176 @@ private String getPrecinctFileString(String precinct, Set<String> filenames) {
...
+            Map.ofEntries(
+                entry("@id", CDF_REPORTING_DEVICE_ID), entry("@type", "CVR.ReportingDevice"))
+        });
+    outputJson.put("Version", "1.0.0");
+    outputJson.put("@type", "CVR.CastVoteRecordReport");
+
+    generateJsonFile(outputPath, outputJson);
+  }
+
+  private List<Map<String, Object>> generateCdfMapForCvrs(List<CastVoteRecord> castVoteRecords) {
```
moldover: comments please.

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -447,6 +470,176 @@ private String getPrecinctFileString(String precinct, Set<String> filenames) {
...
+              entry("CurrentSnapshotId", generateCvrSnapshotID(cvr.getID(), numRounds)),
+              entry("CVRSnapshot", cvrSnapshots),
+              entry("ElectionId", CDF_ELECTION_ID),
+              entry("@type", "CVR.CVR")));
+    }
+
+    return cvrMaps;
+  }
+
+  private Map<String, Object> generateCvrSnapshotMap(
```
moldover: comments please.

src/main/java/com/rcv/CastVoteRecord.java
tarheel: I stopped adhering to the "comments for everything" rule a long time ago because it wasn't clear to me whether we'd actually end up having to follow that convention, it clutters up the code, and it's easy to fix later if we decide that we do need them. So, if it's OK with you, for now I'd prefer only to add them in places where you think it's actually unclear what's happening.

src/main/java/com/rcv/CastVoteRecord.java
tarheel: Did you delete this comment?

> does it make sense to combine this with winnerToFractionalValue?

This is actually a good question. There's definitely overlap between the two data structures. The downside of combining them would be that the new structure would contain a fair amount more data than what winnerToFractionalValue currently holds. If we expect that normal operation will often not enable the CDF output, it's better to keep these two separate. If, on the other hand, we expected CDF output to be enabled usually or always, then yes, we would want to combine them.

src/main/java/com/rcv/CastVoteRecord.java
moldover: Yeah, for sure.  I'm asking here and elsewhere because it's unclear what is happening.  Just a minimum of one line description of what is going on, and think of the audience as outside devs reading this for the first time.

Side note:  I'm continually surprised by how many veteran devs and managers don't encourage this practice.  A good dev will figure things out sooner or later, but the time savings from judicious comments is significant.

src/main/java/com/rcv/CastVoteRecord.java
moldover: Yeah, and that's why I deleted it.  I do not expect anyone to be using CDF tabulation for quite some time.

src/main/java/com/rcv/CastVoteRecord.java
tarheel: Judicious, sure. Maybe I'm overreacting to the trauma of having to comment on every single variable and function.

How do you go on, when in your heart you begin to understand... there is no going back? There are some things that time cannot mend. Some hurts that go too deep.

moldover:
Looks good!  Regarding performance, streaming output makes sense, but not worth addressing now.  I don't think we'll need to worry about that until some vendors or jurisdictions are using CDF.  
Regarding initialisms, how do you go on, when in your heart you begin to understand... there is no going back?

src/main/java/com/rcv/CastVoteRecord.java
moldover: I am probably overreacting to the trauma of the code I'm working on at work :/

PR #250, opened 2019-04-06
fix additional CDF issues

tarheel:
Two more things that John Dziurlaj found in our JSON output:
1. BigDecimals should be numbers, not strings. I realized that this actually works properly with the default behavior. @moldover: do you remember why you added the custom logic for BigDecimals in the first place?
2. IDs can only use certain characters.


moldover:
Great, thanks.  The JSON standard doesn't actually specify precision for [numeric values](https://stackoverflow.com/questions/35709595/why-would-you-use-a-string-in-json-to-represent-a-decimal-number) so what John D. is suggesting--using doubles--is both impossible and logically wrong.  That said, in the interest of practicality, I am fine with a quick and simple (if not 100% correct) solution like using doubles.  I'll respond to the thread with this info so they can make an informed decision.

tarheel:
OK. I'll go ahead with this for now and we can always revert the BigDecimal thing later if we convince John that doubles aren't appropriate.

Issue #244, opened 2019-03-31
Add unit tests for NIST RCV tabulation


moldover:
@gngilbert has already tested these successfully.  Adding them as unit tests involves add the source files (cvr and config) into the test resources area, and then creating unit tests which run them automatically.

tarheel:
@HEdingfield did you want to grab this? I can do it soon if you won't have time for it.

tarheel:
All of the files are attached to an email from George on 4/1 in the thread "NIST Common Data Format Project."

HEdingfield:
Slammed for the next week catching up with budget stuff and doing taxes, but if this is still kicking around in a couple weeks I can.

tarheel:
I'll update all of the test sets to check the CDF output also once that logic has stabilized a bit more. Right now we're still getting feedback from John et al on things we need to change.

moldover:
@tarheel working on this today, good progress, (basically adding precinct ids, which is what GpUnit is for) but hit a snag: there's some error when running 2017 Minneapolis Mayor Precinct when enabling CDF output.  Seems like some CVRs never get recordSelectionForCastVoteRecord called on them, hence they never get logCdfSnapshotData called, hence ResultsWriter has a failed assertion when it can't find any cdf data at all.  I'm working around it for now, will circle back if you don't figure it out before I do.  Maybe they are just exhausted in the first round?

tarheel:
Hmm, interesting. Yes, I'll debug.

PR #253, opened 2019-04-07
convert test input files to CDF

tarheel:
Completes #244.

Added the ability to convert XLSX to CDF, then ran it for all the test sets and updated each of their config files to point to the CDF JSON. I can verify that this works properly once #245 is committed.


src/main/java/com/rcv/TabulatorSession.java
```diff
@@ -55,12 +55,35 @@
...
   }
 
   String getTimestampString() {
     return timestampString;
   }
 
+  // purpose: special mode to just export the CVR as CDF JSON instead of tabulating
+  void convertToCdf() {
+    ContestConfig config = ContestConfig.loadContestConfig(configPath);
+    if (config != null && config.validate()) {
```
moldover: validate is done for you by loadContestConfig

moldover: @tarheel I'm going to push a couple changes to this branch: 
merging in master to get the CDFReader
small fix to the config files (CDFReader expects an empty candidates list).
There's still the old json formatting issue where where you ran the formatting code on expected json files and it doesn't match, but I think it's close now.

moldover: oh... actually i think there are 2 problems (maybe I'm missing something)
1 no way to indicate cdf output in a config file on disk (only through ui which will not work with tests)
2 no way to indicate in tests that cdf output file should be compared (currently test always compares to summary json)

tarheel:
> small fix to the config files (CDFReader expects an empty candidates list)

You mean it will fail if you just omit that field entirely, as I had done? That feels a little unintuitive. Any objection if I change it?

> oh... actually i think there are 2 problems (maybe I'm missing something)
> 1 no way to specify cdf output option using a config file on disk (only through ui which will not work with tests)
> 2 no way to indicate in tests that cdf output file should be compared (currently test always compares to summary json)

1) I added a config option here: https://github.com/BrightSpots/rcv/pull/246/commits/1b361fb1bb12c9a3d6d06a5006a733ffbf82bc65

2) I'll just make it so that if the config has a CDF CVR source, the test will try to compare both the CDF output and the summary output.

src/main/java/com/rcv/TabulatorSession.java
tarheel: > validate is done for you by loadContestConfig

That method validates the RawContestConfig, but not the ContestConfig (which is a different form of validation).

tarheel:
Regarding candidates: I changed my mind. Your approach (have an empty array in the config) is easier because that's what the UI will generate. I did have to tweak how we load them from the CDF to fix a bug, though; the issue was that processCandidateData would get called multiple times and thus add duplicate copies of the CDF candidates to the raw config.

I also reverted the BigDecimal string thing, since I'm guessing that's how we'll end up.

I got the test working for test set 0. Once we agree on the details, I'll follow up and fix all of them.

tarheel:
> I'll just make it so that if the config has a CDF CVR source, the test will try to compare both the CDF output and the summary output.

What I should have said: if the config has generateCdfJson enabled, the test should check the CDF output also.

src/test/java/com/rcv/TabulatorTests.java
```diff
@@ -81,7 +81,8 @@ private static boolean fileCompare(String path1, String path2) {
           break;
         }
         // both files have content so compare it
-        if (!line1.equals(line2)) {
+        if (!(line1.contains("GeneratedDate") && line2.contains("GeneratedDate"))
```
tarheel: This is pretty hacky. The alternative would be to have a dummy timestamp that we always use for test data, and alter ResultsWriter so that it knows to populate that dummy timestamp instead of the real timestamp when it's operating in a test environment. That seems less hacky, I guess...?

moldover: Agreed.  I made an issue for it so we can defer.

src/main/java/com/rcv/TabulatorSession.java
moldover: Ah you are right.  I don't even know my own validate code I just worked on :(

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -91,6 +93,11 @@ private static void generateJsonFile(String path, Map<String, Object> json) thro
     ObjectMapper mapper = new ObjectMapper();
     // set mapper to order keys alphabetically for more legible output
     mapper.configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true);
+    // create a module to contain a serializer for BigDecimal serialization
+    SimpleModule module = new SimpleModule();
```
moldover: One can't help but wonder what other Modules are out there...

PR #255, opened 2019-04-10
fix bug with handling empty CVRs

tarheel:
This addresses the problem you're having with the 2017 Minneapolis data.

### DIFF
--- a/src/main/java/com/rcv/Tabulator.java
+++ b/src/main/java/com/rcv/Tabulator.java
@@ -22,6 +22,7 @@
 package com.rcv;
 
 import com.rcv.CastVoteRecord.VoteOutcomeType;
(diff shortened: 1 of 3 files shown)


src/main/java/com/rcv/Tabulator.java
```diff
@@ -818,6 +824,11 @@ private void recordSelectionForCastVoteRecord(
         continue;
       }
 
+      // check for a CVR with no rankings at all
+      if (cvr.rankToCandidateIDs.isEmpty()) {
```
moldover: So all those people "voted" but didn't vote?

tarheel: Yeah. You can imagine this being pretty common when there are multiple contests on the ballot and some of them are more obscure.

PR #256, opened 2019-04-10
generate valid IDs for candidates in CDF JSON

tarheel:
As requested, we now generate a valid JSON ID for each candidate and include the full candidate name as a separate value.

### DIFF
--- a/src/main/java/com/rcv/ResultsWriter.java
+++ b/src/main/java/com/rcv/ResultsWriter.java
@@ -56,7 +56,7 @@ class ResultsWriter {
   private static final String CDF_GPU_ID = "gpu-001";
   private static final String CDF_REPORTING_DEVICE_ID = "rd-001";
 
-  private static final Map<String, String> candidateCodeToJsonId = new HashMap<>();
+  private static final Map<String, String> candidateCodeToCdfId = new HashMap<>();
 
   // number of rounds needed to elect winner(s)
   private int numRounds;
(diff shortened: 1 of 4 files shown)


src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -56,6 +56,8 @@
   private static final String CDF_GPU_ID = "gpu-001";
   private static final String CDF_REPORTING_DEVICE_ID = "rd-001";
 
+  private static final Map<String, String> candidateCodeToJsonId = new HashMap<>();
```
moldover: Would it be clearer to change JsonId => CdfId ?   Here and elsewhere...

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -607,23 +622,29 @@ void generateCdfJson(List<CastVoteRecord> castVoteRecords) throws IOException {
...
-    List<Map<String, String>> contestSelections = new LinkedList<>();
-    for (String candidate : config.getCandidateCodeList()) {
-      contestSelections.add(
-          Map.ofEntries(entry("@id", candidate), entry("@type", "CVR.ContestSelection")));
+    List<Map<String, Object>> contestSelections = new LinkedList<>();
+    for (String candidateCode : config.getCandidateCodeList()) {
+      Map<String, String> codeMap = Map.ofEntries(
+          entry("@type", "CVR.Code"),
+          entry("Type", "other"),
+          entry("OtherType", "vendor-label"),
```
moldover: vendor-label doesn't seem right

src/main/java/com/rcv/ResultsWriter.java
tarheel: Sure.

src/main/java/com/rcv/ResultsWriter.java
tarheel: I'm just following what John suggested in his email.

moldover: In that case, great :)

PR #257, opened 2019-04-10
CDF Output: add GpUnit ids for election level (jurisdiction) and cvr level (precinct)

moldover:
add precinct example unit test to demonstrate this (the Minneapolis election is too huge to use for testing and iteration with CDF in its current bloated format) 
add reporting device application and vendor fields


src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -53,11 +54,17 @@
...
+  private static final String CDF_GPU_ID = "gpu-election";
+  private static final String CDF_GPU_ID_FORMAT = "gpu-%d";
   private static final String CDF_REPORTING_DEVICE_ID = "rd-001";
 
   // number of rounds needed to elect winner(s)
   private int numRounds;
+  // all precinct Ids which may appear in the output cvrs
+  private Set<String> precinctIds;
+    // cache precinct to GpUnitId map 
+  private Map<String, String>GpUnitIds;
```
tarheel: nit: missing a space here

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -490,7 +504,12 @@ void generateOverallSummaryFiles(Map<Integer, Map<String, BigDecimal>> roundTall
     generateSummaryJson(roundTallies, null, outputPath);
   }
 
+  // create NIST Common Data Format CVR json
   void generateCdfJson(List<CastVoteRecord> castVoteRecords) throws IOException {
+
```
tarheel: super nit: if you don't mind, let's be consistent about not having a blank line to start methods

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -543,23 +597,45 @@ void generateCdfJson(List<CastVoteRecord> castVoteRecords) throws IOException {
       for (int round = 1; round <= numRounds; round++) {
         List<Pair<String, BigDecimal>> currentRoundSnapshotData =
             cvr.getCdfSnapshotData().get(round);
+
+        // this is the hack
```
tarheel: Once you rebase on top of #255, you can remove this abomination.

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -543,23 +597,45 @@ void generateCdfJson(List<CastVoteRecord> castVoteRecords) throws IOException {
...
         }
         cvrSnapshots.add(generateCvrSnapshotMap(cvr, round, currentRoundSnapshotData));
         previousRoundSnapshotData = currentRoundSnapshotData;
       }
+      // create new cvr map entry
 
-      cvrMaps.add(
-          Map.ofEntries(
+      // if using precincts add GpUnitId for cvr precinct
+      if (config.isTabulateByPrecinctEnabled()) {
```
tarheel: I'm guessing you did this because `Map.ofEntries` returns an immutable Map. Let's not duplicate all of this logic, though. The cleanest way I've identified to handle a case like this is to forgo `ofEntries`. Instead, just create a normal empty Map, add each of the common entries using `put`, and then use an `if` block for your conditional `put`.

src/main/java/com/rcv/Tabulator.java
```diff
@@ -64,17 +64,21 @@
   private BigDecimal winningThreshold;
   // tracks residual surplus from multi-seat contest vote transfers
   private Map<Integer, BigDecimal> roundToResidualSurplus = new HashMap<>();
+  // precinct Ids which may appear in the cast vote records
+  private Set<String> precinctIds;
```
tarheel: Would it be more accurate to refer to these as precinct names instead of IDs? It's a bit confusing now that we're generating a separate set of IDs for them in `ResultsWriter`.

src/main/java/com/rcv/Tabulator.java
```diff
@@ -64,17 +64,21 @@
...
   // param: castVoteRecords list of all cast vote records to be tabulated for this contest
   // param: config describes various tabulation rules to be used for tabulation
+  // param: precinct Ids which may appear in the cast vote records
   // returns: the new object
-  Tabulator(List<CastVoteRecord> castVoteRecords, ContestConfig config) {
+  Tabulator(List<CastVoteRecord> castVoteRecords, ContestConfig config, Set<String> precinctIds) {
     this.castVoteRecords = castVoteRecords;
     this.candidateIDs = config.getCandidateCodeList();
     this.config = config;
+    this.precinctIds = precinctIds;
```
tarheel: Also, now that you're doing this, you can change `initPrecinctRoundTallies` to just use this set instead of iterating through the whole set of CVRs looking for distinct precincts.

src/main/java/com/rcv/ResultsWriter.java
moldover: if you keep up writing comments in code I am more than happy to oblige :)

src/main/java/com/rcv/Tabulator.java
moldover: Not more accurate, but I agree it would be less confusing.

src/main/java/com/rcv/ResultsWriter.java
tarheel: Solid whataboutism!

src/main/java/com/rcv/Tabulator.java
moldover: dig

Issue #241, opened 2019-03-29
combine special winner determination variants into single config field


tarheel:
From #240:

I'm also thinking -- now that we have a few mutually exclusive variants like this, I wonder if we should combine them into a pulldown. Something like "special winner determination variant," where the options would be:
None
Continue until two candidates remain
Sequential multi-seat
Bottoms-up
(possibly more in the future)

tarheel:
Fixed in #346.

PR #258, opened 2019-04-11
add support for bottoms-up multi-seat

### LINKED ISSUE #240: bottoms-up primary variant [nsfw]
May be used for primaries in Utah. Allow the operator to specify the number of "winners" (e.g. 2X the number of seats they're actually trying to fill) then run a single-seat contest with no winning threshold until you have exactly the right number of continuing candidates, at which point you stop.

Open questions:
- Should we forbid batch elimination in this mode? Or should we allow it, but make it smart enough so that it doesn't mistakenly eliminate too many candidates?
- How should the operator specify that they want to run in this mode? Initial proposal: they set numWinners to the number of "winners," and there's a new checkbox labeled "bottoms-up mode" (or whatever) that can only be enabled when numWinners > 1.

tarheel:
Fixes #240. I'll add a unit test for this also.

Also see #241. I would go ahead and fix that part now, but I want to make sure we agree on how we're handling backward compatibility (or lack thereof).


src/main/java/com/rcv/Tabulator.java
```diff
@@ -397,9 +398,12 @@ private boolean shouldContinueTabulating() {
     } else {
       // If there are more seats to fill, we should keep going, of course.
       // But also: if we've selected all the winners in a multi-seat contest, we should tabulate one
-      // extra round in order to show the effect of redistributing the final surpluses.
+      // extra round in order to show the effect of redistributing the final surpluses... unless
+      // bottoms-up is enabled, in which case we can stop as soon as we've declared the winners.
```
moldover: 👍

tarheel:
Did you finish reviewing this PR?

Also, what does it mean when it says "moldover self-requested a review 6 hours ago"? I've seen that on a few PRs that I've authored recently.

moldover:
Github UI is weird.  I found by clicking on the arrows next to my name in the requested reviewers area, I can get to the "submit review" button.  I bet that's what is generating the self-requested notifications.

PR #259, opened 2019-04-12
more cdf tweaks

moldover:
Tear it up...

### DIFF
--- a/src/main/java/com/rcv/ResultsWriter.java
+++ b/src/main/java/com/rcv/ResultsWriter.java
@@ -612,7 +612,7 @@ class ResultsWriter {
       for (int round = 1; round <= numRounds; round++) {
         List<Pair<String, BigDecimal>> currentRoundSnapshotData =
             cvr.getCdfSnapshotData().get(round);
-        
+
         if (currentRoundSnapshotData == null) {
           if (previousRoundSnapshotData == null) {
             throw new RoundSnapshotDataMissingException(cvr.getID());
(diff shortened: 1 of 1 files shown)


src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -702,10 +702,13 @@ void generateCdfJson(List<CastVoteRecord> castVoteRecords)
...
           entry("Value", config.getNameForCandidateCode(candidateCode))
       );
+      // codeMap is wrapped in an array
+      ArrayList<Map<String, String>> codeArray = new ArrayList<>();
+      codeArray.add(codeMap);
       contestSelections.add(Map.ofEntries(
           entry("@id", getCdfIdForCandidateCode(candidateCode)),
           entry("@type", "CVR.ContestSelection"),
-          entry("Code", codeMap)
+          entry("Code", codeArray)
```
tarheel: this is simpler and matches what I've done elsewhere:

`entry("Code", new Map[]{codeMap}),`

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -681,7 +681,7 @@ void generateCdfJson(List<CastVoteRecord> castVoteRecords)
     Map<String, Object> contestMap =
         Map.ofEntries(
             entry("ContestId", CDF_CONTEST_ID),
-            entry("CVRContestSelection", selectionMapList),
+            entry("ContestSelection", selectionMapList),
```
tarheel: I couldn't tell for sure from John's email whether this was the thing he wanted us to change.

moldover: Yeah, his email (and the standard) are confusing, but I think it's right.  Looking at all the other @type elements it checks out.

src/main/java/com/rcv/ResultsWriter.java
moldover: 👍

PR #260, opened 2019-04-23
fix CDF bugs and update test sets accordingly

tarheel:
More progress on #244. Fixes various issues that we overlooked previously or introduced with recent updates to the CDF spec. I confirmed my fixes by getting test set 0 to work properly again.


src/test/resources/test_data/test_set_0_skipped_first_choice/test_set_0_skipped_first_choice_cvr.json
```diff
@@ -648,32 +673,59 @@
...
+    "Name" : "TestSets_Single_winner",
+    "OtherType" : "Election Scope Jurisdiction",
     "Type" : "other"
   } ],
   "ReportGeneratingDeviceIds" : [ "rd-001" ],
   "ReportingDevice" : [ {
     "@id" : "rd-001",
-    "@type" : "CVR.ReportingDevice"
+    "@type" : "CVR.ReportingDevice",
+    "Application" : "RCV Universal Tabulator",
```
HEdingfield: Just realized I've been calling it the "Universal RCV Tabulator". Recommend changing the instance of this in the code (in the window title).

tarheel:
> I assume it's expected that any tests with stems like "CDF_multi_winner" or "test_set_..." will fail (besides test_set_0)? However, it also looks like "2015 Portland Mayor Candidate Codes" and "precinct example" are both broken now too, so please fix those before pushing.

You're correct that all the other test sets and the CDF_multi_winner are expected to be broken at the moment. Thanks for flagging the others. My plan is to get the full suite working before I commit this PR.

PR #261, opened 2019-04-23
file picker should let you select CDF JSON

tarheel:
bitmoji

### DIFF
--- a/src/main/java/com/rcv/GuiConfigController.java
+++ b/src/main/java/com/rcv/GuiConfigController.java
@@ -318,7 +318,6 @@ public class GuiConfigController implements Initializable {
     String cvrFilePath = textFieldCvrFilePath.getText().trim();
     String cvrFirstVoteCol = textFieldCvrFirstVoteCol.getText().trim();
     String cvrFirstVoteRow = textFieldCvrFirstVoteRow.getText().trim();
-    String cvrProvider = textFieldCvrProvider.getText().trim();
     boolean fileIsJson = cvrFilePath.toLowerCase().endsWith(".json");
     if (cvrFilePath.isEmpty()) {
       Logger.log(Level.WARNING, "CVR file path is required!");
@@ -332,7 +331,7 @@ public class GuiConfigController implements Initializable {
       cvrSource.setFirstVoteRowIndex(getIntValueOrNull(cvrFirstVoteRow));
       cvrSource.setIdColumnIndex(getIntValueOrNull(textFieldCvrIdCol));
       cvrSource.setPrecinctColumnIndex(getIntValueOrNull(textFieldCvrPrecinctCol));
-      cvrSource.setProvider(cvrProvider);
+      cvrSource.setProvider(textFieldCvrProvider.getText().trim());
       tableViewCvrFiles.getItems().add(cvrSource);
       textFieldCvrFilePath.clear();
       textFieldCvrFirstVoteCol.clear();
--- a/src/main/java/com/rcv/GuiConfigController.java
+++ b/src/main/java/com/rcv/GuiConfigController.java
@@ -303,7 +303,8 @@ public class GuiConfigController implements Initializable {
   public void buttonCvrFilePathClicked() {
     FileChooser fc = new FileChooser();
     fc.setInitialDirectory(new File(FileUtils.getUserDirectory()));
-    fc.getExtensionFilters().add(new ExtensionFilter("Excel files", "*.xls", "*.xlsx"));
+    fc.getExtensionFilters()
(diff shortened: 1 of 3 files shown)


src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -317,19 +318,20 @@ public void buttonAddCvrFileClicked() {
     String cvrFilePath = textFieldCvrFilePath.getText().trim();
     String cvrFirstVoteCol = textFieldCvrFirstVoteCol.getText().trim();
     String cvrFirstVoteRow = textFieldCvrFirstVoteRow.getText().trim();
+    String cvrProvider = textFieldCvrProvider.getText().trim();
     if (cvrFilePath.isEmpty()) {
       Logger.log(Level.WARNING, "CVR file path is required!");
-    } else if (cvrFirstVoteCol.isEmpty()) {
+    } else if (cvrFirstVoteCol.isEmpty() && !cvrProvider.equals("CDF")) {
```
HEdingfield: This seems like a really bizarre UX to me to have to count on the user to type "CDF" exactly in the provider box in order to skip needing to enter row and column fields, especially considering provider is an optional field. Why not instead just check if the file type is .json?

src/main/resources/GuiConfigLayout.fxml
```diff
@@ -103,9 +103,9 @@
                     text="Select"/>
                 </HBox>
                 <HBox alignment="CENTER_LEFT" spacing="4.0">
-                  <TextField prefHeight="25.0" prefWidth="160.0" promptText="First Vote Col Index *"
+                  <TextField prefHeight="25.0" prefWidth="160.0" promptText="First Vote Col Index"
```
HEdingfield: Please update the descriptions in the txt file if you're going to change the "required" behavior here.

HEdingfield:
I forgot to add that the precinct test is broken here too, as is test_set_0 (which I assume is expected since the other PR fixes that). However, it's odd to me that test_set_1 and beyond pass here but fail in #260. Probably just because I'm out of the loop...

src/main/java/com/rcv/GuiConfigController.java
tarheel: OK, that makes sense here.

tarheel:
> I forgot to add that the precinct test is broken here too, as is test_set_0 (which I assume is expected since the other PR fixes that). However, it's odd to me that test_set_1 and beyond pass here but fail in #260. Probably just because I'm out of the loop...

Don't worry! I'll make sure everything is working when I'm done with the other PR.

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -317,19 +318,21 @@ public void buttonAddCvrFileClicked() {
     String cvrFilePath = textFieldCvrFilePath.getText().trim();
     String cvrFirstVoteCol = textFieldCvrFirstVoteCol.getText().trim();
     String cvrFirstVoteRow = textFieldCvrFirstVoteRow.getText().trim();
+    String cvrProvider = textFieldCvrProvider.getText().trim();
```
HEdingfield: You can inline this again in line 335.

src/main/resources/config_file_documentation.txt
```diff
@@ -58,12 +58,12 @@ Config file must be valid JSON format. Examples can be found under the test fold
         example: /Users/test_data/2015-portland-mayor-cvr.xlsx
         value: string of length [1..1000]
 
-      "firstVoteColumnIndex" required
+      "firstVoteColumnIndex" required for Excel files
```
HEdingfield: "for Excel files" seems a little vague. Maybe "required unless cast-vote records are in JSON format"? This applies to all instances where "Excel" is used here.

src/main/resources/config_file_documentation.txt
```diff
@@ -87,7 +87,7 @@ Config file must be valid JSON format. Examples can be found under the test fold
         if not supplied: none
 
   "candidates" required
-    list of registered candidate names and associated candidate code
+    list of registered candidate names and associated candidate code (note: leave empty when CVR is in Common Data Format)
```
HEdingfield: Probably get rid of the * in "Name *" in `GuiConfigLayout.fxml` in light of this. Though at this point I'm feeling like these notes are getting pretty watered down. Maybe we should consider a new symbol (**?) for "required unless CVR is JSON"?

Another thing: you may need to modify the config validation to allow for no candidates if JSON.

tarheel: Well, "Name" is still required if you're actually going to add a candidate. We just don't want you to add any if the input is CDF. I'll enforce that in a separate PR.

> Another thing: you may need to modify the config validation to allow for no candidates if JSON.

This part already works.

PR #262, opened 2019-04-26
handle duplicate candidates in CVRs properly in CDF output

tarheel:
OK, this is pretty gnarly, but I think I got it right.

Previously, if a given ballot had the same candidate ranked in multiple spots, we were generating a separate ContestSelection for each of those ranks. This worked fine -- or, at least, our tabulator was able to understand it fine -- but I realized last night that what the spec intends is for us to handle this case by having a single ContestSelection with multiple SelectionPositions (one for each rank). Transforming our CVR data so that we can export in that format was surprisingly tricky. I also added sorting logic so that the ordering of ContestSelections and SelectionPositions will be deterministic (which is also relevant for overvote situations).

And then I updated a bunch of tests accordingly.

### DIFF
--- a/src/main/java/com/rcv/ResultsWriter.java
+++ b/src/main/java/com/rcv/ResultsWriter.java
@@ -37,6 +37,7 @@ import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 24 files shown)


HEdingfield:
I think I should probably defer to @moldover on this one!

tarheel:
Ha, yeah. This is a scary one.

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -643,48 +682,74 @@ void generateCdfJson(List<CastVoteRecord> castVoteRecords)
...
-                entry("HasIndication", "yes"),
-                entry("IsAllocable", isAllocable),
-                entry("NumberVotes", numberVotes),
-                entry("Rank", rank),
-                entry("@type", "CVR.SelectionPosition"));
+        // didn't find an allocation, i.e. this ballot didn't contribute all or part of a vote to
+        // this candidate in this round
+        if (isAllocable.equals("unknown")) {
+          isAllocable = "no";
+          // not sure what numberVotes should be in this situation
```
moldover: I'd just remove this

src/main/java/com/rcv/ResultsWriter.java
```diff
@@ -643,48 +682,74 @@ void generateCdfJson(List<CastVoteRecord> castVoteRecords)
...
 
-        selectionMapList.add(
-            Map.ofEntries(
-                entry("ContestSelectionId", getCdfIdForCandidateCode(candidateCode)),
-                entry("SelectionPosition", new Map[]{selectionPositionMap}),
-                entry("@type", "CVR.CVRContestSelection")));
+      String fractionalVotes = null;
+      if (!numberVotes.equals(BigDecimal.ONE)) {
+        BigDecimal remainder = numberVotes.remainder(BigDecimal.ONE);
+        if (remainder.signum() == 1) {
```
moldover: Hello signum my old friend!

tarheel: Because a division softly creeping
Left its remainder while I was tabulating

src/main/java/com/rcv/ResultsWriter.java
tarheel: Remove the comment, you mean? Or the field? The current behavior feels kind of weird.

moldover: The current comment sounds like _maybe_ there is some remaining task here?  A bit confusing.
My take is, if it's not allocable, numberVotes is meaningless and it doesn't matter what happens to it.  I'd just remove it, or update it to explain why we have the current behavior.  It's a small thing, also feel free to leave as is.

tarheel: Your take is correct. As you said, it's meaningless. I had only included it because I thought it was required. I agree it would be better to remove it when isAllocable is false.

I'll leave it out of this PR because I want to get confirmation from John first... and once I make this change, I'll have to go back and update all the test output files yet again. :/

Follow-up in #266.

PR #263, opened 2019-04-27
add multi-winner tests in CDF format

tarheel:
Updated our multi-winner test cases to match George's two examples. Also fixed a minor bug in how we compute the inactive ballot count.


tarheel:
Oops. Good catch.

PR #264, opened 2019-05-02
Fix deprecated  API call

### LINKED ISSUE #222: Replace deprecated XMLReaderFactory.createXMLReader()
According to [this](https://stackoverflow.com/questions/51072419/how-use-xmlreaderfactory-now-because-this-is-deprecated), it should be relatively straightforward to replace this line:

`XMLReader xmlReader = XMLReaderFactory.createXMLReader();`

with this one:

`XMLReader xmlReader = SAXParserFactory.newInstance().newSAXParser().getXMLReader();`

However, the `parseCVRFile` function returns an empty `cvrList` if this is done. I tried playing with it for long time but couldn't figure it out...

HEdingfield:
Fixes #222.


tarheel:
bitmoji

Issue #270, opened 2019-05-15
Update to Java 12 and fix link in README to Oracle Java SE


HEdingfield:
We should be linking to and using OpenJDK, per [this article](https://medium.com/@javachampions/java-is-still-free-2-0-0-6b9aa8d6d244).

#239 is likely a blocker for this.

The most straightforward thing to link and use would probably be Oracle's [open source builds of the JDK here](https://jdk.java.net/).

HEdingfield:
Counterpoint: we may want to stick with Java 11, since it's LTS until 2022 per [the table here](https://en.wikipedia.org/wiki/Java_version_history). In either case, we're still updating to a new version of Java.

tarheel:
Any particular downside to going straight to 12?

HEdingfield:
See my previous comment. Means we'd need to stay on top of new versions every 6 months, whereas 11 is LTS until 2022. It's probably still worth going to 12 and beyond, but we need to have a discussion about the certification process, who's responsible for building the code, if we need to stick with images frozen in time, etc.

HEdingfield:
Addressed via #272. Now the tabulator _only_ works in JDK 11+. When distributed using Gradle's `distZip` task, it requires that the user have the JDK installed. When distributed using Gradle's `jlink` task, it doesn't even require that the user have the JDK installed (the zip is 48 MB vs. 26 MB with `distZip`).

Issue #168, opened 2018-09-16
Unable to run on Linux


HEdingfield:
When attempting to debug #160, I was unable to get the tabulator running on my Linux machine.

Couldn't verify because:
`$ /usr/local/buildtools/java/jdk10-64/bin$ ./java -jar ~/Downloads/RCV_UniversalTabulator_20180908/RCV_UniversalTabulator_20180908/rcv.jar `

Would cause:

> Error: Could not find or load main class com.rcv.Main
> Caused by: java.lang.NoClassDefFoundError: javafx/application/Application

I initially assumed it was because there's no MANIFEST.MF file in the version @moldover compiled, but I also tried the version I compiled which included a manifest and had the same problem.

nealmcb:
I still see this problem, on Ubuntu 18.04 LTS Bionic:

```
openjdk 10.0.2 2018-07-17
OpenJDK Runtime Environment (build 10.0.2+13-Ubuntu-1ubuntu0.18.04.4)
OpenJDK 64-Bit Server VM (build 10.0.2+13-Ubuntu-1ubuntu0.18.04.4, mixed mode)

```
 And I don't have another system to run it on.
Any luck figuring out what's going on?

tarheel:
Sorry, we haven't had a chance to debug this yet. :/ We'll try to get to it in the not-too-distant future...

tarheel:
@HEdingfield or @nealmcb: could you try the approach I suggested in #239 and see if that fixes it for you?

Specifically: download the [JavaFX SDK for Linux](https://gluonhq.com/products/javafx/), then add command-line args to reference it. Something like this:

`java --module-path javafx-sdk-11.0.2/lib --add-modules=javafx.controls,javafx.fxml -jar rcv.jar`

I'm not sure if it will make a difference if you're running JRE 11 vs. 10.

I tried testing it myself on a remote Linux machine, but I got a different JavaFX error because I'm just sshing in ("java.lang.UnsupportedOperationException: Unable to open DISPLAY"), and I don't want to deal with setting up VNC or something similar right now.

tarheel:
This rigmarole makes me wonder whether we should start releasing a separate build that's intended only for command-line operation and doesn't have any JavaFX dependencies.

nealmcb:
Thank for follwoing up, @tarheel!

I did this on Ubuntu bionic 18.04, where I have OpenJDK 11:

    $ apt-get install openjfx
[...]

Issue #110, opened 2018-08-06
Use standardized project structure


HEdingfield:
I've been having trouble getting the Help button I'm working on to call up the config documentation txt file. It would help out a lot if we moved to a more standard directory structure like that used by [Apache's Maven project](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html). [Here's](https://github.com/TheItachiUchiha/MediaPlayerFX) a simple example of the layout in action.

@tarheel @moldover sound reasonable?

HEdingfield:
I was researching [Maven](https://maven.apache.org/what-is-maven.html) more, and it seems like migrating to this structure will allow us to build out tests, manage dependencies, and deploy more efficiently. It also integrates well with IntelliJ. Check out the quick video [here](https://www.jetbrains.com/help/idea/maven-support.html).

CalebKleppner:
This is above my pay grade, but if the 3 of you think this makes sense, I am fully supportive.

HEdingfield:
If we convert to Maven, we can probably delete the libs directory.

HEdingfield:
#239 is related, since we discuss migrating everything to Gradle.

tarheel:
Converting to Maven/Gradle to own the libs.

HEdingfield:
Gradle migration complete with #272.

PR #272, opened 2019-05-19
Gradle migration

### LINKED ISSUE #239: tabulator doesn't run in JDK 11
From Greg Dennis:

Hi guys,

The suggestion on rankedchoicevoting.org to use JDK 11 doesn't seem to work:

> java -jar rcv_20190302.jar
Error: Could not find or load main class com.rcv.Main
Caused by: java.lang.NoClassDefFoundError: javafx/application/Application

I think this is because JavaFX is available as a separate module in JDK 11. I had to jump through Oracle's hoops to download their archived JDK 10 version to run it successfully.

There is some guidance on StackOverflow on building a JavaFX app that runs on JDK 11 (and presumably continues to work with JDK 10, too). Have you looked into doing that? If that would take a longer-term effort, maybe the website instructions should be updated in the interim?

Thanks,
Greg

HEdingfield:
Once completed, this fixes or makes progress on #239, #168, #110, and #270.

I need help on this! I've spent many hours on it and pushed it as far as I can for now. You'll probably need to download JDK 12 [here](https://jdk.java.net/12/) and point the IntelliJ to that. Then use the Gradle menu along the right side of the screen, then Tasks => build => build.

The problem I'm running into is one of transitive dependencies, e.g. 100 errors like this:

> error: module com.fasterxml.jackson.databind reads package org.apache.poi from both poi.ooxml and poi

I think it's a matter of tweaking things in the `gradle.build` and `module-info.java` files to get this right. Here are the most useful posts I've found on the subject so far:

- https://discuss.gradle.org/t/how-do-i-exclude-specific-transitive-dependencies-of-something-i-depend-on/17991/9
- https://blog.codefx.org/java/java-9-migration-guide/#Split-Packages
- https://stackoverflow.com/questions/42275516/excluding-tomcat-dependencies-from-spring-boot-in-gradle


moldover:
This is cool!  Did as you suggested (got JDK 12) and gradle downloaded all the packages and I'm able to re-pro the same compilation errors.  Will dig in more!

HEdingfield:
Seven blessings to you 🙏

tarheel:
I'll throw in an eighth blessing if you figure out the errors.

moldover:
@HEdingfield I pushed a couple commits which get the project building and creating .zip distribution.  Unzipping and running the launcher script on my computer does launch the app, but it immediately shuts down:

2019-05-26 15:27:00 PDT INFO: RCV Tabulator logging execution to: /Users/Jon/Documents/rcv/build/distributions/rcv/bin/rcv_%g.log
2019-05-26 15:27:00 PDT INFO: Tabulator is being used via the CLI.
2019-05-26 15:27:00 PDT WARNING: Too many arguments! Max is 2 but got: 3
2019-05-26 15:27:00 PDT SEVERE: Error opening file: -classpath
java.io.FileNotFoundException: -classpath (No such file or directory)
2019-05-26 15:27:00 PDT SEVERE: Check file path and permissions and make sure they are correct!
2019-05-26 15:27:00 PDT SEVERE: Failed to load contest config: -classpath
2019-05-26 15:27:00 PDT SEVERE: Aborting because contest config is invalid!

Seem to be making some progress :)  

However I'm not convinced this .zip file contains the .jdk files which jlink purports to include, since it seems to only contain 3rd-party jar files and the rcv jar file--i don't see any platform libs.

Other side notes:
- gradle build fails near the end, when it tries to execute a test task.  Seems like it's not set up to run against our TabulatorTests yet, but this doesn't impact creating the distribution .zip file.
- running "build" and run/debugging test configurations from intelliJ doe not work

moldover:
I think passing classpath is what jlink does for "default JVM arguments" and can be disabled in the jlink piece of gradle build file.

HEdingfield:
Nice! I'm without a computer until 6/3, but I'll take a look soon after I get back if this still isn't resolved by then.

tarheel:
You haven't downloaded the IntelliJ iPhone app yet?

HEdingfield:
iPhone more like cryPhone cause it means you paid twice as much for inferior technology am I rite guys??

moldover:
A bit more info on this as I learn more (so many opportunities for growth lol)  The poi module compilation errors are actually due to poi being a "split" package.  I did get it to build yes, but we are not including the org.apache.poi:poi package (several other packages were removed as well?) so it will fail at runtime.

I think it is actually going to take some fancy jar repackaging in order to make this work correctly.

See https://poi.apache.org/help/faq.html#faq-N102B0  they are "working" on automatic module support for poi.

More chatter for context: http://apache-poi.1045710.n5.nabble.com/Bug-62355-New-Unsplit-packages-to-conform-Jigsaw-Java-9-module-standard-td5730659.html

tarheel:
I haven't spent any time looking at this, but I have a quick thought: any chance that it would be easier to use something other than poi instead of jumping through hoops to get poi to compile?

moldover:
There are other xlsx [options](http://roide-textcube.blogspot.com/2009/02/java-libraries-to-read-and-write-excel.html
): but I think this would take much more time than tweaking our build.  

Merging a split [package](https://msayag.github.io/SplitPackage/) doesn't seem that hard (it's essentially merging two directory trees and then repackaging as a jar) but after still more reading I think the better solution is just to not modularize our application yet - see this [article](https://blog.joda.org/2017/05/java-se-9-jpms-automatic-modules.html) which seems to be exactly our case.  (you will need this [background](http://tutorials.jenkov.com/java/modules.html) to understand what he is talking about)

TL;DR: one should probably wait till ALL one's dependencies have modular releases before modularizing your overall app. 

However, I think/hope we should be able to do everything we want (use gradle, create platform-specific builds on JDK 12) and still release a self-contained (but non-modular) app.  

@HEdingfield do you know how gradle build steps are defined?  For example, it creates a .tar file, and tries to run tests which are not currently configured.  We will want to tailor these.

moldover:
@HEdingfield & @tarheel I think I have finally fixed this issue.  Turns out apache poi in March released version 4.0.1 which addresses this split packages [issue](https://bz.apache.org/bugzilla/show_bug.cgi?id=62355).  I just pulled the newer version, added back the other required packages and it now complies and I'm able to run and tabulate an election.  There are some things to be fixed:
gradle: There were failing tests. See the report at: file:///Users/Jon/Documents/rcv/build/reports/tests/test/index.html
On app launch:  2019-05-29 23:01:14 PDT WARNING: Resource "GuiStyles.css" not found.

But mainly I think the way we are invoking jlink is not actually adding any java framework classes, so I suspect it won't run if you have java < 10.  What are you guys on?  Hylton, do you know anything about how jlink knows what files to add?

I attached the .zip our gradle build creates (there is also a tar) you can try launching the app by double-clicking bin/rcv 

[rcv.zip](https://github.com/BrightSpots/rcv/files/3235622/rcv.zip)

moldover:
@HEdingfield  @tarheel I have continued to chip away at this and I've got it to a point where it seems to be working.  It builds, debugs, passes regression tests, and my local run of the distribution works.

There are some smaller things which can be improved, but I think the main task is complete and should be reviewed and tested.  I may need some help on any needed changes / additions as I am really burnt out on this.  Thank you.

HEdingfield:
Thanks for all the work, @moldover! It's a big step ahead. I spent a few hours on this and basically got it working, but recorded a bunch of outstanding issues. I _think_ we can consider merging it now (and I'll just file them as issues), with the following caveats:

* I could only get it to run using JDK 11+ now (I tried with both 11 and 12 and they worked, but 10 didn't).
* The new distribution (i.e. the one created using `Tasks > distribution > distZip` in IntelliJ) is platform-specific! The one I generate works in Windows, but the one Jon generated and attached above does NOT work for me, probably because it uses *mac JavaFX libraries (whereas mine includes *win versions).
* It's also currently not possible to use the CLI with the new distribution method.
* I think we could make a truly cross-platform version that doesn't even require Java to be installed if we could get `Tasks > build > jlink` to work, but it's currently failing for me with the following:

> Task :createMergedModule
> Cannot derive uses clause from service loader invocation in: com/fasterxml/jackson/databind/ObjectMapper$2.run().
> Cannot derive uses clause from service loader invocation in: com/fasterxml/jackson/databind/ObjectMapper.secureGetServiceLoader().
> Cannot derive uses clause from service loader invocation in: org/apache/commons/compress/utils/ServiceLoaderIterator.<init>().
> C:\Users\Hylton\IdeaProjects\rcv\build\jlinkbase\tmpjars\rcv.merged.module\module-info.java:392: error: the service implementation does not have a default constructor: XBeansXPath
>     provides org.apache.xmlbeans.impl.store.PathDelegate.SelectPathInterface with org.apache.xmlbeans.impl.xpath.saxon.XBeansXPath;
> C:\Users\Hylton\IdeaProjects\rcv\build\jlinkbase\tmpjars\rcv.merged.module\module-info.java:393: error: the service implementation does not have a default constructor: XBeansXQuery
>     provides org.apache.xmlbeans.impl.store.QueryDelegate.QueryInterface with org.apache.xmlbeans.impl.xquery.saxon.XBeansXQuery;
> 2 errors

@tarheel I leave it up to you to make the final call on whether or not it's ok to merge in like this. Or, even better, to take up the torch and try to resolve these jlink errors.

gngilbert:
I think I would be worried if I knew what you guys were talking about.
Since I don't, however, sounds to me like you've got it all under control!
🤗

HEdingfield:
Ok, I spent 9 more hours on this shit and I feel like it's good to push out now. There are still a few major things I'll want to fix ASAP, but at least I got the tests and the CLI working, and I confirmed that it's possible to get a Linux-friendly distribution generated.

Major outstanding issues (which I'll post after this is merged in):
* Building distributions using distZip is platform-specific due to JavaFX graphics libraries getting included (separate for Windows, Mac, Linux); you currently have to build the dist on an actual system using that OS, though hopefully we can figure out a way around that
* jlink fails due to some annoying errors; if we could get this working, we could probably make a single version that works on all platforms and doesn't even require Java to be installed!
* distZip doesn't include any of the test files, which we may want to include as part of the distribution
* I tried updating beyond Gradle 4.10.3 (i.e. a number of versions all the way up to 5.4.1), and test2013MinneapolisMayorScale() would crap out around 500k votes with the message "TestEngine with ID 'junit-jupiter' failed to execute tests"; sticking with 4.10.3 (the IntelliJ default) for now...
* There are a couple of warnings during the build about unchecked exceptions
* I also discovered a minor weird CLI bug when running with Gradle rcv.bat (and possibly otherwise, and possibly only in Windows): directories specified for CVRs (and output) inject \null\ before... workaround requires using a path like "..\\2013_minneapolis_mayor_cvr.xlsx" instead of "2013_minneapolis_mayor_cvr.xlsx"

@tarheel: I'll merge after you approve.

moldover:
Nice work.  I agree with your assessments and have been trying to find info on this jlink error and have so far failed.  It's unclear what the message really means as there is no requirement for service implementations to have default ctors so I don't have a theory on it.  Maybe just ask them.

HEdingfield:
Cool, thanks for working on this too. Yeah, I don't really have any leads on the jlink issue; I do on the others, and I'll file issues with that info as soon as @tarheel approves this PR.

In the meantime, I just posted [this question](https://stackoverflow.com/questions/56512064/gradle-jlink-plugin-fails-during-createmergedmodule-with-the-service-implementa) on SO.

moldover:
Well played @HEdingfield.  Your SO question got such a quick response and it works!  At least, the entire build + tests completed for me with no errors so I pushed the fix, although I don't entirely understand why.

HEdingfield:
Hell yeah! That Spaniard is a champion!

Confirmed that the jlink image works without needing Java installed (the zip is 48 MB vs. 26 MB with distZip), but unfortunately it looks like it's still packaging Windows-specific libraries. Still, a nice step up and a good option to have.

Once this is submitted, I'll file an issue to look into using jpackage and/or special arguments for jlink or distZip to be able to do builds for all 3 OSes on one system.

 @tarheel: here are the things you'll want to verify on your end for your review:

### Using the IntelliJ Gradle panel

* Run:
Tasks => application => run

* Unit tests:
Tasks => verification => test

* Distribute:
Test both after extracting and navigating to the `bin` directory:
** Tasks => distribution => distZip (outputs to `build/distributions/rcv.zip`; requires that Java 11+ SDK is already installed on your system)
** Tasks => build => jlinkZip (outputs to `build/image.zip`; doesn't require Java to be installed)

src/test/java/com/rcv/TabulatorTests.java
```diff
@@ -41,7 +41,7 @@
 class TabulatorTests {
 
   // folder where we store test inputs
-  private static final String TEST_ASSET_FOLDER = "src/test/resources/test_data";
+  private static final String TEST_ASSET_FOLDER = "src/test/resources/com/rcv/test_data";
```
tarheel: Extrapolating from our trajectory to date, I predict that our test files are 60 levels deep by the time we officially release this app.

HEdingfield:
We did it!!

gngilbert:
Congratulations and great work!! All you have to do now is explain to me
what it was you accomplished.?

HEdingfield:
It basically fixes all 4 of the issues mentioned at the very top. I'll update them and add my last outstanding issues shortly.

gngilbert:
I am getting a new laptop tomorrow so you guys will have to tell me how to
get our latest work loaded and running.

Issue #278, opened 2019-06-09
CLI bug with paths in current directory in config


tarheel:
If you run from the command line and specify a config file with no path (i.e. in the current dir), and that config files also lists a CVR source with no path, we end using the string "null" as the directory.

HEdingfield:
Confirmed and noted this in #272: 

> I also discovered a minor weird CLI bug when running with Gradle rcv.bat (and possibly otherwise, and possibly only in Windows): directories specified for CVRs (and output) inject \null\ before... workaround requires using a path like "..\\2013_minneapolis_mayor_cvr.xlsx" instead of "2013_minneapolis_mayor_cvr.xlsx"

moldover:
Fixed with #287

Issue #271, opened 2019-05-18
Add recommendations to some error messages


HEdingfield:
E.g. "Failed to start system logging!" error message should include recommendation(s) on how to fix, e.g.  "check to make sure you have write access in the current working directory".

Audit existing error messages and see if any others could benefit from this.

Loosely related to #141 and #267.

gngilbert:
This sounds like one we need to do.

HEdingfield:
Tagging as essential, but may just make a narrow-scoped first pass on it, and do a broader, more descriptive pass after certification.

moldover:
Fixed with #287

Issue #141, opened 2018-08-29
Improve error messaging for IOExceptions when loading FXML files


HEdingfield:
`Logger.executionLog(Level.SEVERE, "Failed to open: %s:\n%s", resourcePath, exception.getCause().toString());`

The above doesn't show a proper stacktrace; figure out how to improve upon this. Note: it can be triggered for testing by making an error happen in the config creator `loadConfig` function.

gngilbert:
Sounds like this needs to be completed.

moldover:
@HEdingfield trying to repro/trigger this as part of #271 but I think it may be obsolete.  The only place the error message cited above appears in code is in TieBreak.java when it actually fails to load the FXML file, which presumably would only happen if there was some build/packing issue, and not something we need to worry about better user messaging on.  I'm inclined to close this, but maybe I'm just missing the use-case.  Thanks.

HEdingfield:
I specifically mentioned a lack of stacktrace when this type of error occurs, so do you mean it's now showing a stacktrace when it's triggered (e.g. in the `TieBreak.java` example)? If so, can probably close, otherwise should probably keep it open.

moldover:
Gotcha.  So when I trigger an IOException (by inserting a syntax error into GuiTiebreakLayout.fxml) I get a pretty reasonable/helpful message:

2019-06-15 10:37:00 PDT SEVERE: Failed to open: /com/rcv/GuiTiebreakerLayout.fxml:
javax.xml.stream.XMLStreamException: ParseError at [row,col]:[20,27]
Message: Attribute name "mnemonicP" associated with an element type "Button" must be followed by the ' = ' character.

The lack of stacktrace is expected since getCause() does not include it, but sounds like this was a problem at the time, so I added it.  Did the same at the other callsite.  This catches both can't find resource, and problems parsing / loading resource so we should be in good shape.

HEdingfield:
Obsolete / resolved with #287.

PR #287, opened 2019-06-16
Cert bug fixes

moldover:
This contains fixes for #278 #141 and #271 
I generally tried to improve the log messaging and surface issues and status in a clearer and more consistent way.


src/main/java/com/rcv/Logger.java
```diff
@@ -82,20 +82,27 @@ static void setup() throws IOException {
...
-    logger.addHandler(executionHandler);
+    // executionHandler writes to the execution log file in current working directory
+    try {
+      FileHandler executionHandler =
+          new FileHandler(
+              logPath.toString(), LOG_FILE_MAX_SIZE_BYTES, EXECUTION_LOG_FILE_COUNT, true);
+      executionHandler.setLevel(Level.INFO);
+      logger.addHandler(executionHandler);
+    } catch (IOException exception) {
+      log(Level.WARNING, (String.format("Failed to start system logging!\n" +
```
tarheel: you have an extra set of parens around the call to `String.format`

src/main/java/com/rcv/FileUtils.java
```diff
@@ -45,6 +46,8 @@ static void createOutputDirectory(String dir) throws UnableToCreateDirectoryExce
       // dirFile is the File object for dir
       File dirFile = new File(dir);
       if (!dirFile.exists() && !dirFile.mkdirs()) {
+        Logger.log(Level.SEVERE, "Failed to create output directory: %s", dir);
+        Logger.log(Level.SEVERE, "Check the directory name and permissions.");
         throw new UnableToCreateDirectoryException("Unable to create output directory: " + dir);
```
tarheel: The logging you added here isn't redundant with this exception we're throwing? It looks like we catch this exception and then log the same things in TabulatorSession.

src/main/java/com/rcv/Main.java
```diff
@@ -31,6 +31,7 @@
   // param: args command line argument array
   // returns: N/A
   public static void main(String[] args) {
+    System.out.println("RCVRC Universal Tabulator v 1.0");
```
tarheel: let's put the app name and version in constants

src/main/java/com/rcv/GuiApplication.java
```diff
@@ -28,10 +32,18 @@
...
-    window.setScene(new Scene(root));
+  public void start(Stage window) {
+    String resourcePath = "/com/rcv/GuiConfigLayout.fxml";
+    try {
+      Parent root = FXMLLoader.load(getClass().getResource(resourcePath));
+      window.setTitle("RCVRC Tabulator");
+      window.setScene(new Scene(root));
+    } catch (IOException exception) {
+      StringWriter sw = new StringWriter();
+      PrintWriter pw = new PrintWriter(sw);
```
HEdingfield: What's up with these fancy new `StringWriter` and `PrintWriter` libraries, and why are we only using them in the two FXML cases?

src/main/java/com/rcv/Logger.java
moldover: This is removed in the latest commit.

src/main/java/com/rcv/FileUtils.java
moldover: Fixed.

src/main/java/com/rcv/TieBreak.java
```diff
@@ -303,20 +310,20 @@ public String call() {
       window.initModality(Modality.APPLICATION_MODAL);
       window.setTitle("RCV Tiebreaker");
       String resourcePath = "/com/rcv/GuiTiebreakerLayout.fxml";
-      FXMLLoader loader = new FXMLLoader(getClass().getResource(resourcePath));
-
       try {
+        FXMLLoader loader = new FXMLLoader(getClass().getResource(resourcePath));
```
HEdingfield: I need to verify below once you've reconciled this commit with the master branch. Right now it's in a fucked half-state between network.brightspots.rcv and com.rcv and all the test files have vanished.

OLD MESSAGE: When I repro as you suggest in #141 by adding some bad characters at the top of `GuiTiebreakerLayout.fxml`, I get an infinite loop with messages like this:

> 2019-06-16 22:11:13 PDT SEVERE: Failed to open: /network/brightspots/rcv/GuiTiebreakerLayout.fxml:
> javax.xml.stream.XMLStreamException: ParseError at [row,col]:[2,1]
> Message: Content is not allowed in prolog.
> 2019-06-16 22:11:13 PDT WARNING: Invalid selection! Please try again.

So I wouldn't call this issue fixed yet. :(

src/main/java/com/rcv/FileUtils.java
```diff
@@ -45,6 +46,8 @@ static void createOutputDirectory(String dir) throws UnableToCreateDirectoryExce
       // dirFile is the File object for dir
       File dirFile = new File(dir);
       if (!dirFile.exists() && !dirFile.mkdirs()) {
+        Logger.log(Level.SEVERE, "Failed to create output directory: %s", dir);
+        Logger.log(Level.SEVERE, "Check the directory name and permissions.");
```
HEdingfield: Combine with line breaks here and any other cases where we have multiple lines spitting out like this.

src/main/java/com/rcv/Main.java
HEdingfield: Agree, and please set version to 1.0.0.

src/main/java/com/rcv/GuiConfigController.java
```diff
@@ -177,7 +177,7 @@
 
   public void buttonNewConfigClicked() {
     if (checkForSaveAndContinue()) {
-      Logger.log(Level.INFO, "Creating new contest config...");
+      Logger.log(Level.INFO, "Creating new contest config.");
```
HEdingfield: . implies action that completed, so keep as ... Please apply to / revert other cases.

src/main/java/com/rcv/StreamingCVRReader.java
```diff
@@ -198,7 +198,7 @@ private void endCVR() {
     }
     // provide some user feedback on the CVR count
     if (cvrList.size() % 50000 == 0) {
-      Logger.log(Level.INFO, String.format("Parsed %d cast vote records.", cvrList.size()));
+      Logger.log(Level.INFO, String.format("Parsed %d cast vote records...", cvrList.size()));
```
HEdingfield: ... implies an action that's about to happen, so keep as period. Please apply to / revert other cases.

src/main/java/com/rcv/Tabulator.java
```diff
@@ -124,7 +124,7 @@
   // returns: set containing winner(s)
   Set<String> tabulate() {
     logSummaryInfo();
-    Logger.log(Level.INFO, "Starting tabulation...");
+    Logger.log(Level.INFO, "Starting tabulation for %s", this.config.getContestName());
```
HEdingfield:
```suggestion
    Logger.log(Level.INFO, "Starting tabulation for contest '%s'...", this.config.getContestName());
```

src/main/java/com/rcv/Tabulator.java
```diff
@@ -234,7 +234,7 @@
       }
     }
 
-    Logger.log(Level.INFO, "Tabulation completed.");
+    Logger.log(Level.INFO, "Completed tabulation for %s", this.config.getContestName());
```
HEdingfield:
```suggestion
    Logger.log(Level.INFO, "Completed tabulation for contest '%s'...", this.config.getContestName());
```

src/main/java/com/rcv/ContestConfig.java
```diff
@@ -128,10 +128,17 @@ static ContestConfig loadContestConfig(String configPath) {
...
+        // source folder will be the parent of configPath
+        String parentFolder = new File(configPath).getParent();
+        // if there is no parent folder use current working directory
+        if (parentFolder == null) {
+          parentFolder = System.getProperty("user.dir");
+        }
+        config = loadContestConfig(rawConfig, parentFolder);
       } else {
-        Logger.log(Level.SEVERE, "Failed to create contest config!");
+        Logger.log(Level.SEVERE, "Failed to create raw contest config!");
```
HEdingfield: Please combine all log messages like this into one and use line breaks /n when necessary. Some log parsers will count this as 3 severe errors, for example. There are some other cases I see with 2-3 messages in succession elsewhere, so please combine those too.

src/main/java/com/rcv/Main.java
```diff
@@ -32,11 +31,9 @@
   // param: args command line argument array
   // returns: N/A
   public static void main(String[] args) {
-    try {
-      Logger.setup();
-    } catch (IOException exception) {
-      System.err.print(String.format("Failed to start system logging!\n%s", exception.toString()));
-    }
+    System.out.println("RCVRC Universal Tabulator version 1.0");
```
HEdingfield: This print statement is unnecessary and makes it say the same thing twice since it says it again after the log is loaded. Please delete.

src/main/java/com/rcv/Main.java
```diff
@@ -32,11 +31,9 @@
...
+    Logger.setup();
+    Logger.log(Level.INFO, "RCVRC Universal Tabulator version 1.0");
```
HEdingfield: 1.0.0, and please set as constant as Louis suggested, and reference here.

src/main/java/com/rcv/Main.java
```diff
@@ -20,23 +20,23 @@
...
 
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.logging.Level;
 
 public class Main extends GuiApplication {
 
+  public static String APP_NAME = "RCVRC Universal Tabulator";
+  public static String APP_VERSION = "1.0";
```
HEdingfield: 1.0.0 is the correct semantic version.

src/main/java/com/rcv/Main.java
moldover: fixed

src/main/java/com/rcv/GuiApplication.java
moldover: See my comment in #141.  We use them here because you requested them, and these particular errors should only occur during development, in which case a stack trace is reasonable to log.  In other cases where the exception is expected and handled, or never happens, there is no reason to print a stack trace.

src/main/java/com/rcv/TieBreak.java
moldover: That is correct behavior.  It's a development-only error, and it provides the correct message to address the problem.  There's no reason to waste time fixing the loop.

src/main/java/com/rcv/FileUtils.java
moldover: I actually like this style better.  It's more readable.  Is there some advantage to doing this?

src/main/java/com/rcv/ContestConfig.java
moldover: that's a good point, will do

src/main/java/com/rcv/Main.java
moldover: yep, got it in next commit.

src/main/java/com/rcv/Main.java
moldover: So the thinking here is, the System.out message will be the first thing displayed to users launching from the CLI, before logging is setup.  If there is a logging setup issue at this point, the console output makes more sense because of that line.  The repeated message is added to get the same information into the log file, in the interest of improved forensics / reproducibility.

src/main/java/com/rcv/GuiConfigController.java
moldover: So I put this back since you like it, but I respectfully disagree with your assessment.  The . implies only the end of a sentence.  In this case the word "Creating" says explicitly, I am currently, or about to do some creating.  Is there some style guide you are drawing from?

src/main/java/com/rcv/GuiApplication.java
HEdingfield: Yeah, got it, I was just wondering why it required bringing those new libraries in, rather than just using the standard exception stacktrace functionality. No big deal though, I'm sure you tested it and the normal method wasn't working for some reason.

src/main/java/com/rcv/FileUtils.java
HEdingfield: Yeah, sorry comments got messed up in GH; this was regarding certain log parsers that count number of errors. Looks like it made sense to you below so you fixed, thanks.

src/main/java/com/rcv/GuiConfigController.java
HEdingfield: Appreciate it. I spent a good amount of time a while back checking all the log messages and making these changes to bring them in line with standard CLI messaging. I think the logic behind this is because it helps the user tell when something is frozen / still processing (e.g. if the last line was a ...) vs. safe to stop / kill (e.g. if the last line was a .). It also helps a lot with bug reports. So the idea was to creates pairs of "Doing this..." "This complete." statements so there's a clear delineation between what's happening and what's already finished. Hopefully that makes sense.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -166,7 +173,7 @@ RawContestConfig getRawConfig() {
   // purpose: validate the correctness of the config data
   // returns any detected problems
   boolean validate() {
-    Logger.log(Level.INFO, "Validating contest config...");
+    Logger.log(Level.INFO, "Validating contest config.");
```
```suggestion
    Logger.log(Level.INFO, "Validating contest config...");
```

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -443,7 +443,7 @@ public void changeCandidateCode(CellEditEvent cellEditEvent) {
   @Override
   public void initialize(URL location, ResourceBundle resources) {
     Logger.addGuiLogging(this.textAreaStatus);
-    Logger.log(Level.INFO, "Opening tabulator GUI...");
+    Logger.log(Level.INFO, String.format("Welcome to the %s!", Main.APP_NAME));
```
HEdingfield: Fine to keep this line, but I'd still like to keep mention of opening the GUI here. That way, when we're parsing the logs, we can tell if they used GUI vs. CLI, e.g.:

`Logger.log(Level.INFO, "Opening tabulator GUI...");`

src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -22,24 +22,24 @@
...
   // purpose: main entry point to the rcv tabulator program
   // param: args command line argument array
   // returns: N/A
   public static void main(String[] args) {
-    try {
-      Logger.setup();
-    } catch (IOException exception) {
-      System.err.print(String.format("Failed to start system logging!\n%s", exception.toString()));
-    }
+    System.out.println(String.format("Launching %s version %s", APP_NAME, APP_VERSION));
```
HEdingfield:
```suggestion
    System.out.println(String.format("%s version %s", APP_NAME, APP_VERSION));
```
Just for the sake of having at least a little less duplication than the line that's officially logged.

src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -22,24 +22,24 @@
...
+    Logger.setup();
+    Logger.log(Level.INFO, String.format("Launching %s version %s", APP_NAME, APP_VERSION));
```
HEdingfield:
```suggestion
    Logger.log(Level.INFO, String.format("Launching %s version %s...", APP_NAME, APP_VERSION));
```

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -126,7 +126,7 @@
   // returns: set containing winner(s)
   Set<String> tabulate() {
     logSummaryInfo();
-    Logger.log(Level.INFO, "Starting tabulation...");
+    Logger.log(Level.INFO, "Starting tabulation for %s", this.config.getContestName());
```
HEdingfield:
```suggestion
    Logger.log(Level.INFO, "Starting tabulation for contest '%s'...", this.config.getContestName());
```

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -236,7 +236,7 @@
       }
     }
 
-    Logger.log(Level.INFO, "Tabulation completed.");
+    Logger.log(Level.INFO, "Completed tabulation for %s", this.config.getContestName());
```
HEdingfield:
```suggestion
    Logger.log(Level.INFO, "Completed tabulation for contest '%s'.", this.config.getContestName());
```

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -100,10 +100,9 @@ void convertToCdf() {
   // purpose: run tabulation
   // returns: list of winners
   void tabulate() {
+    Logger.log(Level.INFO, "Starting tabulation session.");
```
```suggestion
    Logger.log(Level.INFO, "Starting tabulation session...");
```

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -166,9 +161,8 @@ private boolean setUpLogging(ContestConfig config) {
   // param: config object containing CVR file paths to parse
   // returns: set of winners from tabulation
   private Set<String> runTabulationForConfig(ContestConfig config) {
+    Logger.log(Level.INFO, "Beginning tabulation for config:%s", configPath);
```
```suggestion
    Logger.log(Level.INFO, "Beginning tabulation for config: %s", configPath);
```

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -188,15 +182,14 @@ private boolean setUpLogging(ContestConfig config) {
         Logger.log(Level.SEVERE, "No cast vote records found!");
       }
     } else {
-      Logger.log(Level.SEVERE, "Skipping tabulation due to source file errors!");
+      Logger.log(Level.SEVERE, "Skipping tabulation due to source CVR file errors!");
     }
 
     if (winners.size() > 0) {
-      Logger.log(Level.INFO, "Tabulation process completed.");
+      Logger.log(Level.INFO, "Tabulation session completed.  Results written to %s", outputPath);
```
```suggestion
      Logger.log(Level.INFO, "Tabulation session completed.  Results written to: %s", outputPath);
```

src/main/java/com/rcv/GuiConfigController.java
moldover: Sure I get the concept.  What is this 'standard CLI messaging' you're talking about?

Issue #197, opened 2018-12-17
Create Herb60 and maybe tiebreak regression tests


moldover:
I'm not sure how these are supposed to play out.
Multi_winner_Herb60
test_tiebreak

gngilbert:
I don't recall the reason for originating these suggested tests but they probably don't need to be pursued as long as we cover the needed functionality tests.

HEdingfield:
Both tests were deleted in #295, but per my comments in that PR, I'd still like us to keep a copy of the tiebreaker files:

> Can we please keep `tiebreak_test_TODO` (and maybe just rename it to `test_interactive_tiebreaker`)? It's currently the only way I know of to trigger the interactive tiebreaker with our test files, and that's important for both development purposes and for allowing users to see the full extent of the software's capabilities (I'm planning to include this with the download).

Again, as I mentioned in the PR comments, I think we should make the distinction between "test" input and "example" input here. This is a case of example input; it's not directly related to a unit test, but I want to include it as a sample file for users to play with in #281.

@moldover please re-add and rename the config and CVR when you get a chance. I'll reorganize the unit tests and example input as part of #78.

moldover:
@HEdingfield yeah that makes sense.  I'll put them in the next PR and close this since this issue has been addressed.

HEdingfield:
I'd rather keep this one open as opposed to creating a new one to remember to get those files re-added. I'll add myself as an assignee too so I can remember to close it once the files are back in.

HEdingfield:
Closed with #316.

Issue #136, opened 2018-08-29
Align default values in ElectionConfig with GUI, and consider default values for all


HEdingfield:
1. Make sure all the de facto default values in the `ElectionConfig` getters/setters align with the documentation and config creator GUI
2. Discuss pre-populating / offering default values for all fields

chughes297:
Doesn't sound like this is necessary for cert but happy to provide feedback on what default values make the most sense.

HEdingfield:
May be related to #211, and may be a moot point now. Verify.

moldover:
@HEdingfield @tarheel circling back to part 2 of this.  In general I think default-values-for-all is the right way to go, but I'm not sure if it's too late for this given the flurry of recent settings changes, and the fact that this would break all the unit tests again.

The way I envision it, every setting would have a default specified in the documentation.txt, which would be applied when config is loaded into the gui, to any missing values.  For example, today a config is valid without any undervoteLabel setting, or if undervoteLabel=null.  We would change this to require undervoteLabel be present to validate.  On loading a file into the GUI a null undervoteLabel would be replaced with "".

HEdingfield:
I'm 95% sure what you just described is indeed the case now that I've switched a lot of the fields to Strings. If you load a config without `undervoteLabel` present, or with the value set to `null` in the JSON, it'll then ask you to save before you tabulate and when you save it'll replace it with "".

The only other situation I can think of that might be relevant is when a key isn't present for something like `maxRankingsAllowed`, you load it, then it tells you you need to save; it saves it as `maxRankingsAllowed` = "", instead of `maxRankingsAllowed` = "max", which is the GUI suggested value. I think that's fine though, as the idea is that the GUI should save whatever's in the fields and do no subbing behind the scenes. The suggested values should only apply when you start a new config file.

moldover:
So saving shouldn't apply deafult, but loading would, right?
In this example the max rankings key isn't present (null) on loading GuiConfigController.loadConfig would notice, and set "max" into the text editor instead and log it.  The ui edit listener counts this as an edit and will prompt to save.  User can now make any edits.  On saving, no defaults are applied, simply write whatever is in the boxes.  They should all be strings at this point.

tarheel:
I believe that @moldover's maxRankingsAllowed example is analogous to what I did with winnerElectionMode: the UI populates it with "standard" by default, and if you load a config that has a missing or invalid value, the config value gets ignored and the selector stays set to "standard".

[...]

Issue #43, opened 2018-04-12
Improve forensics / reproducibility of output


HEdingfield:
Forensics is critical: output should include details on the machine/system, version, date/time, config file, etc. In addition to log and results visualizer files, also output the config file used for tabulation.

HEdingfield:
It's also probably really important that we ensure any future versions of the tabulator are backwards-compatible with older config files. Should we open a separate issue for this?

tarheel:
> It's also probably really important that we ensure any future versions of the tabulator are backwards-compatible with older config files. Should we open a separate issue for this?

Yes. I would say once we have any actual external users, we should start supporting config versions. But for now the config format is still changing quite a bit...

moldover:
Good call, we should def spit out the config, version (date/time is in every log statement, but sure).  I'm not sure what aspects of the system are relevant, but certainly current user, working directory and input/output file hashes when we have them.

As for versions, that's another good point.  I agree with Louis, things are way too in flux, and no external users to justify this yet, but we should plan on it.  I'm not convinced we need to be backwards compatible in general.  It's totally reasonable to expect users who want to run old configs to use an older version of the tabulator.  That said, we may as well and add a version number to the config so it's available when we do break something.

HEdingfield:
This popped into my consciousness again when working on stuff recently: we should output the config file along with the csv and visualizer json whenever we run the tabulator.

nealmcb:
Most important is transparency around backwards-compatiibility.  I recommend semantic versioning: you get to do backwards-incompatible changes, but have to make it obvious to the users.

See also comments in #170 since getting any necessary randomness to be generated and reported in a reproducible fashion is tricky and important.  See e.g. Rivest's "sampler" PRNG at https://github.com/nealmcb/rivest-sampler-tests/tree/python3-port

gngilbert:
Not sure how much we can or need to accomplish for this round of testing but I think this needs to be addressed seriously asap.

HEdingfield:
We'll:
- Adopt semantic versioning (this first certification deliverable will be tagged as 1.0.0)
- Include RCV Tabulator version in the log
- Spit raw contents of config file into log
- Try to include info on machine / OS running the tabulation in the log

TBD:
- Including RCV Tabulator version in the config files somehow

HEdingfield:
[...]

Issue #281, opened 2019-06-10
Decide if Gradle distribution should include test files and docs


HEdingfield:
Currently using `distZip` and `jlink` to create distributions doesn't include any of the test configs / CVRs or docs, which we may want to include somehow in the distribution. Do we want to include these and, if so, which specifically?

Some info on adding static files here:
https://docs.gradle.org/4.10.3/userguide/application_plugin.html#sec:the_distribution

tarheel:
Yes, but let's limit it to the smaller tests, maybe?

HEdingfield:
Makes sense to me. Maybe we only need to include a couple of them? Enough to get coverage of most of the tabulator's functionality? I'll rely on you and/or George to make the call on which to include. Just list them here, along with the docs that should be included (guessing README and UserGuide at least).

freedomcounts:
I agree - README, UserGuide, and to be determined test configs & CVR's.

HEdingfield:
#78 is a partial blocker.

HEdingfield:
I've made progress including files with the `distZip` method, but we're probably going to want to rely on `jlink` instead, and I haven't had any luck so far figuring out how to include files in the `jlink` image. Filed a question on SO [here](https://stackoverflow.com/questions/56778477/how-to-bundle-extra-files-with-jlink-image).

HEdingfield:
Closed with #316.

PR #295, opened 2019-06-21
More cert fixes

moldover:
This addresses #197 and #43.  For #43, I do think we should add a version field to the config file, it's going to be worth the effort, but I'm going to do that with the work for #136 which will also require updating all the regression test files.


src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -40,7 +40,8 @@ public static void main(String[] args) {
     System.out.println(String.format("%s version %s", APP_NAME, APP_VERSION));
     Logger.setup();
     Logger.log(Level.INFO, String.format("Launching %s version %s...", APP_NAME, APP_VERSION));
-
+	  Main.logSystemInfo();
```
tarheel: indentation is off

tarheel: also, the `Main.` part is unnecessary here

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -162,6 +164,18 @@ private boolean setUpLogging(ContestConfig config) {
   // returns: set of winners from tabulation
   private Set<String> runTabulationForConfig(ContestConfig config) {
     Logger.log(Level.INFO, "Beginning tabulation for config: %s", configPath);
+    try {
+      BufferedReader reader = new BufferedReader(new FileReader(configPath));
+      String line = reader.readLine();
+      while(line != null) {
```
tarheel: Should we have log lines before and after, e.g. "Begin config file FILENAME" and "End config file"?

src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -83,4 +84,13 @@ public static void main(String[] args) {
 
     System.exit(0);
   }
+
+  public static void logSystemInfo() {
```
HEdingfield:
```suggestion
  private static void logSystemInfo() {
```

src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -83,4 +84,13 @@ public static void main(String[] args) {
...
+    Logger.log(Level.INFO, String.format("Launching %s version %s...", APP_NAME, APP_VERSION));
```
HEdingfield: This line is duplicated; either remove here or just before this function is called.

src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -83,4 +84,13 @@ public static void main(String[] args) {
...
+    Logger.log(Level.INFO, "Host system: %s version %s",
+        System.getProperty("os.name"),
+        System.getProperty("os.version"));
+    Logger.log(Level.INFO, "User: %s", System.getProperty("user.name"));
```
HEdingfield: I'm a little iffy on including this line, as it's getting into PII territory. If you feel strongly about it, we can keep, but I'd prefer we delete it.

src/main/java/network/brightspots/rcv/TabulatorSession.java
HEdingfield: Agree with this. I don't think the filename is necessary though, since it's output immediately before.
`Begin config file contents:`
`End config file contents.`

src/main/java/network/brightspots/rcv/Main.java
HEdingfield: Realized that this is probably somewhat of a moot point, since the file paths to the logs include username too (on Windows at least).

src/main/java/network/brightspots/rcv/Main.java
moldover: yeah totes

moldover: On second thought, before I commit to that I need to know how / what should the interactive tiebreak test, test?

src/main/java/network/brightspots/rcv/Main.java
moldover: I'm not sure about this one either.  I saw the api and thought, that might be useful, but it is a different and new category of data.  @CalebKleppner @gngilbert @chughes297  do you have an opinion: should we log the current user?

src/main/java/network/brightspots/rcv/Main.java
tarheel: Are you proposing to have an automated test for interactive in which the test code simulates a user entering a tiebreak selection? Seems maybe excessive. But having a test case available for humans to try out would be reasonable.

HEdingfield: Agree, I think we need to start distinguishing between "test" files and "sample / example" files. Interactive tiebreaker should definitely be included as sample files in #281, but I think an automated version is much more of an integration / functional test than a unit test, so is out of scope for now.

PR #297, opened 2019-06-23
Fixes #138

### LINKED ISSUE #138: Better resizing reactivity in GUI
- If window is super small, it should have scrollbars
- If window is fully expanded on desktop it should use the space better (especially config creator)
- It should adjust to resizing in between as expected too

HEdingfield:
Fixes #138. Adds scrollbars when window size is small; prevents cutting off the top bar for low resolution displays.

### DIFF
--- a/src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
+++ b/src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
@@ -23,6 +23,7 @@
 <?import javafx.scene.control.ChoiceBox?>
 <?import javafx.scene.control.DatePicker?>
 <?import javafx.scene.control.Label?>
+<?import javafx.scene.control.ScrollPane?>
 <?import javafx.scene.control.Tab?>
 <?import javafx.scene.control.TableColumn?>
 <?import javafx.scene.control.TableView?>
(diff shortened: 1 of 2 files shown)


HEdingfield:
Ignore the weird diff for the FXML file; all I did was wrap the `BorderPanel` with a `ScrollPanel`, so it's mostly just indentation differences.

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
```diff
@@ -33,291 +34,295 @@
...
-              <TextField maxWidth="220.0" promptText="Max Skipped Ranks Allowed"
-                fx:id="textFieldMaxSkippedRanksAllowed"/>
-              <Label text="Max Rankings Allowed"/>
-              <TextField maxWidth="220.0" promptText="Max Rankings Allowed"
-                fx:id="textFieldMaxRankingsAllowed"/>
-              <padding>
-                <Insets bottom="12.0" left="12.0" right="12.0" top="12.0"/>
-              </padding>
-            </VBox>
+  <BorderPane maxHeight="-Infinity" maxWidth="-Infinity" minHeight="-Infinity" minWidth="-Infinity"
```
tarheel: the maxes shouldn't be positive infinity?

HEdingfield: Weird, huh? It's a standard value created by SceneBuilder, and I see it in all sorts of examples online. Assuming it's fine, since it seems to be.

Regarding your main comments: yes, only scrollbars when necessary. Filed #303 for your other question.

PR #300, opened 2019-06-23
output precinct results files properly in sequential mode

### LINKED ISSUE #291: Sequential IRV tabulation only produces precinct-level results for the final instance of IRV
When running a sequential IRV election (as will be used in Utah this fall), precinct-level reports are only produced for the tabulation of the final seat being filled. Need to produce precinct reports for each seat being filled. 

(same example file as in Issue #290)
[2019-06-18_14-36-55_MINNEAPOLIS_W-2_P-05_precinct_summary.xlsx](https://github.com/BrightSpots/rcv/files/3303329/2019-06-18_14-36-55_MINNEAPOLIS_W-2_P-05_precinct_summary.xlsx)

[2017-park-district-1-cvr.xlsx](https://github.com/BrightSpots/rcv/files/3303347/2017-park-district-1-cvr.xlsx)

[Utah Test 2.json.zip](https://github.com/BrightSpots/rcv/files/3303378/Utah.Test.2.json.zip)

tarheel:
Fixes #291.

This change is less complicated than it looks. `getPrecinctFileString` got moved within the file because I made it static. All I did was 1) factor out the existing redundant `getOutputFilePath` logic and 2) make `generatePrecinctSummarySpreadsheets` use that instead of its own logic (which was failing to account for sequential mode).

When @moldover gets to #195, we should include an automated test for this... but for now, I verified it manually. Also ran the full existing test suite, of course.

### DIFF
--- a/src/main/java/network/brightspots/rcv/ResultsWriter.java
+++ b/src/main/java/network/brightspots/rcv/ResultsWriter.java
@@ -366,7 +366,7 @@ class ResultsWriter {
     } catch (IOException exception) {
       Logger.log(
           Level.SEVERE,
-          "Error creating CSV file: %s\n%s\n" + "Please check the file path and permissions!",
+          "Error creating CSV file: %s\n%s\nPlease check the file path and permissions!",
           csvPath,
           exception.toString());
       throw exception;
--- a/src/main/java/network/brightspots/rcv/ResultsWriter.java
+++ b/src/main/java/network/brightspots/rcv/ResultsWriter.java
@@ -123,8 +123,11 @@ class ResultsWriter {
     try {
       jsonWriter.writeValue(outFile, json);
     } catch (IOException exception) {
-      Logger.log(Level.SEVERE, "Error writing to JSON file: %s\n%s\n"
-          + "Please check the file path and permissions!", path, exception.toString());
+      Logger.log(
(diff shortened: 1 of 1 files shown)


src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -336,8 +364,11 @@ private void generateSummarySpreadsheet(
       BufferedWriter writer = Files.newBufferedWriter(Paths.get(csvPath));
       csvPrinter = new CSVPrinter(writer, CSVFormat.DEFAULT);
     } catch (IOException exception) {
-      Logger.log(Level.SEVERE, "Error creating CSV file: %s\n%s\n"
-          + "Please check the file path and permissions!", csvPath, exception.toString());
+      Logger.log(
+          Level.SEVERE,
+          "Error creating CSV file: %s\n%s\n" + "Please check the file path and permissions!",
```
moldover: Can combine these strings now.

tarheel:
> We should have a precinct test cvr file which only has 2 or 3 different precincts so the output isn't so overwhelming as the Minneapolis one.

Yeah, it would be nice to dramatically simplify the source files for a number of our existing tests. Agree that it's especially important for ones that involve precincts.

PR #302, opened 2019-06-23
Enables cancel during interactive tiebreaker

### LINKED ISSUE #132: Enable cancel during interactive tiebreaker
Actually enable cancel button for interactive tiebreaker GUI, and also introduce cancel option for CLI version; see:

**Tabulator.java**
`// TODO: If returned eliminatedCandidate is null, infinite loop!`

**TieBreak.java**
```
// TODO: Create and enable cancel option for interactive tiebreaker CLI
// TODO: actually enable cancel button for interactive tiebreaker GUI
```

HEdingfield:
Allows user to cancel during interactive tiebreaker (fixes #132); adds example config file for testing interactive tiebreaker with sequential multi-seat enabled; adds commented command in build.gradle for executing CLI arguments.


src/main/java/network/brightspots/rcv/TabulationCancelledException.java
```diff
@@ -0,0 +1,21 @@
...
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+ * the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with this
+ * program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package network.brightspots.rcv;
+
+public class TabulationCancelledException extends Exception {
```
tarheel: For the other custom exceptions we've created, we've nested them within the most appropriate class. Maybe put this in `Tabulator` or `TieBreak`?

src/main/java/network/brightspots/rcv/GuiTiebreakerController.java
```diff
@@ -28,6 +28,7 @@
 public class GuiTiebreakerController {
 
   private String candidateToEliminate = null;
+  private boolean tabulationCancelled = false;
```
tarheel: Super nitpicky, but American English favors canceled/canceling over cancelled/cancelling. Whose side are you on?

src/test/resources/network/brightspots/rcv/test_data/tiebreak_test_TODO/tiebreak_interactive_sequential_config.json
```diff
@@ -0,0 +1,57 @@
...
+  ],
+  "rules": {
+    "tiebreakMode": "interactive",
+    "overvoteRule": "alwaysSkipToNextRank",
+    "numberOfWinners": 2,
+    "decimalPlacesForVoteArithmetic": 4,
+    "minimumVoteThreshold": 0,
+    "maxSkippedRanksAllowed": 0,
+    "maxRankingsAllowed": 3,
+    "sequentialMultiSeat": true,
```
tarheel: why did you pick this complicated setup for your test?

src/main/java/network/brightspots/rcv/TabulationCancelledException.java
HEdingfield: Done, thanks.

src/main/java/network/brightspots/rcv/GuiTiebreakerController.java
HEdingfield: I hadn't thought about this too much, but I'm on the side of "cancelled" because "cancellation" is double-l no matter which version of English you use.

src/test/resources/network/brightspots/rcv/test_data/tiebreak_test_TODO/tiebreak_interactive_sequential_config.json
HEdingfield: I just duplicated the existing tiebreak* test and switched it to sequentialMutliSeat true so it would explore that branch of the code. This was necessary for making sure the cancellation worked in every case, and figured we might as well keep it around for better coverage.

Issue #304, opened 2019-06-23
fix noinspection unchecked for excluded CheckBox in GuiConfigController


tarheel:

HEdingfield:
This applies to `GuiConfigController.initialize()`:
```
// noinspection unchecked
return new SimpleObjectProperty(checkBox);
```

HEdingfield:
I believe this was also resolved with #462.

HEdingfield:
The "fix" in #462 resolved the warning, but it also broke checkboxes in the Candidates table (now they only appear as "true" or "false" and it's uneditable. Need to roll this back.

HEdingfield:
Duplicate of #280.

PR #306, opened 2019-06-23
remove TODOs from code

tarheel:
For each one, I either a) verified that it was no longer relevant, b) verified that an issue already existed (see #132), or c) created a new issue (see #304 and #305).

### DIFF
--- a/src/main/java/network/brightspots/rcv/Tabulator.java
+++ b/src/main/java/network/brightspots/rcv/Tabulator.java
@@ -915,12 +915,6 @@ class Tabulator {
             continue;
           }
 
-          // handle testing in unit tests, remove this assert, and add break statement
-          // at end of this for loop
-
-          // If this fails, it means the code failed to handle an overvote with multiple
-          // continuing candidates.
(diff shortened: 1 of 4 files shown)


src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -917,7 +915,6 @@ private void recordSelectionForCastVoteRecord(
             continue;
           }
 
-          // TODO: it's weird to have this check here
```
HEdingfield: It seems like the rest of the stuff below here may have been part of the TODO? Does it make sense on its own?
```
// handle testing in unit tests, remove this assert, and add break statement
// at end of this for loop
```

tarheel: Oops. I just went ahead and did what @moldover suggested, except for adding a unit test. I'll create a new issue for that part.

HEdingfield:
We've also got FIXMEs in there that we should probably remove too.

PR #307, opened 2019-06-23
add Winner(s) row in CSV output

### LINKED ISSUE #236: Revisions to .csv summary file
Quick notes here, with more to follow, but here are two quick and easy changes (I hope) for our .csv summary file (not important for .json).

1.	List all winners above threshold (new row in report)
2.	Change the term “Exhausted ballots” to “Inactive ballots”

[Issue -- summary format changes.pdf](https://github.com/BrightSpots/rcv/files/2922494/Issue.--.summary.format.changes.pdf)

tarheel:
Fixes #236

### DIFF
--- a/src/main/java/network/brightspots/rcv/CastVoteRecord.java
+++ b/src/main/java/network/brightspots/rcv/CastVoteRecord.java
@@ -104,7 +104,7 @@ class CastVoteRecord {
     if (outcomeType == VoteOutcomeType.IGNORED) {
       logStringBuilder.append(" [was ignored] ");
     } else if (outcomeType == VoteOutcomeType.EXHAUSTED) {
-      logStringBuilder.append(" [was exhausted] ");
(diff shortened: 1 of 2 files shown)


tarheel:
> One small note: in issue #236, you mentioned that you changed "exhausted" to "inactive", and I verified this in the CSV. I noticed that the log file stills says "was exhausted" though; is this something we also want to change?

Um, I'm not sure. I think "exhausted" is more clear for the logs. @gngilbert: what do you think? Why do we use "inactive" in the CSV output (which was a request from @CalebKleppner)? Is there any difference between the two terms?

CalebKleppner:
It's a public relations thing.?? People seem to accept "inactive" better 
than "exhausted".?? Not important to change the log entries, but if it's 
easy, you could change "was exhausted" to "became inactive".

Caleb Kleppner
203-376-4080

PR #311, opened 2019-06-24
add onlyOneWinnerPerRound option

### LINKED ISSUE #288: add option for batch winner election
The current behavior in the code is to elect all candidates exceeding the winning threshold simultaneously (i.e. in a single round). However, some jurisdictions (e.g. Eastpointe) have rules stating the only the candidate with the highest total should be elected in a given round. We need a boolean config param for this.

I figure we should call this "batch winner election" or something, because it's the counterpart to batch elimination. But let me know if you have a better suggestion for the name.

tarheel:
Fixes #288.

Will update with some kind of tiebreak logic once we have consensus on how that should work.


src/test/resources/network/brightspots/rcv/test_data/test_set_only_one_winner_per_round/test_set_only_one_winner_per_round_config.json
```diff
@@ -0,0 +1,57 @@
...
+    "contestName" : "One Winner per Round Test",
+    "outputDirectory" : "output",
+    "contestDate" : "2019-01-18",
+    "contestJurisdiction" : "",
+    "contestOffice" : "",
+    "tabulateByPrecinct" : false,
+    "generateCdfJson" : false
+  },
+  "cvrFileSources" : [ {
+    "filePath" : "/Users/leisenberg/code/rcv/src/test/resources/network/brightspots/rcv/test_data/test_set_only_one_winner_per_round/test_set_only_one_winner_per_round_cvr.xlsx",
```
HEdingfield: Bad file path in here; it's hard-coded to your system.

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
```diff
@@ -253,6 +256,12 @@
                     <Insets top="4.0"/>
                   </VBox.margin>
                 </CheckBox>
+                <CheckBox mnemonicParsing="false" text="Only One Winner per Round"
```
HEdingfield: Is "Allow only one winner per round" a better description? "Force only one winner per round"? If so, I'd refactor all `*OnlyOneWinnerPerRound` variables accordingly. Or maybe I'm off base and this is fine.

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
```diff
@@ -156,6 +156,11 @@ Config file must be valid JSON format. Examples can be found under the test fold
         value: true | false
         if not supplied: false
 
+      "onlyOneWinnerPerRound" optional
+        in a multi-seat contest, elect no more than one winner per round, even when there are multiple candidates exceeding the winning threshold
```
HEdingfield: This makes me think we should add a validation rule that checks to make sure it's not on in a non-multi-seat election.

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -473,6 +473,24 @@ private CandidateStatus getCandidateStatus(String candidate) {
...
 
+    // Edge case: if we've identified multiple winners in this round but we're only supposed to
+    // elect one winner per round, pick the top vote-getter and defer the others to subsequent
+    // rounds.
+    if (config.isOnlyOneWinnerPerRoundEnabled() && selectedWinners.size() > 1) {
+      BigDecimal maxTally = BigDecimal.ZERO;
+      String candidateWithMaxTally = null;
+      for (String candidate : selectedWinners) {
+        BigDecimal tally = currentRoundCandidateToTally.get(candidate);
+        // TODO: need a non-arbitrary way of settling a tie here
```
HEdingfield: Definitely need to address this before submitting this PR. Maybe email George and the others directly to get a quicker response?

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -473,6 +473,24 @@ private CandidateStatus getCandidateStatus(String candidate) {
...
+        if (tally.compareTo(maxTally) > 0) {
+          maxTally = tally;
+          candidateWithMaxTally = candidate;
```
HEdingfield: Can you updated `selectedWinners` here and `break`?

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
moldover: Fine by me either way.  I'll throw out there the concise "One winner per round"

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
moldover: The validation feedback on existing rules was quite helpful when I was testing and trying to break things :)

moldover:
Looks great!  If you can hook into the existing tiebreak logic as you propose this would be super solid, but that might be a big task.  Does that mean a new tiebreak gui resource and logic?

tarheel:
> Looks great! If you can hook into the existing tiebreak logic as you propose this would be super solid, but that might be a big task. Does that mean a new tiebreak gui resource and logic?

I'm hoping I can repurpose all the existing logic in an elegant way. If it proves difficult, I'll consider other options.

src/test/resources/network/brightspots/rcv/test_data/test_set_only_one_winner_per_round/test_set_only_one_winner_per_round_config.json
tarheel: bitmoji

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
tarheel: I think "allow only one winner per round" is good.

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
tarheel: I thought about doing that, but I wasn't sure if it made sense. Setting this to true isn't actually incompatible with a single-seat contest; it's just irrelevant.

src/main/java/network/brightspots/rcv/Tabulator.java
tarheel: No, I need to look at all of them to find the max.

tarheel:
OK, I've addressed your comments, including adding full tie-break support.

src/main/java/network/brightspots/rcv/TieBreak.java
```diff
@@ -135,61 +138,64 @@ String getExplanation() {
...
   // function: breakTie
   // purpose: execute the tiebreak logic given the tiebreak rule in use
-  // returns: losing candidate
-  String selectLoser() throws TabulationCancelledException {
+  // returns: selected candidate
+  String selectCandidate() throws TabulationCancelledException {
     switch (tieBreakMode) {
       case INTERACTIVE:
-        loser = doInteractive();
+        doInteractive();
```
HEdingfield: It feels creepy and uncomfortable to me to have this (and the others below) modify a class variable and then return that instead of working with an instance variable like before. I'd advise against it unless there's a really good reason.

src/main/java/network/brightspots/rcv/TieBreak.java
```diff
@@ -135,61 +138,64 @@ String getExplanation() {
...
   }
 
   // function: doPermutationSelection
-  // purpose: select the loser based on the provided permutation order
-  // returns: selected loser
-  private String doPermutationSelection() {
-    // loser to return
-    String selectedCandidate = null;
+  // purpose: select the candidate based on the provided permutation order
+  private void doPermutationSelection() {
```
HEdingfield: This method is over my head. Hoping @moldover can review.

tarheel: It's nothing fancy. The idea is that there's a permutation (ordering) of all the candidates that's intended to be used to break any ties, and the order is from most favored to least favored. So when you're breaking a tie among a set of candidates, if it's to find a loser, you identify the candidate among the set who appears latest in the ordering; and if it's to find a winner, you identify the one who appears earliest. Does that make sense?

src/main/java/network/brightspots/rcv/TieBreak.java
tarheel: fiiiiiine

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -66,6 +66,7 @@
   static final boolean SUGGESTED_CANDIDATE_EXCLUDED = false;
   static final boolean SUGGESTED_SEQUENTIAL_MULTI_SEAT = false;
   static final boolean SUGGESTED_BOTTOMS_UP_MULTI_SEAT = false;
+  static final boolean SUGGESTED_ALLOW_ONLY_ONE_WINNER_PER_ROUND = false;
```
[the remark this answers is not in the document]
tarheel: Nope. This whole setting is irrelevant if it's a single-seat contest; it doesn't matter if it's false or true.

src/main/java/network/brightspots/rcv/TieBreak.java
HEdingfield: Sure, sounds good, I trust you. 💯

PR #313, opened 2019-06-24
add random seed

### LINKED ISSUE #170: random seed for tiebreaking
If tiebreakMode = random, we should require the config to specify a random seed also. This will ensure consistent results across runs.

tarheel:
Fixes #170.

I'll need to update some of the existing test configs before committing this, since they'll now fail the validation that I added.


src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -352,17 +354,26 @@ private void validateCandidates() {
...
       Logger.log(Level.SEVERE, "Invalid tie-break mode!");
     }
 
-    if (getOvervoteRule() == Tabulator.OvervoteRule.RULE_UNKNOWN) {
+    if ((getTiebreakMode() == TieBreakMode.RANDOM ||
+        getTiebreakMode() == TieBreakMode.PREVIOUS_ROUND_COUNTS_THEN_RANDOM ||
+        getTiebreakMode() == TieBreakMode.GENERATE_PERMUTATION) &&
+        getRandomSeed() == null) {
+      isValid = false;
+      Logger.log(Level.SEVERE, "When tiebreakMode involves a random element, randomSeed must be "
```
HEdingfield: If I set the tiebreak mode to "generatePermutation" and leave the seed box empty, it doesn't report this message for some reason (but it does with the other two). Instead I get:

> 2019-06-25 10:02:36 PDT SEVERE: Error processing candidate data:
> 2019-06-25 10:02:36 PDT SEVERE: Error during validation:
> java.lang.NullPointerException
> Validation failed!

src/main/java/network/brightspots/rcv/TieBreak.java
```diff
@@ -88,6 +91,10 @@
     this.candidatePermutation = candidatePermutation;
   }
 
+  static void setRandomSeed(int randomSeed) {
+    random = new Random(randomSeed);
```
HEdingfield: Minor code smell for me here... I think it would be better if you actually had a `randomSeed` field, not just a `random` field, stored in the class, and here you both set the `randomSeed` field and initialize `random` using that field. From a cursory Google search, it seems to be relatively unintuitive to extract a seed from an initialized `Random` object. I think we probably want both fields to be `final` too.

src/test/resources/network/brightspots/rcv/test_data/tiebreak_seed_test/tiebreak_seed_test_config.json
```diff
@@ -0,0 +1,49 @@
...
+  }, {
+    "name" : "Yinka Dare",
+    "excluded" : false
+  }, {
+    "name" : "George Gervin",
+    "excluded" : false
+  } ],
+  "rules" : {
+    "tiebreakMode" : "random",
+    "randomSeed" : 1,
```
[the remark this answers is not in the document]
moldover: It is plenty random, but just not reproducible if you don't use the same seed every time.  There were no tests using this.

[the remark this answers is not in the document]
moldover: It's worth noting, adding this test doesn't say anything about randomness in our code.  We rely on the java.lang.Math.random implementation.  The test only says that using the same seed returns the same sequence every time.

src/main/java/network/brightspots/rcv/TieBreak.java
tarheel: We're not accessing `randomSeed` in `TieBreak`, though. If we decide we want to in the future, we can just refactor it accordingly.

And I can't make `random` final because it doesn't necessarily get initialized.

tarheel:
OK, it's ready for review again.

tarheel:
Rebased on top of the latest changes in `develop`.

src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -175,6 +175,7 @@ public void setExcluded(boolean excluded) {
   public static class ContestRules {
     // tiebreak mode to use
     public String tiebreakMode;
+    public Integer randomSeed;
```
HEdingfield: Nit: put this after overvoteRule since that's its order everywhere else.

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
```diff
@@ -111,6 +111,12 @@ Config file must be valid JSON format. Examples can be found in the test_data fo
         how the program should handle an overvote when it encounters one
         value: "alwaysSkipToNextRank" | "exhaustImmediately" | "exhaustIfMultipleContinuing"
 
+      "randomSeed" optional
+        the integer seed for the application's pseudorandom number generator
+        required if tiebreakMode is random, previousRoundCountsThenRandom, or generatePermutation
```
HEdingfield: Move this "required" line up to replace "optional" , as we've done with other conditionally required fields.

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
```diff
@@ -111,6 +111,12 @@ Config file must be valid JSON format. Examples can be found in the test_data fo
...
+        value: [0..2147483647]
```
HEdingfield: Please also add this upper bound to the two other fields in this file that have "..]".

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -803,7 +824,11 @@ private void processCandidateData() {
     }
 
     if (getTiebreakMode() == TieBreakMode.GENERATE_PERMUTATION) {
-      Collections.shuffle(candidatePermutation);
+      // If the random seed is null, validation will fail anyway, so let's not throw an exception
+      // here.
```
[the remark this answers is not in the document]
tarheel: Hmm, I think I disagree. Without the comment, someone reading this code might think that it's a legitimate use case to have mode = GENERATE_PERMUTATION and random seed = null.

@moldover thoughts?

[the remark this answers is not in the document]
tarheel: I've updated the comment to make it more clear what the issue is.

Issue #83, opened 2018-07-23
better error detection for malformed source files


tarheel:
If a source file exists but has issues, the code doesn't handle it gracefully.

Things to check:
- invalid file type (renamed to .xlsx) (see #267)
- missing columns
- missing IDs or precincts
- unexpected cell type (see #180)  firstVoteColumnIndex was mistakenly pointing at the column with the ballot IDs,
- other?

CalebKleppner:
This should be taken care of when our validate config file routine starts working in v 0.8 GUI, so I put this on milestone v 0.9 to make sure we've covered all the bases by then.

moldover:
@gngilbert @CalebKleppner  I'm putting in some stricter checks around CVR file parsing to help users identify problems with CVR files.  If a user specifies precinct ID column, should we fail the tabulation if we encounter ballots which don't have a precinct ID?

CalebKleppner:
I'd say that we should not stop the tabulation, since we sometimes get 
an odd file or two lacking a precinct identifier.  What would happen to 
precinct report if you have records with no precinct identifier?  If 
they can all be collected in [null] or [precinct ID missing], that would 
be ideal.

Caleb Kleppner
MK Election Services, LLC
203-376-4080

CalebKleppner:
I'd say that we should not stop the tabulation, since we sometimes get 
an odd file or two lacking a precinct identifier.  What would happen to 
precinct report if you have records with no precinct identifier?  If 
they can all be collected in [null] or [precinct ID missing], that would 
be ideal.

Caleb Kleppner
MK Election Services, LLC
203-376-4080

On 6/24/2019 12:53 PM, Jon Moldover wrote:

Caleb Kleppner
203-376-4080

moldover:
Sure, that should be easy enough.

moldover:
Closed #314

PR #314, opened 2019-06-25
Bad cvr checks

moldover:
Fixes for #83.  Handles ballots missing precinct Ids by assigning them to missing_precinct_id and added a unit test for this case.  Issues warnings for missing ballot Ids and some user suggestions when errors are encountered.


src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -103,30 +105,40 @@ void tabulate() {
...
-        config.setNumberOfWinners(numWinners);
-        for (String winner : config.getSequentialWinners()) {
-          config.setCandidateExclusionStatus(winner, false);
+        if (config.isSequentialMultiSeatEnabled()) {
+          int numWinners = config.getNumberOfWinners();
+          // temporarily set config to single-seat so we can run sequential elections
+          config.setNumberOfWinners(1);
+          while (config.getSequentialWinners().size() < numWinners) {
+            // Read cast vote records and precinct IDs from CVR files
+            List<CastVoteRecord> castVoteRecords = parseCastVoteRecords(config, precinctIDs);
```
HEdingfield: This same block seems to appear in both parts of the if statement (if and else); maybe it's possible to factor outside of it? Ignore me if this is wrong.

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -103,30 +105,40 @@ void tabulate() {
...
+            if (castVoteRecords == null) {
```
HEdingfield: No need to check for `castVoteRecords.isEmpty()` also? I see that was done previously.

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -103,30 +105,40 @@ void tabulate() {
...
+              Logger.log(Level.SEVERE, "Aborting tabulation due to cvr errors!");
```
HEdingfield: Here and also in other log messages in this PR: you use "cvr", but you also use "cast vote record" in some other log messages. I recommend doing a find / replace to consistently stick with the latter.

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -231,21 +228,39 @@ private boolean setUpLogging(ContestConfig config) {
...
         }
+        // various incorrect settings can lead to UnrecognizedCandidatesException so it's hard
+        // to know exactly what the problem is
+        Logger.log(Level.INFO, "Check config settings for candidate names, firstVoteRowIndex, "
+            + "firstVoteColumnIndex, and precinctColumnIndex to make sure they are correct!");
+        Logger.log(Level.INFO, "See config_file_documentation.txt for more details.");
         encounteredSourceProblem = true;
       } catch (IOException e) {
-        Logger.log(Level.SEVERE, "Error opening source file %s", cvrPath);
+        Logger.log(Level.SEVERE, "Error opening cast vote record file %s", cvrPath);
```
```suggestion
        Logger.log(Level.SEVERE, "Error opening cast vote record file: %s", cvrPath);
```

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -231,21 +228,39 @@ private boolean setUpLogging(ContestConfig config) {
...
+      } catch (ParserConfigurationException |
+          SAXException |
+          OpenXML4JException |
+          POIXMLException e) {
         Logger.log(Level.SEVERE, "Error parsing source file %s", cvrPath);
+        Logger.log(Level.INFO, "ES&S cast vote record files must be Microsoft Excel Workbook "
+            + "format.\nStrict Open XML and Open Office are not supported." );
+        encounteredSourceProblem = true;
+      } catch (CvrDataFormatException e) {
+        Logger.log(Level.SEVERE, "Data format error while parsing source file %s", cvrPath);
```
```suggestion
        Logger.log(Level.SEVERE, "Data format error while parsing source file: %s", cvrPath);
```

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -160,50 +172,35 @@ private boolean setUpLogging(ContestConfig config) {
...
+    Tabulator tabulator = new Tabulator(castVoteRecords, config, precinctIDs);
+    // do the tabulation
+    winners = tabulator.tabulate();
+    // generate visualizer spreadsheet data
+    try {
+      tabulator.generateSummaryFiles(timestampString);
+    } catch (IOException e) {
+      Logger.log(Level.SEVERE, "Error writing summary files:\n%s", e.toString());
     }
+    Logger.log(Level.INFO, "Tabulation session completed.  Results written to: %s", outputPath);
```
```suggestion
    Logger.log(Level.INFO, "Tabulation session completed. Results written to: %s", outputPath);
```

src/main/java/network/brightspots/rcv/StreamingCVRReader.java
```diff
@@ -80,9 +82,10 @@
   private List<CastVoteRecord> cvrList;
   // store precinctIDs (new IDs will be added as we parse)
   private Set<String> precinctIDs;
-
   // last rankings cell observed for CVR in progress
   private int lastRankSeen;
+  // flag indicating data issues during parsing
+  public Boolean encounteredDataErrors = false;
```
HEdingfield:
```suggestion
  private boolean encounteredDataErrors = false;
```

src/main/java/network/brightspots/rcv/StreamingCVRReader.java
```diff
@@ -184,6 +187,24 @@ private void endCVR() {
...
+        // group precincts with missing Ids here
+        Logger.log(Level.WARNING, "Precinct identifier not found for Cvr: %s",
+            computedCastVoteRecordID);
+        currentPrecinct = MISSING_PRECINCT_ID;
+      }
+      precinctIDs.add(currentPrecinct);
+    }
+
+    // look for missing Cvr Id
+    if (idColumnIndex != null && currentSuppliedCVRID == null) {
```
HEdingfield: `currentSuppliedCVRID` => `currentSuppliedCvrId`

src/main/java/network/brightspots/rcv/StreamingCVRReader.java
```diff
@@ -50,6 +50,8 @@
 
 class StreamingCVRReader {
 
+  // this indicates a missing precinct Id in output files
+  private final String MISSING_PRECINCT_ID = "missing_precinct_id";
```
HEdingfield: This can be a local variable. Same with `tabulationLogPath`.

src/main/java/network/brightspots/rcv/TabulatorSession.java
moldover: Right, that's what I was slacking with Louis about before.  This needs to be called in both scenarios but for sequential it happens multiple times.

src/main/java/network/brightspots/rcv/TabulatorSession.java
moldover: That's right, I moved that check inside of parseCastVoteRecords.

src/main/java/network/brightspots/rcv/StreamingCVRReader.java
tarheel: I prefer to pull something like this out into a constant, since we might end up referencing it elsewhere.

src/main/java/network/brightspots/rcv/StreamingCVRReader.java
tarheel: +1. no need to use an object here

src/main/java/network/brightspots/rcv/StreamingCVRReader.java
```diff
@@ -338,6 +356,15 @@ public void headerFooter(String s, boolean b, String s1) {
     if (this.unrecognizedCandidateCounts.size() > 0) {
       throw new UnrecognizedCandidatesException(unrecognizedCandidateCounts);
     }
+    // throw if there were data issues
```
tarheel: not trying to nitpick, but I don't think comments like this make the code more readable; your code is already super clear here because you're using descriptive names

src/main/java/network/brightspots/rcv/StreamingCVRReader.java
```diff
@@ -338,6 +356,15 @@ public void headerFooter(String s, boolean b, String s1) {
     if (this.unrecognizedCandidateCounts.size() > 0) {
```
tarheel: you don't need `this.`

...unless this is a deliberate stylistic thing? But we don't do it elsewhere, AFAIK.

src/main/java/network/brightspots/rcv/StreamingCVRReader.java
```diff
@@ -338,6 +356,15 @@ public void headerFooter(String s, boolean b, String s1) {
...
       throw new UnrecognizedCandidatesException(unrecognizedCandidateCounts);
     }
+    // throw if there were data issues
+    if (this.encounteredDataErrors) {
```
tarheel: no `this.`

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -103,30 +105,40 @@ void tabulate() {
...
-          assert (newWinnerSet.size() == 1);
-          String newWinner = (String) newWinnerSet.toArray()[0];
-          config.setCandidateExclusionStatus(newWinner, true);
-          config.addSequentialWinner(newWinner);
-        }
-        // revert config to original state
-        config.setNumberOfWinners(numWinners);
-        for (String winner : config.getSequentialWinners()) {
-          config.setCandidateExclusionStatus(winner, false);
+        if (config.isSequentialMultiSeatEnabled()) {
```
tarheel: this whole block is indented too much, which is why GitHub thinks all the lines have been changed even though some are identical

src/main/java/network/brightspots/rcv/TabulatorSession.java
tarheel: I don't quite understand why you pulled `parseCastVoteRecords` out of `runTabulationForConfig`. If you left it in there, you could avoid this duplicated logic. If the issue is that you want to break out of the sequential loop, you could do that by just testing whether it returns an empty Set... or you could make it throw an exception that you catch.

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -103,30 +105,40 @@ void tabulate() {
...
+          config.setNumberOfWinners(1);
+          while (config.getSequentialWinners().size() < numWinners) {
+            // Read cast vote records and precinct IDs from CVR files
+            List<CastVoteRecord> castVoteRecords = parseCastVoteRecords(config, precinctIDs);
+            if (castVoteRecords == null) {
+              Logger.log(Level.SEVERE, "Aborting tabulation due to cvr errors!");
+              break;
+            }
+            Set<String> newWinnerSet = runTabulationForConfig(config, castVoteRecords);
+            assert (newWinnerSet.size() == 1);
```
tarheel: to be consistent with elsewhere, no parens

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -231,21 +228,39 @@ private boolean setUpLogging(ContestConfig config) {
...
     }
 
-    if (!encounteredSourceProblem) {
-      Logger.log(Level.INFO, "Parsed %d cast vote records successfully.", castVoteRecords.size());
-    } else {
+    if (encounteredSourceProblem) {
       Logger.log(Level.SEVERE, "Parsing cast vote records failed!");
       castVoteRecords = null;
+    } else if (castVoteRecords.isEmpty()) {
+      Logger.log(Level.SEVERE, "No cast vote records round!");
```
tarheel: s/round/found/

src/main/java/network/brightspots/rcv/StreamingCVRReader.java
moldover: Any constant can be converted to a local variable.  It's here as a more central location self-documenting location.

src/main/java/network/brightspots/rcv/TabulatorSession.java
moldover: Yeah, I actually tried those variations and landed on this because of the way it more clearly shows the  castVoteRecord parsing as a prerequisite to tabulating.  The high level flow and failure paths made more sense.

src/main/java/network/brightspots/rcv/StreamingCVRReader.java
moldover: Oh that's just a habit from writing old c++

src/main/java/network/brightspots/rcv/TabulatorSession.java
moldover: good catch

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -201,14 +201,14 @@ private void validateOutputSettings() {
   private void validateCvrFileSources() {
     if (rawConfig.cvrFileSources == null || rawConfig.cvrFileSources.isEmpty()) {
       isValid = false;
-      Logger.log(Level.SEVERE, "Contest config must contain at least 1 CVR file!");
+      Logger.log(Level.SEVERE, "Contest config must contain at least 1 cast Vote Record file!");
```
```suggestion
      Logger.log(Level.SEVERE, "Contest config must contain at least 1 cast vote record file!");
```

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -201,14 +201,14 @@ private void validateOutputSettings() {
...
     } else {
       HashSet<String> cvrFilePathSet = new HashSet<>();
       for (CVRSource source : rawConfig.cvrFileSources) {
         // perform checks on source input path
         if (source.getFilePath() == null || source.getFilePath().isEmpty()) {
           isValid = false;
-          Logger.log(Level.SEVERE, "filePath is required for each CVR file!");
+          Logger.log(Level.SEVERE, "filePath is required for each cast Vote Record file!");
```
```suggestion
          Logger.log(Level.SEVERE, "filePath is required for each cast vote record file!");
```

src/main/java/network/brightspots/rcv/StreamingCVRReader.java
```diff
@@ -184,23 +187,39 @@ private void endCVR() {
...
+        Logger.log(Level.WARNING, "Precinct identifier not found for cast vote record: %s",
+            computedCastVoteRecordID);
+        currentPrecinct = MISSING_PRECINCT_ID;
+      }
+      precinctIDs.add(currentPrecinct);
+    }
+
+    // look for missing Cvr Id
+    if (idColumnIndex != null && currentSuppliedCvrId == null) {
+      Logger.log(Level.SEVERE, "Cast vote record identifier not found for : %s",
```
HEdingfield:
```suggestion
      Logger.log(Level.SEVERE, "Cast vote record identifier not found for: %s",
```

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -108,8 +110,14 @@ void tabulate() {
         // temporarily set config to single-seat so we can run sequential elections
         config.setNumberOfWinners(1);
         while (config.getSequentialWinners().size() < numWinners) {
-          Set<String> newWinnerSet = runTabulationForConfig(config);
-          assert (newWinnerSet.size() == 1);
+          // Read cast vote records and precinct IDs from CVR files
+          List<CastVoteRecord> castVoteRecords = parseCastVoteRecords(config, precinctIDs);
+          if (castVoteRecords == null) {
+            Logger.log(Level.SEVERE, "Aborting tabulation due to cvr errors!");
```
```suggestion
            Logger.log(Level.SEVERE, "Aborting tabulation due to cast vote record errors!");
```

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -121,11 +129,15 @@ void tabulate() {
         }
       } else {
         // normal operation (not sequential multi-seat)
-        runTabulationForConfig(config);
+        // Read cast vote records and precinct IDs from CVR files
+        List<CastVoteRecord> castVoteRecords = parseCastVoteRecords(config, precinctIDs);
+        if (castVoteRecords == null) {
+          Logger.log(Level.SEVERE, "Aborting tabulation due to cvr errors!");
```
```suggestion
          Logger.log(Level.SEVERE, "Aborting tabulation due to cast vote records errors!");
```

src/main/java/network/brightspots/rcv/StreamingCVRReader.java
moldover: Yeah, when I wrote the comment there was actually more logic here.

src/main/java/network/brightspots/rcv/TabulatorSession.java
moldover: [facepalm] this is what i get for working past my bedtime lol

Issue #211, opened 2019-01-21
Safeguards against changing "if not supplied" behavior in future versions


HEdingfield:
PR #203 included the following conversation:

> @HEdingfield
> Something is telling me that it's better to keep the original design here (using ContestConfig). Maybe this stems from me thinking that explicit is always better than implicit, even for unspecified "default" settings.
> 
> For example, say we have a default value of 2 for a setting in one version of the software. By using ContestConfig, it would actually populate this default value in the GUI, then force the user to re-save before running the tabulator and there's a clear record of what all the settings were. Then in a later update to the software we change the default value to 3. If users don't have that explicitly-defined (default) value in the config, I can imagine them being very confused as to why they can't replicate past results unless they've been following the patch notes closely. Does that make sense?
> 
> Of course, this leaves us with the problem of how to indicate "unlimited" for maxSkippedRank. I think maybe -1 could be a good value for that, and just note the significance in the documentation.

> @tarheel
> I don't really like the idea of using -1 to signify unlimited. It's more intuitive to indicate no limit by just not specifying a limit. But I think we can handle this fine if we update the code so that we don't try to add the default value back when the user has cleared a non-required field.

> @HEdingfield
> I'd like to play around with it a bit once a commit is in to see how this feels, but I think that sounds reasonable.

This solution mostly addresses the problem I brought up, but it seems to me that we'd have to be very careful to not modify "if not supplied" behavior in future versions. Any thoughts on how to safeguard against this, or arguments against it being necessary?

HEdingfield:
We discussed not allowing any "if not supplied", i.e. forcing the user to specify in every single case. The `unlimited` case can be handled by setting an upper limit for `MaxSkippedRanks`, like 10000, and instructing users to input that number.

The result is a config file that explicitly includes all arguments.

moldover:
@HEdingfield @tarheel:  Thoughtful default values (particularly for the many arcane RCV variations we support) provide the most value to our users.  We shouldn't force everyone to spend time learning what a Hare Quota is to use our software.  It isn't scalable.  Put another way, if a user doesn't know or care enough about a parameter to set it, and we are thoughtful in our default values, and we communicate this behavior ("warning: config file doesn't specify 'useHareQuota' defaulting to 'false') we will be in good shape.

moldover:
[...]

PR #322, opened 2019-07-02
Config fixes

moldover:
This addresses #211.  

Null is now written to json in a few places:
maxSkippedRanksAllowed (means "unlimited")
maxRankingsAllowed (means numDeclaredCandidates)
precinctColumnIndex, idColumnIndex, candidate code: means unspecified

This does not change any of the parameter rules or logic.

"" is written for all unsupplied text fields (except as noted, code, which we could change if anyone has strong feelings).


src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -584,6 +584,16 @@ int getMaxRankingsAllowed() {
         : getNumDeclaredCandidates();
   }
 
+  // function: getMaxSkippedRanksAllowed
+  // purpose: getter for maxSkippedRanksAllowed rule
+  // returns: max skipped ranks allowed in this config
+  // a null value here means "unlimited"
+  Integer getMaxSkippedRanksAllowed() {
+    return rawConfig.rules.maxSkippedRanksAllowed != null
```
tarheel: I don't understand this change. The old logic (just `return rawConfig.rules.maxSkippedRanksAllowed`) behaves identically.

src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -47,40 +45,8 @@
   RawContestConfig() {
   }
 
-  // function: validate
-  // purpose: perform some basic validation checks
-  // returns: true if checks pass otherwise false
-  public boolean validate() {
```
tarheel: Confirming my understanding: you removed this method because Jackson will now automatically throw an exception if any of these fields are missing?

moldover: this was never called, I think Hylton removed the calls in an earlier code change.

src/main/java/network/brightspots/rcv/ContestConfig.java
moldover: you are correct [face palm]

moldover:
Not sure how you got overvoteLabel to null.  Lemme do some more testing.

HEdingfield:
@moldover the easiest way to test this is to take an existing config file and go into the json and delete every single field that isn't "required". Then load it in the config, save it without making any changes, and see how everything turns out. This is how I came up with the big list in #211.

moldover:
Ok, thanks.   I think I see what's up now.  Lemme make a few more changes.

moldover:
Ok @tarheel I think I fixed the bug you found.  @HEdingfield I think it's pretty solid.  I tested loading and saving starting from minimal configs, removing various fields,  and everything seems to behave now.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -632,32 +651,25 @@ BigDecimal getMinimumVoteThreshold() {
...
-  Integer getMaxSkippedRanksAllowed() {
-    return rawConfig.rules.maxSkippedRanksAllowed;
-  }
-
   // function: getUndeclaredWriteInLabel
   // purpose: getter for UWI label
   // returns: UWI label for this config
   String getUndeclaredWriteInLabel() {
-    return rawConfig.rules.undeclaredWriteInLabel;
+    return getString(rawConfig.rules.undeclaredWriteInLabel);
```
tarheel: The treatBlankAsUndeclaredWriteIn option doesn't seem to work, but I don't think it's your fault. I'll deal with that in a separate issue: #324

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -793,27 +795,27 @@ private void loadConfig(ContestConfig config) {
...
-    checkBoxSequentialMultiSeat.setSelected(rules.sequentialMultiSeat);
-    checkBoxBottomsUpMultiSeat.setSelected(rules.bottomsUpMultiSeat);
-    checkBoxNonIntegerWinningThreshold.setSelected(rules.nonIntegerWinningThreshold);
-    checkBoxHareQuota.setSelected(rules.hareQuota);
-    checkBoxBatchElimination.setSelected(rules.batchElimination);
-    checkBoxContinueUntilTwoCandidatesRemain.setSelected(rules.continueUntilTwoCandidatesRemain);
-    checkBoxExhaustOnDuplicateCandidate.setSelected(rules.exhaustOnDuplicateCandidate);
-    checkBoxTreatBlankAsUndeclaredWriteIn.setSelected(rules.treatBlankAsUndeclaredWriteIn);
+        textFieldDecimalPlacesForVoteArithmetic, config.getDecimalPlacesForVoteArithmetic());
+    setTextFieldToInteger(textFieldMinimumVoteThreshold, config.getMinimumVoteThreshold().intValue());
```
tarheel: is the call to `intValue` necessary?

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -793,27 +795,27 @@ private void loadConfig(ContestConfig config) {
...
+    setTextFieldToInteger(textFieldMaxSkippedRanksAllowed, config.getMaxSkippedRanksAllowed());
+    // use raw value for storage / persistence as it can be null
+    setTextFieldToInteger(textFieldMaxRankingsAllowed, config.rawConfig.rules.maxRankingsAllowed);
```
tarheel: I introduced the special logic for this field originally, but I'm reconsidering it now. I don't really like that we have to handle this one field differently from all the others here. If we have to read from the config instead of the raw config here (see my other comment), maybe we should push the special logic (fall back to numDeclaredCandidates when this is null) into the callsites? There are only a few.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -769,21 +770,22 @@ private boolean checkForSaveAndTabulate() {
...
 
-  private void loadConfig(ContestConfig config) {
-    clearConfig();
+  // clear widgets then set to default values
+  // then populate widgets with values from input config
+  private void loadConfigIntoGui(ContestConfig config) {
+    clearWidgetsAndSetToDefaults();
     RawContestConfig rawConfig = config.getRawConfig();
 
-    OutputSettings outputSettings = rawConfig.outputSettings;
```
tarheel: What motivated you to change everything in this method to read from the ContestConfig instead of the RawContestConfig? @HEdingfield and I debated this a while ago. To reiterate the point I made at the time: the raw config is what we're saving, so shouldn't it be what we use to populate the form fields? This ensures that any normalizing that happens in ContestConfig (e.g. the special logic for maxRankingsAllowed) doesn't corrupt the actual config file.

But perhaps I'm missing the reason that your change to how we handle strings necessitates these changes as well.

tarheel:
Oh, also, I'm still wondering why we're now using the empty string for most string fields, but we're setting candidate codes explicitly to null.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -640,7 +641,7 @@ private void setDefaultValues() {
         String.valueOf(ContestConfig.SUGGESTED_MINIMUM_VOTE_THRESHOLD));
   }
 
-  private void clearConfig() {
+  private void clearWidgetsAndSetToDefaults() {
```
HEdingfield: Can we change this to `clearGuiAndSetToDefaults()`?

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -613,7 +613,8 @@ private void setTextFieldToInteger(TextField textField, Integer value) {
     textField.setText(value != null ? Integer.toString(value) : "");
   }
 
-  private void setDefaultValues() {
+  // populate widgets with suggested default values for settings which allow them
+  private void setWidgetDefaultValues() {
```
HEdingfield: Please change to `setGuiDefaultValues()`.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -769,21 +770,22 @@ private boolean checkForSaveAndTabulate() {
...
+  private void loadConfigIntoGui(ContestConfig config) {
+    clearWidgetsAndSetToDefaults();
     RawContestConfig rawConfig = config.getRawConfig();
 
-    OutputSettings outputSettings = rawConfig.outputSettings;
-    textFieldContestName.setText(outputSettings.contestName);
+    textFieldContestName.setText(config.getContestName());
     textFieldOutputDirectory.setText(config.getOutputDirectoryRaw());
-    if (outputSettings.contestDate != null && !outputSettings.contestDate.isEmpty()) {
+    if (!config.getContestDate().isEmpty()) {
```
HEdingfield: Make this same change for the other fields you changed to use `getString()`, i.e. getting rid of the != null check. I see other scattered instances throughout the code where this hasn't been done yet.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -839,10 +841,13 @@ private String getChoiceElse(ChoiceBox choiceBox, Enum defaultValue) {
     return choiceBox.getValue() != null ? choiceBox.getValue().toString() : defaultValue.toString();
   }
 
+  // creates a new rawContestConfig object from the contents of the UI widgets
+  // used when creating a new config and to see if we need to save the currently loaded config
   private RawContestConfig createRawContestConfig() {
     RawContestConfig config = new RawContestConfig();
 
     OutputSettings outputSettings = new OutputSettings();
+    // textFields default to returning "" when no text has been set which is expected
```
HEdingfield: Unnecessary comment.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Yeahhh, shit, ughhh, I dunnooooo........

moldover: There needs to be logic somewhere to apply the default "" to a null string parameter.  I followed the logic of keep raw raw, and put it in contestConfig which is better than in the UI code.

src/main/java/network/brightspots/rcv/GuiConfigController.java
moldover: How about setUiDefaultValues?

src/main/java/network/brightspots/rcv/GuiConfigController.java
tarheel: My concern is that if we do it this way, we can't have any special normalization logic in ContestConfig, because every field always has to match the raw version. We already see an example of the problem this creates with maxRankingsAllowed below.

But I don't feel that strongly about it. If you keep it this way, please consider my suggestion for how to change maxRankingsAllowed.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: I guess? Technically the flags we feed into the CLI count as "UI" too.

src/main/java/network/brightspots/rcv/GuiConfigController.java
moldover: Gotcha.  I think we can put the "default" logic into RawConfig and keep the normalization in ContestConfig.  I'll try that and see how it looks.

src/main/java/network/brightspots/rcv/GuiConfigController.java
moldover: >    I believe this is because of the isCdf() function recently added to RawContestConfig; I think we should move that to ContestConfig instead

Yeah, that sounds right.  

>    code should be "" instead of null (may require an additional check in the main tabulation logic to know to ignore "" strings for codes)
>    provider should also be "" instead of null
>    Then you should also be able to replace any instances of != null for these two throughout the code to call the get directly, like you did for contestDate

I didn't get these (wtf???) but I suspect if I move the "" default logic into RawContestConfig will be fixed.

>    Fix test2013MinneapolisMayorScale()
Nooo???

>    In future versions, we might change what "null" means for these fields. I prefer that they're made explicit. I'll bitch about it in #211.

Ok, will continue there.

HEdingfield:
Superseded by #328.

PR #336, opened 2019-07-05
Config versioning

moldover:
Add tabulatorVersion to config file as a required parameter for #43 .  Gui will set this to the current app version on loading.  Validation will fail if this is anything other than the current app version.


src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -218,6 +219,21 @@ private void checkStringToIntWithBoundaries(
     }
   }
 
+  // version validation and migration logic goes here
+  // e.g. unsupported versions would fail or be migrated
+  // in this release we support only the current app version
+  private void validateTabulatorVersion() {
+    if (isNullOrBlank(getTabulatorVersion())) {
+      isValid = false;
+      Logger.log(Level.SEVERE, "Tabulator version is required!");
```
```suggestion
      Logger.log(Level.SEVERE, "tabulatorVersion is required!");
```
Would you also mind changing the "Contest name" message to `contestName` while you're at it? These are the only two fields where the messages don't use the property names.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -218,6 +219,21 @@ private void checkStringToIntWithBoundaries(
...
+    } else {
+      if (!getTabulatorVersion().equals(Main.APP_VERSION)) {
+        isValid = false;
+        Logger.log(Level.SEVERE, "Tabulator version %s not supported!", getTabulatorVersion());
```
```suggestion
        Logger.log(Level.SEVERE, "tabulatorVersion %s is not supported!", getTabulatorVersion());
```

HEdingfield: Can we offer some recourse or advice here? CLI users who manually build their config files are kind of screwed...

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -246,8 +246,6 @@ private void saveFile(File fileToSave) {
     FileUtils.setUserDirectory(fileToSave.getParent());
     // create a rawConfig object from GUI content and serialize it as json
     JsonParser.writeToFile(fileToSave, createRawContestConfig());
-    // Reload to keep GUI fields updated in case invalid values are replaced during save process
-    loadFile(fileToSave, true);
```
HEdingfield: This shouldn't cause double log messages anymore like you mentioned in #43, because I added the silent mode (hence the extra `true` boolean). Please re-add it, because there are some subtle things it can correct for, just to make sure the GUI is fully synced with what's saved to the HDD.

src/main/java/network/brightspots/rcv/ContestConfig.java
tarheel: Should we say that the config file must set tabulatorVersion to %d?

src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -31,6 +31,8 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class RawContestConfig {
 
+  // tabulator version which created this file
```
tarheel:
```suggestion
  // tabulator version that created this file
```

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
```diff
@@ -1,6 +1,12 @@
 This document lists the parameters that can be included in a config file suitable for input tabulation.
 Config file must be valid JSON format. Examples can be found in the test_data folder.
 
+  "tabulatorVersion" required
+    version of the application which created this file
```
tarheel:
```suggestion
    version of the application that created this file
```

src/main/java/network/brightspots/rcv/ContestConfig.java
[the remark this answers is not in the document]
tarheel: I'd like to reintroduce my earlier proposal: let's be OK with a few PRs that break the tests, just for like the next 24 hours. There are a number of changes in flight or coming soon that require updating every single test in some tedious way. Of course each PR needs to be tested enough to feel confident about the changes, but we can save some time if we don't make each PR update _all_ of the tests. And then we can have a final PR that fixes all of them at once.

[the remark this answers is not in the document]
moldover: Haha yea I test brah :)  Plan is to update all configs to include version 1.0.0 along with randomSeed etc. as Louis suggests.  Sorry, I should have indicated that in the PR

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -772,7 +770,12 @@ public LocalDate fromString(String string) {
   private void loadConfig(ContestConfig config) {
     clearConfig();
     RawContestConfig rawConfig = config.getRawConfig();
-
+    if (config.rawConfig.tabulatorVersion != Main.APP_VERSION) {
```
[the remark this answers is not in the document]
moldover: [face palm]

src/main/java/network/brightspots/rcv/GuiConfigController.java
moldover: Tried this and silent mode does indeed work.

src/main/java/network/brightspots/rcv/ContestConfig.java
moldover: yeah sounds good.

PR #337, opened 2019-07-06
add residual surplus to overall summary JSON

### LINKED ISSUE #237: include residual surplus in summary JSON
We don't actually want to visualizer to display it, but having it in the JSON will allow us to add a proper test for it.

tarheel:
Fixes #237.

Some of the changes in this PR are just automated style updates for code that I didn't touch. I have no idea why our copies of IntelliJ seem to disagree about these things. :(

### DIFF
--- a/src/main/java/network/brightspots/rcv/ContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/ContestConfig.java
@@ -154,6 +154,23 @@ class ContestConfig {
     return source.getProvider() != null && source.getProvider().toUpperCase().equals(CDF_PROVIDER);
   }
 
+  private static boolean candidateStringMatchesLabel(
+      String candidateString, String field, String label, String labelField) {
+    boolean match = false;
+
+    if (!isNullOrBlank(label) && label.equals(candidateString)) {
+      match = true;
+      Logger.log(
+          Level.SEVERE,
+          "\"%s\" can't be used as a candidate %s if it's also being used as the %s!",
+          candidateString,
+          field,
+          labelField);
+    }
+
+    return match;
+  }
+
   // function: resolveConfigPath
   // purpose: given a path returns absolute path for use in File IO
   // param: path from this config file (cvr or output folder)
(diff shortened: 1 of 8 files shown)


src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -147,7 +147,7 @@ private static void runTabulationTest(String stem) {
       for (File file : outputFolder.listFiles()) {
         if (!file.isDirectory()) {
           //noinspection ResultOfMethodCallIgnored
-          file.delete();
+          //file.delete();
```
HEdingfield: Undo this?

src/main/java/network/brightspots/rcv/TallyTransfers.java
```diff
@@ -24,20 +24,40 @@
...
 import java.util.Map;
 
 // TallyTransfers class stores summary info on vote transfers
-// used primary as visualizer input to help build Sankey plots
+// used primarily as visualizer input to help build Sankey plots
 class TallyTransfers {
 
+  static final String EXHAUSTED = "exhausted";
+  static final String RESIDUAL_TARGET = "residual surplus";
+  static final String UNCOUNTED = "uncounted";
```
HEdingfield: Should we include a test with "uncounted"?

src/main/java/network/brightspots/rcv/TallyTransfers.java
```diff
@@ -55,11 +75,11 @@
   void addTransfer(int round, String sourceCandidate, String targetCandidate, BigDecimal value) {
     // null source means we are transferring the initial count
     if (sourceCandidate == null) {
-      sourceCandidate = "uncounted";
+      sourceCandidate = UNCOUNTED;
```
HEdingfield: Minor alarm bells going off with setting directly equal to a declared string. It points to the object rather than the value. But I guess that's ok here?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -335,6 +334,18 @@ private void validateCandidates() {
         isValid = false;
         Logger.log(
             Level.SEVERE, "Duplicate candidate names are not allowed: %s", candidate.getName());
+      } else if (TallyTransfers.candidateStringIsReserved(candidate.getName())) {
```
HEdingfield: Cool check. Dumb question, but we don't want/need to do something similar for other key words like "UWI", "overvote", etc. right?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -76,6 +60,21 @@
   static final int SUGGESTED_DECIMAL_PLACES_FOR_VOTE_ARITHMETIC = 4;
   static final BigDecimal SUGGESTED_MINIMUM_VOTE_THRESHOLD = BigDecimal.ZERO;
   static final int SUGGESTED_MAX_SKIPPED_RANKS_ALLOWED = 1;
+  private static final int MIN_COLUMN_INDEX = 1;
```
HEdingfield: I hate this change and fought tooth and nail against the code reorganization during commit to make sure it didn't happen. Just ended up unchecking it in the commit dialog.

src/main/java/network/brightspots/rcv/TallyTransfers.java
tarheel: I tested to confirm that this isn't an issue. This line makes `sourceCandidate` a pointer to the value, not to `UNCOUNTED`. Even if `UNCOUNTED` were to change, `sourceCandidate` would remain the same.

src/main/java/network/brightspots/rcv/ContestConfig.java
tarheel: Hmm. I don't want to have to remember not to let the code reorganization run every time I edit this file. Can we just live with the way that it wants to do it?

src/main/java/network/brightspots/rcv/ContestConfig.java
tarheel: Not a dumb question. We should do this, actually.

src/main/java/network/brightspots/rcv/TallyTransfers.java
tarheel: "uncounted" never actually appears in the output

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: Ok ok, just make sure the comment about false default booleans is in a place that makes sense. There's a small chance that me changing this to a String might resolve the issue, but won't know until I work on it more on Sunday.

tarheel:
OK, it's ready for review again.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -323,38 +340,50 @@ private void validateCvrFileSources() {
     }
   }
 
+  private boolean candidateStringIsAlreadyInUseElsewhere(
+      String candidateString, String field, Set<String> candidateStringsSeen) {
+    boolean foundError = false;
+
+    if (candidateStringsSeen.contains(candidateString)) {
+      isValid = false;
```
HEdingfield: This should be `foundError = true;` since you already set `isValid` to `false` after calling this function both times.

src/main/java/network/brightspots/rcv/TallyTransfers.java
```diff
@@ -24,20 +24,40 @@
...
+  static final String UNCOUNTED = "uncounted";
+  static final String[] RESERVED_STRINGS = {EXHAUSTED, RESIDUAL_TARGET, UNCOUNTED};
+
   // Map of round number to vote transfers which occurred in that round
   // transfers for a round are a map of SOURCE candidate(s) to one or more TARGET candidates.
   // For each target candidate the map value is total vote values received from that source.
   // For round 1 source candidate is marked "uncounted" since the votes had no prior recipient.
   private final Map<Integer, Map<String, Map<String, BigDecimal>>> tallyTransfers = new HashMap<>();
 
+  static boolean candidateStringIsReserved(String candidateString) {
```
HEdingfield: This seems more like it's exclusively a validation function at this point; should it live in `ContestConfig`?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -324,26 +340,50 @@ private void validateCvrFileSources() {
     }
   }
 
+  private boolean candidateStringIsAlreadyInUseElsewhere(
```
HEdingfield: Please add a line or two comment to describe what this does. The parameter names are a bit confusing.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -155,6 +154,23 @@ static boolean isCdf(CVRSource source) {
     return source.getProvider() != null && source.getProvider().toUpperCase().equals(CDF_PROVIDER);
   }
 
+  private static boolean candidateStringMatchesLabel(
```
HEdingfield: Please add a line or two comment to describe what this does. The parameter names are a bit confusing.

PR #342, opened 2019-07-06
Better label and reserved string validation

HEdingfield:
Simplifies `TallyTransfers.RESERVED_STRINGS` check. Checks label fields against each other and `RESERVED_STRINGS`. Needs a bit more work (hence the TODOs), but please review and/or finish off if you can.

### DIFF
--- a/src/main/java/network/brightspots/rcv/ContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/ContestConfig.java
@@ -179,19 +179,6 @@ class ContestConfig {
     return match;
   }
 
-  private static boolean candidateStringIsReservedForTallyTransfers(String candidateString) {
-    boolean found = false;
-    if (!isNullOrBlank(candidateString)) {
-      for (String s : TallyTransfers.RESERVED_STRINGS) {
-        if (s.equalsIgnoreCase(candidateString)) {
-          found = true;
-          break;
-        }
-      }
-    }
-    return found;
-  }
-
   // function: resolveConfigPath
   // purpose: given a path returns absolute path for use in File IO
   // param: path from this config file (cvr or output folder)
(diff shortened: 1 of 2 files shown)


src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -361,36 +347,52 @@ private void validateCvrFileSources() {
...
-          candidateString,
-          field);
-    } else if (candidateStringMatchesLabel(
-        candidateString, field, getUndeclaredWriteInLabel(), "undeclaredWriteInLabel")
-        || candidateStringMatchesLabel(candidateString, field, getOvervoteLabel(), "overvoteLabel")
-        || candidateStringMatchesLabel(
-        candidateString, field, getUndervoteLabel(), "undervoteLabel")) {
-      foundError = true;
+    } else {
+      inUse = stringAlreadyInUseElsewhere(candidateString, "candidate " + field);
```
tarheel: I think this should be "a candidate " instead of "candidate "

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -568,6 +570,20 @@ private void validateRules() {
       isValid = false;
       Logger.log(Level.SEVERE, "randomSeed can't be less than %d!", MIN_RANDOM_SEED);
     }
+
+    // TODO: Fix so they don't check each other
```
tarheel: Right, because currently you'll get duplicate errors. Not immediately obvious to me how to solve this in an elegant way. Oh, how about each of these fields only checks other fields that follow it in a lexical sorting?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -839,6 +855,7 @@ boolean isExhaustOnDuplicateCandidateEnabled() {
   // param: code the code of the candidate whose name we want to look up
   // returns: the full name for the given candidateID
   String getNameForCandidateCode(String code) {
+    // TODO pull "Undeclared" into constant and add to reserved terms check?
```
tarheel: yes, please

src/main/java/network/brightspots/rcv/TallyTransfers.java
```diff
@@ -35,7 +36,7 @@
   static final String RESIDUAL_TARGET = "residual surplus";
   private static final String EXHAUSTED = "exhausted";
   private static final String UNCOUNTED = "uncounted";
-  static final String[] RESERVED_STRINGS = {EXHAUSTED, RESIDUAL_TARGET, UNCOUNTED};
+  static final Set<String> RESERVED_STRINGS = Set.of(RESIDUAL_TARGET, EXHAUSTED, UNCOUNTED);
```
tarheel: oh, duh

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: Filed #352 to take care of this in the future, since it's a relatively minor issue.

Issue #338, opened 2019-07-06
Audit logging bug after cancelling tiebreaker


HEdingfield:
If you tabulate using the interactive tiebreaker, e.g. with `sample_interactive_tiebreak`, and then cancel during the tiebreak, the original audit log continues to grow indefinitely (recording any and all subsequent tabulations) until you close the tabulator.

HEdingfield:
Closed with #343.

Issue #299, opened 2019-06-23
"Cleaning up unclosed ZipFile" message


tarheel:
Doesn't seem like it's breaking anything, but this has started showing up in the console output. Maybe the XLSX reader needs to close the file when it's done?

HEdingfield:
Closed with #343.

PR #343, opened 2019-07-06
cancel tiebreak handling - fix for #338

moldover:
Fix for #338.  I also removed logging of 'user' during startup, seems kinda weird and no one has any strong feelings on it.

Just added a fix for #299

### DIFF
--- a/src/main/java/network/brightspots/rcv/TabulatorSession.java
+++ b/src/main/java/network/brightspots/rcv/TabulatorSession.java
@@ -114,7 +114,7 @@ class TabulatorSession {
   void tabulate() {
     Logger.log(Level.INFO, "Starting tabulation session...");
     ContestConfig config = ContestConfig.loadContestConfig(configPath);
-    Boolean tabulationSuccess = false;
+    boolean tabulationSuccess = false;
     if (config != null && config.validate() && setUpLogging(config)) {
       try {
         Logger.log(Level.INFO, "Begin config file contents:");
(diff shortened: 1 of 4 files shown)


src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -111,9 +111,10 @@ void convertToCdf() {
   // function: tabulate
   // purpose: run tabulation
   // returns: list of winners
-  void tabulate() throws TabulationCancelledException {
+  void tabulate() {
     Logger.log(Level.INFO, "Starting tabulation session...");
     ContestConfig config = ContestConfig.loadContestConfig(configPath);
+    Boolean tabulationSuccess = false;
```
tarheel: boolean

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -143,7 +144,13 @@ void tabulate() throws TabulationCancelledException {
             Logger.log(Level.SEVERE, "Aborting tabulation due to cast vote record errors!");
             break;
           }
-          Set<String> newWinnerSet = runTabulationForConfig(config, castVoteRecords);
+          Set<String> newWinnerSet;
+          try {
+            newWinnerSet = runTabulationForConfig(config, castVoteRecords);
+          } catch (TabulationCancelledException e) {
+              Logger.log(Level.SEVERE, "Tabulation was cancelled by the user!");
```
tarheel: indentation is off

src/main/java/network/brightspots/rcv/TabulatorSession.java
moldover: Why this?  I remember we decided this at some point, but I think we should at least consider using Booleans in RawContestConfig.  It allows us to distinguish between missing (null value, or no key) vs. false values in Jackson.

moldover:
Made edits.  @tarheel what logic are you talking about?  all i did was add a flag which controls a log message.

src/main/java/network/brightspots/rcv/TabulatorSession.java
tarheel: Sure, I'm open to Booleans (and other object equivalents for primitives) in RawContestConfig if we don't decide to use all Strings.

But in a case like this, there's no reason not to use a primitive; all you want to store is false or true.

tarheel:
> what logic are you talking about? all i did was add a flag which controls a log message.

You changed the level in the stack where the exceptions are caught, and that's not a part of the code that I've thought about before.

PR #349, opened 2019-07-07
update invalid_params_test config file

### LINKED ISSUE #279: GUI falsely reports that invalid config is unsaved
Repro:
- Load invalid_params_test_config.json
- Click the Load button again. It will ask if you want to save your changes, even though you didn't change anything.

tarheel:
To address #279 (see my latest comment there).


HEdingfield:
We shouldn't replace this file because it "corrects" some of the invalid params, which is what we're testing in the first place.

tarheel:
Oh, OK, let me look at it again. There's still an issue here, though, because when you load it and don't modify anything, it asks you to save... which is not ideal.

PR #350, opened 2019-07-07
fix minor output bug and add test case for previousRoundCountsThenRandom

### LINKED ISSUE #341: add automated test for previousRoundCountsThenRandom tie-breaking mode

tarheel:
Fixes #341.


src/main/java/network/brightspots/rcv/TieBreak.java
```diff
@@ -118,15 +118,15 @@ String nonSelectedCandidateDescription() {
       nonselected = options.get(0);
     } else if (options.size() == 2) {
       // if there are only 2 candidates don't use a comma
-      nonselected = options.get(0) + " and " + options.get(1);
+      nonselected = '"' + options.get(0) + "\" and \"" + options.get(1) + '"';
```
HEdingfield: It's a little weird to inconsistently render quotes like this in our messages. I recommend consistently sticking with `\"` instead of `'"'`, since that's what I've been doing elsewhere. If you change here, please change elsewhere too.

PR #354, opened 2019-07-07
Misc cleanup

HEdingfield:
Resizes GUI to fit new rules. Flips ternaries that start with negative to make them more readable. Text nit cleanup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,7 @@ build
 out
 tmp
 
-# Excel temp files
+### Excel temp files
 ~$*.xlsx
 
 ### JetBrains IntelliJ exclusions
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -50,7 +50,7 @@ tmp
 
 # Gradle and Maven with auto-import
 # When using Gradle or Maven with auto-import, you should exclude module files,
-# since they will be recreated, and may cause churn.  Uncomment if using
+# since they will be recreated, and may cause churn. Uncomment if using
```
tarheel:
```suggestion
# since they will be re-created, and may cause churn. Uncomment if using
```

src/main/java/network/brightspots/rcv/Logger.java
```diff
@@ -123,7 +123,7 @@ static void addTabulationFileLogging(String outputPath) throws IOException {
   }
 
   // function: removeTabulationFileLogging
-  // purpose: remove file logging once a tabulation run is complete
+  // purpose: remove file logging once a tabulation run is completed
```
tarheel: debatable!

HEdingfield: Hah, throw another one up on the board! In my view, "complete" is the state of something being full, whereas completed means a task was literally finished, which is the context here.

PR #356, opened 2019-07-08
Random documentation

moldover:
Took a stab at this for #170.   This should enable reproducibility of our results as Neal describes.


src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -950,6 +950,9 @@ private void processCandidateData() {
       // catch that and report a helpful error. Validation also hits this code path, though, so we
       // need to prevent a NullPointerException here.
       if (getRandomSeed() != null) {
+	    // sort candidate permutation for reproducibility
+	    Collections.sort(candidatePermutation);
```
tarheel: This seems reasonable. Is it actually possible for this to be in a different order, though?

Also, indentation is off.

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
```diff
@@ -113,10 +113,60 @@ Config file must be valid JSON format. Examples can be found in the test_data fo
...
+        the integer seed for the application's pseudorandom number generator
+        value: [0..2147483647]
+        if not supplied: null
+
       "tiebreakMode" required
-        how the program should decide which candidate to eliminate when multiple candidates are tied for last place among continuing candidates in a given round
+        how the program should decide which candidate to eliminate when multiple candidates are tied for last place
+        or which candidate to elect first when:
+          electionWinnerMode is set to multiSeatAllowOnlyOneWinnerPerRound and
+          multiple candidates exceed the winning threshold and
```
tarheel:
```suggestion
          2) multiple candidates exceed the winning threshold in the same round, and
```

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
```diff
@@ -113,10 +113,60 @@ Config file must be valid JSON format. Examples can be found in the test_data fo
...
+      "randomSeed" required when tiebreakMode is random, previousRoundCountsThenRandom, or generatePermutation
+        the integer seed for the application's pseudorandom number generator
+        value: [0..2147483647]
+        if not supplied: null
+
       "tiebreakMode" required
-        how the program should decide which candidate to eliminate when multiple candidates are tied for last place among continuing candidates in a given round
+        how the program should decide which candidate to eliminate when multiple candidates are tied for last place
+        or which candidate to elect first when:
+          electionWinnerMode is set to multiSeatAllowOnlyOneWinnerPerRound and
```
tarheel:
```suggestion
          1) electionWinnerMode is set to multiSeatAllowOnlyOneWinnerPerRound, and
```

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
```diff
@@ -113,10 +113,60 @@ Config file must be valid JSON format. Examples can be found in the test_data fo
...
+          multiple candidates exceed the winning threshold and
+          also tie for the highest vote total in the same round
```
tarheel:
```suggestion
          3) at least two of thoes candidates are tied for the highest vote total in that round 
```

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
```diff
@@ -113,10 +113,60 @@ Config file must be valid JSON format. Examples can be found in the test_data fo
...
+            int randomCandidateIndex = (int) Math.floor(randomDouble * (double) tiedCandidates.size());
+          the selected candidate will be the winner or loser for that round
+
+        "interactive":
+          the user is presented with a list of tied candidates
+          the user will input their selection manually
+
+        "previousRoundCountsThenRandom"
+          the tied candidate with the highest vote total in the previous round is selected
+          if there is a tie for the vote count in the previous round as well, a candidate is selected from the still tying candidates as described under tiebreakMode "random"
```
tarheel: I actually haven't implemented this part properly. If it has to fall back to random or interactive, it picks from the full set of tied candidates, not just the ones who were still alive at the end of the previousRoundCounts logic. I'll send a PR to fix this shortly.

src/main/java/network/brightspots/rcv/ContestConfig.java
moldover: In the current code, i don't think so.  But this hook helps us define the input to shuffle() in case other processing changes at some point.

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
```diff
@@ -113,10 +113,59 @@ Config file must be valid JSON format. Examples can be found in the test_data fo
...
+      "randomSeed" required when tiebreakMode is random, previousRoundCountsThenRandom, or generatePermutation
+        the integer seed for the application's pseudorandom number generator
+        value: [0..2147483647]
+        if not supplied: null
+
       "tiebreakMode" required
-        how the program should decide which candidate to eliminate when multiple candidates are tied for last place among continuing candidates in a given round
+        how the program should decide which candidate to eliminate when multiple candidates are tied for last place
+        or which candidate to elect first when:
+          1) electionWinnerMode is set to multiSeatAllowOnlyOneWinnerPerRound, and
```
nealmcb: Would this also apply if it is the final round even if multiSeatAllowOnlyOneWinnerPerRound is not the current mode?

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
```diff
@@ -113,10 +113,59 @@ Config file must be valid JSON format. Examples can be found in the test_data fo
...
+          2) multiple candidates exceed the winning threshold in the same round, and
+          3) at least two of those candidates are tied for the highest vote total in that round
         value: "random" | "interactive" | "previousRoundCountsThenRandom" | "previousRoundCountsThenInteractive" | "usePermutationInConfig" | "generatePermutation"
 
+        we use Java 12 java.util.random for randomness in our tiebreak implementations
```
nealmcb: Saying "12" makes it seem specific to that version, which we all dearly hope it will never be.
This phrasing hopefully makes it seem more stable:
> "we use java.util.random for randomness in our tiebreak implementations. Compatible
 methods exist for other languages, e.g. https://pypi.org/project/java-random/

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
```diff
@@ -113,10 +113,59 @@ Config file must be valid JSON format. Examples can be found in the test_data fo
     set of configuration parameters that specify the tabulation rules
     the "rules" section contains the following parameters:
 
+      "randomSeed" required when tiebreakMode is random, previousRoundCountsThenRandom, or generatePermutation
+        the integer seed for the application's pseudorandom number generator
+        value: [0..2147483647]
```
nealmcb: It would be more robust and safer to allow larger seeds, ala the 48 bits actually used by Java, or the 20 decimal digits typical for election auditing approaches.

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
tarheel: When multiSeatAllowOnlyOneWinnerPerRound isn't enabled and multiple candidates exceed the threshold in the same round, we consider them to be elected simultaneously. Their surplus redistributions happen after they've been all been elected.

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
tarheel: Sure, we can do this.

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
moldover: The .jar file we create is compiled against Java 12, so this statement is accurate.  If one wanted to reproduce our results you would need to do the same, no?  It's not to say we're somehow tied to this implementation in future releases.

moldover:
@nealmcb would you be interested in working on this code?  We could make you a collaborator!

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
nealmcb: You bring up an important topic: what sort of reproducability are we trying to achieve?
As I just noted at https://github.com/BrightSpots/rcv/issues/43#issuecomment-509695201
I think the documentation, code and comments should all help clarify and implement the goal of defining and producing a stable standard definition of everything necessary for folks who want to write compatible software that comes up with exactly the same results as this VVSG reference implementation.

So I'm looking for a more stable definition of the PRNG in use than saying it is what Java ships in version 12. I think it has actually be stable for a long time and will remain stable, and we should frame comments to reflect that, and to help those who are writing compatible software in other languages.

HEdingfield: @nealmcb changed to your wording in PR #370.

moldover: Gotcha.  That is the notion of reproducibility I tried to facilitate in this PR so I think we are on the same page.  Is there a library you have in mind?  Practically speaking, for this release we are not going to include any code changes like this, but we could further enhance the documentation, and potentially upgrade our implementation in the future.

Issue #332, opened 2019-07-05
Bad string for contestDate causes silent error during load


HEdingfield:
If you enter a bad string in contestDate JSON and try to load it in the GUI, it says it loaded successfully in the GUI console even through it throws a DateTimeParseException in IntelliJ.

1) Need to fix handling of bad date.
2) Need to fix silent errors like this occurring; I thought this was resolved by #174, but it can apparently still happen in certain cases.

moldover:
@HEdingfield you ok with closing this?  If you still want to have a ticket around silent exceptions in general I'd propose to edit this one.

HEdingfield:
Thanks! I'll keep it open in case we want to deal with part 2, but I'll remove the tags.

PR #357, opened 2019-07-08
fix for #332

moldover:
This fix handles bad date strings.  

The second idea in the ticket to more broadly handle all "silent" exceptions is tricky.  The only way I can see to do that is add generic exception handling on all ui callback code.  I don't think it's worth doing, certainly not for this release.

### DIFF
--- a/src/main/java/network/brightspots/rcv/GuiConfigController.java
+++ b/src/main/java/network/brightspots/rcv/GuiConfigController.java
@@ -794,7 +794,7 @@ public class GuiConfigController implements Initializable {
         datePickerContestDate.setValue(
             LocalDate.parse(outputSettings.contestDate, DATE_TIME_FORMATTER));
       } catch (DateTimeParseException exception) {
-        Logger.log(Level.SEVERE, "invalid contestDate: %s", outputSettings.contestDate);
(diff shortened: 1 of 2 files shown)


src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -789,8 +790,13 @@ private void loadConfig(ContestConfig config) {
     textFieldContestName.setText(outputSettings.contestName);
     textFieldOutputDirectory.setText(config.getOutputDirectoryRaw());
     if (!isNullOrBlank(outputSettings.contestDate)) {
-      datePickerContestDate.setValue(
-          LocalDate.parse(outputSettings.contestDate, DATE_TIME_FORMATTER));
+      try {
+        datePickerContestDate.setValue(
+            LocalDate.parse(outputSettings.contestDate, DATE_TIME_FORMATTER));
+      } catch (DateTimeParseException exception) {
+        Logger.log(Level.SEVERE, "invalid contestDate: %s", outputSettings.contestDate);
```
tarheel:
```suggestion
        Logger.log(Level.SEVERE, "Invalid contestDate: %s!", outputSettings.contestDate);
```

PR #358, opened 2019-07-08
fix fallback logic for previous round counts tie-breaking modes

tarheel:
I only noticed this yesterday! Oops. Before this PR, the code has been handling the fallback part of previousRoundCountsThen[Random|Interactive] incorrectly: instead of only considering the candidates who were still tied at the end of the round counts comparison, it was reverting back to the full set of tied candidates.

### DIFF
--- a/src/main/java/network/brightspots/rcv/TieBreak.java
+++ b/src/main/java/network/brightspots/rcv/TieBreak.java
@@ -46,13 +46,14 @@ import javafx.scene.Scene;
 import javafx.stage.Modality;
 import javafx.stage.Stage;
 import network.brightspots.rcv.Tabulator.TabulationCancelledException;
+import network.brightspots.rcv.Tabulator.TieBreakMode;
 
 class TieBreak {
 
   private static Random random;
 
-  // list of candidates who are tied
-  private final List<String> tiedCandidates;
+  // list of all candidates who are tied for this tie-breaker
+  private final List<String> allTiedCandidates;
   // mode we're using for breaking ties
   private final Tabulator.TieBreakMode tieBreakMode;
   // ordering to use if we're doing permutation-based tie-breaking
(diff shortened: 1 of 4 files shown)


src/main/java/network/brightspots/rcv/TieBreak.java
```diff
@@ -312,42 +293,63 @@ private String doRandom() {
...
+      String explanationPrefix =
+          String.format(
+              "Comparing previous round counts still resulted in a tie %s %s, so we fell back to %s.",
+              candidatesInContention.size() > 2 ? "among" : "between",
+              Utils.listToSentenceWithQuotes(candidatesInContention),
+              tieBreakMode == TieBreakMode.PREVIOUS_ROUND_COUNTS_THEN_RANDOM
+                  ? "random"
+                  : "interactive");
+      if (tieBreakMode == TieBreakMode.PREVIOUS_ROUND_COUNTS_THEN_INTERACTIVE) {
+        selection = doInteractive(candidatesInContention);
```
moldover: I hope someday, somewhere, somehow, this line of code is executed in a non-test tabulation and gets its wings.

src/main/java/network/brightspots/rcv/Utils.java
```diff
@@ -16,9 +16,31 @@
...
 
+import java.util.List;
+
 class Utils {
 
   static boolean isNullOrBlank(String s) {
     return s == null || s.isBlank();
   }
+
+  static String listToSentenceWithQuotes(List<String> list) {
```
moldover: not really a "sentence" but the attention to detail here more than makes up for it

tarheel: Good point... but other languages call this a "sentence" for some reason, and who am I to think that I'm better than they are?

https://apidock.com/rails/Array/to_sentence

src/main/java/network/brightspots/rcv/TieBreak.java
tarheel: I'll take the under.

PR #359, opened 2019-07-08
Stringify config

### LINKED ISSUE #139: Improve TextFieldListenerNonNegInt for config creator GUI
`TextFieldListenerNonNegInt` isn't the best UX, since it just prevents the entering of invalid characters without giving any other feedback. Ideally we'd have a pop-up note that says "integers only" or something.

### LINKED ISSUE #211: Safeguards against changing "if not supplied" behavior in future versions

### LINKED ISSUE #325: firstVoteColumnIndex and firstVoteRowIndex issues for JSON CVRs
The GUI allows you to add a CVR to the list without first vote row and first vote column if the filename contains `.json`. However, it doesn't run this check again if you double click the fields in the list and modify them, thus forcing you to keep a value in these fields.

Also encountering this when I try to run validation:
```
2019-07-04 09:35:52 PDT SEVERE: Cast vote record file not found: C:\Users\Me\IdeaProjects\rcv\test.json
2019-07-04 09:35:52 PDT SEVERE: firstVoteColumnIndex is required: C:\Users\Me\IdeaProjects\rcv\test.json
2019-07-04 09:35:52 PDT SEVERE: firstVoteRowIndex is required: C:\Users\Me\IdeaProjects\rcv\test.json
```
Not sure if these latter two notes about the indices being required would disappear if it were actually able to load the .json file? If not, we may have another problem here.

### LINKED ISSUE #345: inaccurate "Migrated tabulator version from 0.1.0 to 0.1.0" message
If you open a config file that has no version number, you get this message. Instead, it should probably say that it's migrating from unversioned...?

HEdingfield:
Fixes #211. Cleaned up migrateConfigVersion() (fixes #345); enum choiceBox elements in GUI now load as blank when bad values are present in the config.

Set up suggested outputDirectory of "output".

Changes RawConfig firstVoteColumnIndex and firstVoteRowIndex fields to Strings; tightens up editing of CVR sources in GUI (fixes #325); ContestConfig.isCdf also checks file extension now.

Converted RawContestConfig fields randomSeed, numberOfWinners, decimalPlacesForVoteArithmetic, minimumVoteThreshold to Strings; refactored and cleaned up validation logic; got rid of TextFieldListenerNonNegInt (fixes #139).

### DIFF
--- a/src/main/java/network/brightspots/rcv/ContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/ContestConfig.java
@@ -462,7 +462,7 @@ class ContestConfig {
     }
   }
 
-  private boolean checkIfInt(String s) {
+  private boolean isInt(String s) {
     boolean isInt = true;
     try {
       Integer.parseInt(s);
@@ -545,7 +545,7 @@ class ContestConfig {
         MIN_MINIMUM_VOTE_THRESHOLD, MAX_MINIMUM_VOTE_THRESHOLD, true);
 
     // If this is a multi-seat contest, we validate a couple extra parameters.
-    if (!isNullOrBlank(getNumberOfWinnersRaw()) && checkIfInt(getNumberOfWinnersRaw())
+    if (!isNullOrBlank(getNumberOfWinnersRaw()) && isInt(getNumberOfWinnersRaw())
         && getNumberOfWinners() > 1) {
       if (isSingleSeatContinueUntilTwoCandidatesRemainEnabled()) {
         isValid = false;
@@ -960,7 +960,7 @@ class ContestConfig {
       // It's not valid to have a null random seed with this tie-break mode; the validation will
       // catch that and report a helpful error. Validation also hits this code path, though, so we
       // need to prevent a NullPointerException here.
-      if (!isNullOrBlank(getRandomSeedRaw()) && checkIfInt(getRandomSeedRaw())) {
+      if (!isNullOrBlank(getRandomSeedRaw()) && isInt(getRandomSeedRaw())) {
         Collections.shuffle(candidatePermutation, new Random(getRandomSeed()));
       }
     }
--- a/src/main/java/network/brightspots/rcv/ContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/ContestConfig.java
@@ -226,24 +226,58 @@ class ContestConfig {
     return isValid;
   }
 
+  private void invalidateAndLog(String message, String inputLocation) {
+    isValid = false;
+    message += inputLocation == null ? "!" : ": " + inputLocation;
+    Logger.log(Level.SEVERE, message);
+  }
+
+  // Makes sure String input can be converted to an int, and checks that int against boundaries
+  private void checkStringToIntWithBoundaries(String input, String inputName, Integer lowerBoundary,
+      Integer upperBoundary, boolean isRequired) {
+    checkStringToIntWithBoundaries(input, inputName, lowerBoundary, upperBoundary, isRequired,
+        null);
+  }
+
   // Makes sure String input can be converted to an int, and checks that int against boundaries
   private void checkStringToIntWithBoundaries(
-      String input, String inputName, int lowerBoundary, int upperBoundary, String inputLocation) {
-    try {
-      int stringInt = Integer.parseInt(input);
-      if (stringInt < lowerBoundary || stringInt > upperBoundary) {
-        isValid = false;
-        Logger.log(
-            Level.SEVERE,
-            "%s must be from %d to %d if supplied: %s",
-            inputName,
-            lowerBoundary,
-            upperBoundary,
-            inputLocation);
+      String input, String inputName, Integer lowerBoundary, Integer upperBoundary,
+      boolean isRequired, String inputLocation) {
+    String message = String.format("%s must be", inputName);
+    if (lowerBoundary != null && upperBoundary != null) {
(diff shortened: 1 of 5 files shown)


src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -349,25 +349,29 @@ public void buttonCvrFilePathClicked() {
     }
   }
 
+  private boolean cvrSourceHasRequiredFields(CVRSource source) {
+    boolean hasRequiredFields = true;
+    if (source.getFilePath().isBlank()) {
+      hasRequiredFields = false;
+      Logger.log(Level.WARNING, "filePath is required!");
```
tarheel: here and below: why only WARNING instead of SEVERE?

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -386,50 +390,61 @@ public void buttonDeleteCvrFileClicked() {
...
     CVRSource cvrSelected = tableViewCvrFiles.getSelectionModel().getSelectedItem();
-    String cvrFilePath = cellEditEvent.getNewValue().toString().trim();
-    if (cvrFilePath.isBlank()) {
-      Logger.log(Level.WARNING, "Cast vote record file path is required!");
-    } else {
-      cvrSelected.setFilePath(cvrFilePath);
+    String oldFilePath = cvrSelected.getFilePath();
+    cvrSelected.setFilePath(cellEditEvent.getNewValue().toString().trim());
+    if (!cvrSourceHasRequiredFields(cvrSelected)) {
+      cvrSelected.setFilePath(oldFilePath);
```
tarheel: Could you add a comment somewhere explaining this and the similar logic below? The UI is rejecting any editing of the CVR source that creates an invalid state?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -607,15 +615,19 @@ private void validateRules() {
...
+  private String getNumberOfWinnersRaw() {
+    return rawConfig.rules.numberOfWinners;
+  }
+
   // function: getNumberWinners
   // purpose: how many winners for this contest
   // returns: number of winners
   Integer getNumberOfWinners() {
-    return rawConfig.rules.numberOfWinners;
+    return Integer.parseInt(rawConfig.rules.numberOfWinners);
```
tarheel: So now we're trusting that a callsite will never call a getter like this unless the config has already been validated... right?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -850,10 +868,14 @@ TieBreakMode getTiebreakMode() {
     return mode == null ? TieBreakMode.MODE_UNKNOWN : mode;
   }
 
-  Integer getRandomSeed() {
+  private String getRandomSeedRaw() {
     return rawConfig.rules.randomSeed;
   }
 
+  Integer getRandomSeed() {
+    return Integer.parseInt(rawConfig.rules.randomSeed);
```
tarheel: why don't your `getX` methods reference the corresponding `getXRaw` methods?

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: Yes, that's correct. I checked carefully, but if you want to add some further warnings or checks somewhere, we can file an issue.

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: Good call, switching.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Convention in the GUI. It's not making anything crash or throwing any real errors, it just prevents the add.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Done.
`// Cache the original value to revert back to in case the change fails`

PR #365, opened 2019-07-08
Moves all randomSeed logic inside of Tabulator.tabulate().

HEdingfield:
Related to #170.

### DIFF
--- a/src/main/java/network/brightspots/rcv/Tabulator.java
+++ b/src/main/java/network/brightspots/rcv/Tabulator.java
@@ -131,12 +131,13 @@ class Tabulator {
   // returns: set containing winner(s)
   Set<String> tabulate() throws TabulationCancelledException {
     if (config.needsRandomSeed()) {
+      Random random = new Random(config.getRandomSeed());
       if (config.getTiebreakMode() == TieBreakMode.GENERATE_PERMUTATION) {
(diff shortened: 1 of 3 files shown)


src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -128,12 +130,18 @@
...
   // returns: set containing winner(s)
   Set<String> tabulate() throws TabulationCancelledException {
-    logSummaryInfo();
-
     if (config.needsRandomSeed()) {
-      TieBreak.setRandomSeed(config.getRandomSeed());
+      if (config.getTiebreakMode() == TieBreakMode.GENERATE_PERMUTATION) {
+        // sort candidate permutation for reproducibility
+        Collections.sort(config.getCandidatePermutation());
+        Collections.shuffle(config.getCandidatePermutation(), new Random(config.getRandomSeed()));
```
tarheel: It feels a little weird that we instantiate a second `Random` here. Maybe `Tabulator` should just have this as a member variable and pass it to the constructor of `TieBreak`?

Shouldn't matter in practice, but it feels like it centralizes the random logic a bit more clearly.

HEdingfield: Done.

PR #370, opened 2019-07-09
randomSeed can now be a 48-bit number

HEdingfield:
Addresses comments in #170.

### DIFF
--- a/src/main/java/network/brightspots/rcv/ContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/ContestConfig.java
@@ -243,6 +243,7 @@ class ContestConfig {
       Integer lowerBoundary, Integer upperBoundary, boolean isRequired, String inputLocation) {
     lowerBoundary = lowerBoundary != null ? lowerBoundary : Integer.MIN_VALUE;
     upperBoundary = upperBoundary != null ? upperBoundary : Integer.MAX_VALUE;
+    // "integer" in the mathematical sense, not the Java sense
     String message = String.format("%s must be an integer", fieldName);
     if (lowerBoundary.equals(upperBoundary)) {
       message += String.format(" equal to %d", lowerBoundary);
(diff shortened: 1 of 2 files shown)


src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -271,6 +272,40 @@ private void checkStringToIntWithBoundaries(
     }
   }
 
+  // Makes sure String input can be converted to a long, and checks that long against boundaries
+  private void checkStringToLongWithBoundaries(String input, String inputName, Long lowerBoundary,
+      Long upperBoundary, boolean isRequired) {
+    lowerBoundary = lowerBoundary != null ? lowerBoundary : Long.MIN_VALUE;
+    upperBoundary = upperBoundary != null ? upperBoundary : Long.MAX_VALUE;
+    // "integer" in the mathematical sense, not the computer science sense
```
tarheel: well, s/computer science/Java/

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -271,6 +272,40 @@ private void checkStringToIntWithBoundaries(
     }
   }
 
+  // Makes sure String input can be converted to a long, and checks that long against boundaries
+  private void checkStringToLongWithBoundaries(String input, String inputName, Long lowerBoundary,
```
tarheel: We should be able to factor out the common logic between this and `checkStringToIntWithBoundaries`. I think you just need to take advantage of method references, which we haven't done yet in this repo: https://www.codementor.io/eh3rrera/using-java-8-method-reference-du10866vx

HEdingfield: I was thinking the same thing, but lamdas continually baffle me. Would you mind taking a crack at it?

tarheel: Can't during the work day, but I can look later. I don't think you need lambdas, though. I think you can make `numberParsingMethod` one of the params for your generic function and have it call `numberParsingMethod.apply(s)`, and then the callsites would pass `Integer::parseInt` or `Float::parseFloat`.

I've never done this in Java myself, though, so I'm not sure.

tarheel: If you want, you can move ahead with what you have, and I can try to send out a follow-up PR.

HEdingfield: What about this: I could make the original method have long parameters, which should accept ints as well (if not, maybe I could abstract it to `Number`). I can also make the boundaries primitives to force us to provide them, in which case we'd explicitly provide `Integer.MAX_VALUE` for ints, or whatever is appropriate.

tarheel: That's fine. I'm still interested to try my approach later, but proceed as you see fit for now.

HEdingfield:
> Code looks good. Did you test it using numbers at the edges? That's the only thing I'd want to do.

Yep, validation rejects one over the limits either way.

Issue #371, opened 2019-07-09
Consolidate validation logic for sources and candidates


HEdingfield:
Validation logic for sources and candidates should be moved to static boolean methods, so they can be used both by `ContestConfig` and also `GuiConfigController`'s `buttonAddCvrFileClicked` and `buttonAddCandidateClicked` methods.

Also consider creating a separate `Validator` class that takes config as a constructor to handle all validation logic.

HEdingfield:
Closed with #372. Feel free to re-open if you like the idea of having a separate `Validator` class in the future.

PR #372, opened 2019-07-09
Consolidates validation logic

HEdingfield:
Consolidates validation logic for sources and candidates (progress on #371); disables validation check when changing cells to avoid situation where config with bad sources is loaded in the GUI and can't get corrected. Renames and inverts `checkStringToIntWithBoundaries` to `fieldOutOfRangeOrNotInt`.

### DIFF
--- a/src/main/java/network/brightspots/rcv/ContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/ContestConfig.java
@@ -58,10 +58,10 @@ class ContestConfig {
   static final BigDecimal SUGGESTED_MINIMUM_VOTE_THRESHOLD = BigDecimal.ZERO;
   static final int SUGGESTED_MAX_SKIPPED_RANKS_ALLOWED = 1;
   static final WinnerElectionMode SUGGESTED_WINNER_ELECTION_MODE = WinnerElectionMode.STANDARD;
-  static final int MIN_COLUMN_INDEX = 1;
-  static final int MAX_COLUMN_INDEX = 1000;
-  static final int MIN_ROW_INDEX = 1;
-  static final int MAX_ROW_INDEX = 100000;
+  private static final int MIN_COLUMN_INDEX = 1;
+  private static final int MAX_COLUMN_INDEX = 1000;
+  private static final int MIN_ROW_INDEX = 1;
+  private static final int MAX_ROW_INDEX = 100000;
   private static final int MIN_MAX_RANKINGS_ALLOWED = 1;
   private static final int MIN_MAX_SKIPPED_RANKS_ALLOWED = 0;
   private static final int MIN_NUMBER_OF_WINNERS = 1;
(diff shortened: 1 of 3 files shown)


src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -224,51 +224,55 @@ boolean validate() {
     return isValid;
   }
 
-  private void invalidateAndLog(String message, String inputLocation) {
-    isValid = false;
+  private static void logWithLocation(String message, String inputLocation) {
```
tarheel: `logValidationErrorWithLocation`?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -296,78 +300,48 @@ private void validateOutputSettings() {
...
+      if (fieldOutOfRangeOrNotInt(
+          source.getPrecinctColumnIndex(), "precinctColumnIndex", MIN_COLUMN_INDEX,
+          MAX_COLUMN_INDEX, false, source.getFilePath())) {
+        sourceValid = false;
       }
     }
+    return sourceValid;
+  }
+
+  static boolean isCandidateValid(Candidate candidate) {
```
tarheel: `isCandidateNameValid`

This doesn't cover other ways in which the candidate might be invalid, e.g. a name or code collision or a missing code (when other candidates have codes).

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -412,19 +386,76 @@ private boolean candidateStringAlreadyInUseElsewhere(
...
   }
 
+  private void validateCvrFileSources() {
+    if (rawConfig.cvrFileSources == null || rawConfig.cvrFileSources.isEmpty()) {
+      isValid = false;
+      Logger.log(Level.SEVERE, "Contest config must contain at least 1 cast vote record file!");
+    } else {
+      HashSet<String> cvrFilePathSet = new HashSet<>();
+      for (CVRSource source : rawConfig.cvrFileSources) {
+        if (!isCvrSourceValid(source)) {
```
tarheel: I also find this method name (`isCvrSourceValid`) misleading. You actually validate a bunch of additional stuff related to the source below, outside of that method.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -400,80 +370,45 @@ public void buttonDeleteCvrFileClicked() {
...
 
   public void changeCvrFilePath(CellEditEvent cellEditEvent) {
-    CVRSource cvrSelected = tableViewCvrFiles.getSelectionModel().getSelectedItem();
-    // Cache the original value to revert back to in case the change fails
-    String oldFilePath = cvrSelected.getFilePath();
-    cvrSelected.setFilePath(cellEditEvent.getNewValue().toString().trim());
-    if (!cvrSourceHasRequiredFields(cvrSelected)) {
-      cvrSelected.setFilePath(oldFilePath);
-    }
+    tableViewCvrFiles.getSelectionModel().getSelectedItem()
```
tarheel: Could you explain what these changes do?

src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -153,6 +153,16 @@ public void setProvider(String provider) {
   @JsonIgnoreProperties(ignoreUnknown = true)
   @JsonInclude(JsonInclude.Include.NON_NULL)
   public static class Candidate {
+
+    public Candidate() {
```
tarheel: shouldn't these constructors go below the member variable declarations?

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: Yeah, but that forces it to be aware of what's going on elsewhere in the config, and thus it can't be `static` anymore, which is what I need. Could I just rename this to something like `initialCvrSourceValidationCheck`?

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Per the commit message: "disables validation check when changing cells to avoid situation where config with bad sources is loaded in the GUI and can't get corrected." So if I create a bad entry directly in the JSON with multiple bad fields, this allows you to actually fix them all. Otherwise, it resets to the bad fields across the board.

src/main/java/network/brightspots/rcv/ContestConfig.java
tarheel: Yeah, I'm fine renaming it. I didn't mean that you needed to change its behavior.

HEdingfield:
> Much cleaner. Does this actually change any of the validation logic?

Isn't it?? :man_dancing: The only thing it changes is if you supply garbage values for row or column index for a CDF source, it no longer ignores it, but says "must be an integer between x and y if supplied", which is better IMO because it's consistent with everything else and there's no reason to let people store obviously bad values in a tabulation-ready config.

Would one of you mind approving? I'll make @tarheel's changes and merge when I get home.

PR #395, opened 2019-09-10
fix two bugs related to undeclared write-ins

tarheel:
Bug 1: in a multi-seat contest, if someone wins in the first round, we don't automatically eliminate undeclared write-ins before we eliminate any other candidates. Instead, we treat UWI like a normal candidate, which means we potentially eliminate other candidates with lower tallies first.

Bug 2: if UWI exceeds the winning threshold in the initial count, we mistakenly elect this candidate.

### DIFF
--- a/src/main/java/network/brightspots/rcv/Tabulator.java
+++ b/src/main/java/network/brightspots/rcv/Tabulator.java
@@ -442,7 +442,12 @@ class Tabulator {
           if (tally.compareTo(winningThreshold) >= 0) {
             // we have winner(s)
             List<String> winningCandidates = currentRoundTallyToCandidates.get(tally);
-            selectedWinners.addAll(winningCandidates);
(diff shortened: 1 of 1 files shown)


tarheel:
> I'd like to see tests included that trigger these two bugs and fail in 1.0.1, but now successfully pass with these changes.

Yeah, good call. I don't have time to add them at the moment, but will do it soon.

HEdingfield:
@tarheel request on this, and for future PRs: could you please make sure to file issues for each bug (in this case, file two issues and then mark them as resolved immediately by this PR)? It'll make it a lot easier to write the release notes / change log for the next version. I think we need to be more rigorous about this now that we've gone official.

Could you also please submit an issue for anything you defer in a PR, like adding tests?

tarheel:
Yes, will do.

Issue #407, opened 2020-02-25
Read Dominion CVR Json format


moldover:
Give path to a folder containing Dominion manifest files, and using candidate and precinct data from #406, read Dominion CVR manifest file into memory as CastVoteRecords.  
To support #404 CastVoteRecord must be extended to include fields for:
tabulatorID, batchID, RecordID 

To support exporting individual contests (and future support of multiple contests) contestID must also be included.

Exhaustive CVR data (e.g. markDensity) need NOT be included.

moldover:
Done via #411

Issue #404, opened 2020-02-21
export Dominion CVR in generic CSV format


tarheel:
The tabulator needs functionality to export in-memory CastVoteRecords to a generic .csv file for human-friendly verification of the CSV import process with non-trivial processing (Dominion, ClearVote, ?) (#407).

@CalebKleppner and @moldover defined this spec around the 2020 Democratic Primaries in Kansas, Alaska, and Wyoming.

The tabulator exports a csv for each contest.
The export file location is the same folder as the input cvr json file.
  The export file name is:
CvrExport-[contestId].csv.  (if there is no contest ID just omit that portion).

For each contest, the exported .csv file is like this:
Header:
ContestId | TabulatorId | BatchId | RecordId | PrecinctId | Rank 1 | Rank 2 |  ... | Rank N

If contestId is NULL (ES&S) use Contest Name.

N = maxRanks for that contest.  maxRanks values for each contest are parsed from ContestManifest.json.

Columns #1 .. #n are populated with:
candidateID selected (marked) at that rank OR 
the string "over" / "under" if an under/over vote occurred

tarheel:
Should we also export a config file with some of the values (e.g. the candidates and the CVR source) prepopulated?

HEdingfield:
#434 filed to account for only lingering question in this issue.

PR #410, opened 2020-02-25
CastVoteRecord extensions

moldover:
Additional CastVoteRecord fields to support #407 and #404.

When #407 is complete these fields will be populated with data from imported CVR json data.  A user input contestId can then be used to filter exported output by contestId.  In the interim some hard-coded or programmatic values can be faked here to move forward with #404 in parallel.

### DIFF
--- a/src/main/java/network/brightspots/rcv/CastVoteRecord.java
+++ b/src/main/java/network/brightspots/rcv/CastVoteRecord.java
@@ -46,7 +46,7 @@ class CastVoteRecord {
   private final String precinct;
   // contest associated with this CVR
   private String contestId = null;
-  //tabulatorId parsed from Dominion CVR data
+  // tabulatorId parsed from Dominion CVR data
   private String tabulatorId = null;
   // batchId parsed from Dominion CVR data
   private String batchId = null;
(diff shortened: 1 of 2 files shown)


src/main/java/network/brightspots/rcv/CastVoteRecord.java
```diff
@@ -44,6 +44,12 @@
   private final String suppliedId;
   // which precinct this ballot came from
   private final String precinct;
+  // contest associated with this CVR
+  private String contestId = null;
+  //tabulatorId parsed from Dominion CVR data
```
tarheel: add space

src/main/java/network/brightspots/rcv/CastVoteRecord.java
```diff
@@ -44,6 +44,12 @@
   private final String suppliedId;
   // which precinct this ballot came from
   private final String precinct;
+  // contest associated with this CVR
+  private String contestId = null;
```
tarheel: Why aren't these new Strings final? Also, I think giving them default values of null happens automatically.

tarheel: Why aren't these new Strings final? Also, I think giving them default values of null happens automatically.

moldover:
I fixed the merge base and your comment on strings was deleted?  Weird.  Anyways, they might be reassigned, that's why not final, and yes java objects will be null by default.  I expect this to change as the other stuff is built.  I'd like to get a prototype working and into Caleb's hands for iteration first, then worry about things like that if it still matters.

PR #417, opened 2020-03-03
export generic csv functionality #404

moldover:
Adds support for writing generic .csv.  This allows verifying Dominion csv json import, and tabulation.

### DIFF
--- a/src/main/java/network/brightspots/rcv/CastVoteRecord.java
+++ b/src/main/java/network/brightspots/rcv/CastVoteRecord.java
@@ -59,13 +59,13 @@ class CastVoteRecord {
   // a set is used to handle overvotes
   SortedMap<Integer, Set<String>> rankToCandidateIds;
   // contest associated with this CVR
-  private Integer contestId;
+  Integer contestId;
   // tabulatorId parsed from Dominion CVR data
-  private Integer tabulatorId;
+  Integer tabulatorId;
   // batchId parsed from Dominion CVR data
-  private Integer batchId;
+  Integer batchId;
   // ballotTypeId parsed from Dominion CVR data
-  private Integer ballotTypeId;
+  Integer ballotTypeId;
   // whether this CVR is exhausted or not
   private boolean isExhausted;
   // tells us which candidate is currently receiving this CVR's vote (or fractional vote)
@@ -204,6 +204,10 @@ class CastVoteRecord {
 
   String getPrecinct() {
     return precinct;
+  }
+
+  String getPrecinctId() {
+    return suppliedId != null ? suppliedId : computedId;
   }
 
   Map<String, BigDecimal> getWinnerToFractionalValue() {
--- a/src/main/java/network/brightspots/rcv/CastVoteRecord.java
+++ b/src/main/java/network/brightspots/rcv/CastVoteRecord.java
@@ -58,14 +58,31 @@ class CastVoteRecord {
   // map of round to all candidates selected for that round
   // a set is used to handle overvotes
   SortedMap<Integer, Set<String>> rankToCandidateIds;
+
   // contest associated with this CVR
-  Integer contestId;
+  private Integer contestId;
   // tabulatorId parsed from Dominion CVR data
-  Integer tabulatorId;
+  private Integer tabulatorId;
   // batchId parsed from Dominion CVR data
-  Integer batchId;
+  private Integer batchId;
   // ballotTypeId parsed from Dominion CVR data
-  Integer ballotTypeId;
+  private Integer ballotTypeId;
+
+  public Integer getContestId() {
+    return contestId;
+  }
+
+  public Integer getTabulatorId() {
+    return tabulatorId;
+  }
+
+  public Integer getBatchId() {
+    return batchId;
+  }
+
+  public Integer getBallotTypeId() {
+    return ballotTypeId;
+  }
   // whether this CVR is exhausted or not
   private boolean isExhausted;
   // tells us which candidate is currently receiving this CVR's vote (or fractional vote)
--- a/src/main/java/network/brightspots/rcv/CastVoteRecord.java
+++ b/src/main/java/network/brightspots/rcv/CastVoteRecord.java
@@ -68,21 +68,22 @@ class CastVoteRecord {
   // ballotTypeId parsed from Dominion CVR data
   private Integer ballotTypeId;
 
-  public Integer getContestId() {
+  Integer getContestId() {
     return contestId;
   }
 
-  public Integer getTabulatorId() {
+  Integer getTabulatorId() {
     return tabulatorId;
   }
 
-  public Integer getBatchId() {
+  Integer getBatchId() {
     return batchId;
   }
 
-  public Integer getBallotTypeId() {
+  Integer getBallotTypeId() {
     return ballotTypeId;
   }
+
   // whether this CVR is exhausted or not
   private boolean isExhausted;
   // tells us which candidate is currently receiving this CVR's vote (or fractional vote)
--- a/src/main/java/network/brightspots/rcv/DominionCvrReader.java
+++ b/src/main/java/network/brightspots/rcv/DominionCvrReader.java
@@ -32,14 +32,20 @@ import network.brightspots.rcv.RawContestConfig.Candidate;
 class DominionCvrReader {
 
   // canonical manifest file names
-  static final String PRECINCT_MANIFEST = "PrecinctPortionManifest.json";
-  static final String CANDIDATE_MANIFEST = "CandidateManifest.json";
-  static final String CONTEST_MANIFEST = "ContestManifest.json";
-  static final String CVR_EXPORT = "CvrExport.json";
+  private static final String PRECINCT_MANIFEST = "PrecinctPortionManifest.json";
+  private static final String CANDIDATE_MANIFEST = "CandidateManifest.json";
+  private static final String CONTEST_MANIFEST = "ContestManifest.json";
+  private static final String CVR_EXPORT = "CvrExport.json";
 
   private final String manifestFolder;
   private Map<Integer, String> precincts;
-  public Map<Integer, Contest> contests;
+
+  private Map<Integer, Contest> contests;
+
+  public Map<Integer, Contest> getContests() {
+    return contests;
+  }
+
   private List<Candidate> candidates;
 
   DominionCvrReader(String manifestFolder) {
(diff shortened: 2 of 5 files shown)


src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -39,15 +39,15 @@
 
   private final String manifestFolder;
   private Map<Integer, String> precincts;
-  private Map<Integer, Contest> contests;
+  public Map<Integer, Contest> contests;
```
HEdingfield: Please keep private and make a project-visible getter (i.e. no need for `public` even on that).

src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -42,6 +42,12 @@ public static void main(String[] args) {
     Logger.setup();
     logSystemInfo();
 
+    // for testing
+    String wyoming = "/Flair/Downloads/Dominion/Wyoming";
+    String sf = "/Flair/Downloads/Dominion/SF";
+    TabulatorSession.convertDominionCvrJsonToGenericCsv(wyoming);
```
HEdingfield: Remember to revert.

src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -245,14 +245,14 @@ void parseCvrFile(String filePath, List<CastVoteRecord> castVoteRecords) {
...
-    private final Integer id;
-    private final Integer numCandidates;
-    private final Integer maxRanks;
+    public String name;
+    public Integer id;
+    public Integer numCandidates;
+    public Integer maxRanks;
 
-    private Contest(String name, Integer id, Integer numCandidates, Integer maxRanks) {
+    public Contest(String name, Integer id, Integer numCandidates, Integer maxRanks) {
```
HEdingfield: I don't think anything in this entire project should be `public` unless it's being directly accessed by the Jackson library (e.g. `RawContestConfig`).

The first place I see this being accessed is in `writeGenericCvrCsv`. It only looks like `maxRanks` and `id` are retrieved, so I'd write project-visible getters for these.

The second place I see this being accessed is in the `TabulatorSession` line: `dominionCvrReader.contests.values()`.

I've never seen `.values()` of a class used like that before, but for safety's sake, I'd lean strongly towards keeping all these `private` and `private final` and maybe making a project-visible method that explicitly accesses, formats, and returns these values in the way that makes `writer.writeGenericCvrCsv` happy (i.e. a `Collection<Contest>`).

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -488,6 +491,76 @@ void generateOverallSummaryFiles(
     generateSummaryJson(roundTallies, tallyTransfers, null, outputPath);
   }
 
+  // write CastVoteRecords for all contests to the same folder as the csvInputFile
+  // create the folder if needed
```
HEdingfield: How is it possible that the folder would need to be created if it's the same as the input?

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -488,6 +491,76 @@ void generateOverallSummaryFiles(
...
+  // folder will default to the input Dominion csv folder
```
HEdingfield: This function isn't specific to Dominion is it? If so, recommend removing this comment, particularly since it's already mentioned above that it writes to the same folder as the csvInputfile.

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -488,6 +491,76 @@ void generateOverallSummaryFiles(
...
+      throws IOException {
+    try {
+      File csvFile = new File(csvInputFile);
+      String outputFolder = csvFile.getParent();
+      String slug = csvFile.getName();
+      for (Contest contest : contests) {
+        String fileName = String.format("%s-contest-%d.csv", slug, contest.id);
+        Path outputPath = Paths.get(outputFolder, fileName);
+        Logger.log(Level.INFO,
+            "Writing generic cast vote record: %s...",
```
HEdingfield:
```suggestion
            "Writing cast vote records in generic format to file: %s...",
```

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -488,6 +491,76 @@ void generateOverallSummaryFiles(
...
+        csvPrinter = new CSVPrinter(writer, CSVFormat.DEFAULT);
+        // print header:
+        // ContestId, TabulatorId,  BatchId, RecordId, PrecinctId, rank 1 selection,
+        // rank 2 selection, ... rank maxRanks selection
+        csvPrinter.print("ContestId");
+        csvPrinter.print("TabulatorId");
+        csvPrinter.print("BatchId");
+        csvPrinter.print("RecordId");
+        csvPrinter.print("PrecinctId");
+        Integer numranks = contest.maxRanks;
```
HEdingfield:
```suggestion
        Integer numRanks = contest.maxRanks;
```

src/main/java/network/brightspots/rcv/CastVoteRecord.java
```diff
@@ -59,13 +59,13 @@
...
+  Integer contestId;
   // tabulatorId parsed from Dominion CVR data
-  private Integer tabulatorId;
+  Integer tabulatorId;
   // batchId parsed from Dominion CVR data
-  private Integer batchId;
+  Integer batchId;
   // ballotTypeId parsed from Dominion CVR data
-  private Integer ballotTypeId;
+  Integer ballotTypeId;
```
HEdingfield: Please keep all these private (including `rankToCandidateIds`) and create getters instead (like we've done in other classes). I believe there's even a shortcut to do this in IntelliJ with the push of a button.

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -488,6 +491,76 @@ void generateOverallSummaryFiles(
...
+        for (Integer rank = 1; rank <= numranks; rank++) {
```
HEdingfield:
```suggestion
        for (int rank = 1; rank <= numRanks; rank++) {
```

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -488,6 +491,76 @@ void generateOverallSummaryFiles(
...
+        csvPrinter.println();
+        // print rows:
+        for (CastVoteRecord castVoteRecord : castVoteRecords) {
+          csvPrinter.print(castVoteRecord.contestId);
+          csvPrinter.print(castVoteRecord.tabulatorId);
+          csvPrinter.print(castVoteRecord.batchId);
+          csvPrinter.print(castVoteRecord.getId());
+          csvPrinter.print(castVoteRecord.getPrecinctId());
+          // for each rank determine what candidate id, overvote, or undervote ocurred
+          for (Integer rank = 1; rank <= contest.maxRanks; rank++) {
```
HEdingfield:
```suggestion
          for (int rank = 1; rank <= numRanks; rank++) {
```

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -488,6 +491,76 @@ void generateOverallSummaryFiles(
...
+            } else {
+              csvPrinter.print("undervote");
+            }
+          }
+          csvPrinter.println();
+        }
+        // finalize the file
+        csvPrinter.flush();
+        csvPrinter.close();
+        Logger.log(Level.INFO, "Wrote %s.", outputPath.toString());
```
HEdingfield: I think we can probably get rid of this, since in our other instances of creating notifications that we were writing files, I don't think we provided a follow-up message confirming it was completed.

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -488,6 +491,76 @@ void generateOverallSummaryFiles(
...
+      }
+    } catch (IOException exception) {
+      Logger.log(Level.SEVERE, "Error writing generic cast vote record: %s", exception.toString());
```
HEdingfield:
```suggestion
      Logger.log(Level.SEVERE, "Error writing cast vote records in generic format to file: %s\n%s", outputPath.toString(), exception.toString());
```

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -76,6 +76,28 @@ String getTimestampString() {
...
   }
 
+  // given a dominion style csv json file path:
+  // read associated manifest data
+  // read Dominion cvr json into CastVoteRecords
+  // write CastVoteRecords to generic cvr csv files: one per contest
+  // return true if files are successfully written
+  // return false if an error is encountered
+  public static Boolean convertDominionCvrJsonToGenericCsv(String dominionCvrJsonFile) {
+    Boolean success = true;
```
HEdingfield: Is there a particular reason we're returning a boolean here instead of making this `void` like `convertToCdf()`? I'd lean towards copying that design pattern if not.

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -76,6 +76,28 @@ String getTimestampString() {
...
+    DominionCvrReader dominionCvrReader = new DominionCvrReader(dominionCvrJsonFile);
+    List<CastVoteRecord> castVoteRecords = new ArrayList<>();
+    try {
+      dominionCvrReader.readCastVoteRecords(castVoteRecords);
+      ResultsWriter writer = new ResultsWriter();
+      writer.writeGenericCvrCsv(castVoteRecords, dominionCvrReader.contests.values(),
+          dominionCvrJsonFile);
+    } catch (Exception exception) {
+      Logger.log(Level.SEVERE, "Failed to convert Cvr Json: %d" + exception.toString());
```
```suggestion
      Logger.log(Level.SEVERE, "Failed to convert Dominion CVR to CSV:\n%s", exception.toString());
```

src/main/java/network/brightspots/rcv/TabulatorSession.java
moldover: yeah, it's a better api.  The caller should have a way of knowing if it succeeded or not.

tarheel: nit: boolean, not Boolean

src/main/java/network/brightspots/rcv/ResultsWriter.java
moldover: I'm going to keep it - it's informative and useful.  Unless @CalebKleppner  disagrees.

src/main/java/network/brightspots/rcv/TabulatorSession.java
HEdingfield: In that case, should we file and issue to do this for `convertToCdf()` and `tabulate()` as well?

Issue #408, opened 2020-02-25
Add user interface to support Dominion conversion process


moldover:
After talking with @CalebKleppner we arrived at the following:

The tabulator needs to ingest and process Dominion json data for various elections (Wyoming, Alaska, Kansas Democratic primaries, and eventually NYC local).  

To allow users to input, convert (verification / auditing - #404) and tabulate, we need a way for users to specify Dominion as a data provider.  For Dominion data providers, the user needs to specify an input Dominion folder location.  (First Vote Column Index, etc... are not needed)

**CLI support for conversion** this can be run from CLI: user supplies input Dominion data folder location.  Tabulator will input Dominion cvr json data from this folder.
Tabulator will export generic cvr csv to the same folder (#404) with a file per contest.

**GUI support for conversion**

User can click a "convert dominion cvr json to generic csv" button on the main panel.  This opens a file chooser.  User can select a file to convert.  Exported files will be written to the same folder as the input file.  Per @CalebKleppner no need to select an export folder.

Moved the Tabulation support tasks to a new ticket: #418

tarheel:
Should there also be an arg for the output file path?

moldover:
@CalebKleppner what do you think about Louis' question above?

CalebKleppner:
A few comments:  1. The format from SF in Nov 2019 is different from the format going forward in SF and the Dem primaries.  The Dem primaries are using version 5.10.11.24.

HEdingfield:
Relevant code that was used to temporarily test the new functionality:

```
    // for testing	
    String wyoming = "/Flair/Downloads/Dominion/Wyoming";	
    String sf = "/Flair/Downloads/Dominion/SF";	
    List<CastVoteRecord> castVoteRecords = new ArrayList<>();	
    DominionCvrReader test = new DominionCvrReader(wyoming);	
    try {	
      test.readCastVoteRecords(castVoteRecords);	
    } catch (CvrParseException e) {	
      e.printStackTrace();	
    }
```

moldover:
Lol!  So I was looking for this sample code yesterday and the squash commit didn't have it. Sadness.  Going forward I'm going to use my judgement to squash or not.  Good to have in the toolbox.

[...]

PR #419, opened 2020-03-08
Adds CLI option to convert Dominion CVR to generic .csv (progress on #408)

HEdingfield:


README.md
```diff
@@ -85,14 +85,18 @@ Or, if you're compiling and running using Gradle:
...
-Finally, you can activate a special `convert-to-cdf` function via the command line to export the CVR as a NIST common data format (CDF) .json instead of tabulating the results, e.g.:
+You can also activate a special `convert-to-cdf` function via the command line to export the CVR as a NIST common data format (CDF) .json instead of tabulating the results, e.g.:
 
 `$ rcv -cli path/to/config convert-to-cdf`
 
 Or, again, if you're compiling and running using Gradle:
 
 `$ gradlew run --args="-cli path/to/config  convert-to-cdf"`
 
+Finally, there's a `convert-dominion-to-generic-csv` option as well for converting Dominion JSON CVRs into a standardized, generic .csv format:
```
tarheel: Perhaps here (and above) should mention where the output file will be located.

src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -73,12 +73,15 @@ public static void main(String[] args) {
...
-      String configPath = argsCli.get(0);
-      boolean convertToCdf = argsCli.size() == 2 && argsCli.get(1).equals("convert-to-cdf");
-      // session object will manage the tabulation process
-      TabulatorSession session = new TabulatorSession(configPath);
-      if (convertToCdf) {
+      // Path to either: config file for configuring the tabulator, or Dominion JSONs
+      String providedPath = argsCli.get(0);
+      if (argsCli.size() == 2 && argsCli.get(1).equals("convert-dominion-to-generic-csv")) {
+        TabulatorSession.convertDominionCvrJsonToGenericCsv(providedPath);
+        System.exit(0);
```
tarheel: I don't feel strongly about this, but I think it would be a bit cleaner to combine this with the if block below and just have a single `System.exit` call at the end of the method.

HEdingfield: Sure, I guess it's slightly more readable that way. I was trying to avoid creating `TabulatorSession session = new TabulatorSession(providedPath);` without needing to use it, but no big deal.

README.md
HEdingfield: I don't think it's necessary because it says clearly where it outputs when you run it (and I anticipate this changing and don't want to have to remember to update the docs).

PR #424, opened 2020-03-22
Dem primary

### LINKED ISSUE #403: Incorporating Democratic Party Primary RCV tabulation in to Tabulator
As mentioned on the call a few weeks ago, the Democratic parties in Wyoming, Kansas, and Alaska are planning on using the Universal Tabulator to determine the results of their RCV presidential primaries this year. Clear Ballot is providing the scanners for the election; I spoke with Don Deford, who works at Clear Ballot, last week about what they need from the tabulator to run the election. Right now it's possible to get very close to the 15% threshold required by the rules (electing six candidates in a bottoms up contest would get you very close with a 14.3% threshold; five candidates would, also, with a 16.67% threshold, but not precise enough if a candidate comes within .7% to 15%). So, the only missing feature from the Tabulator for the Primaries is the ability to set an arbitrary threshold. 

As I mentioned, the Democratic Primaries use an arbitrary 15% threshold to award delegates. That is the threshold of election they'll use for the RCV contests too, paired with bottoms up RCV allocation. So the election will follow typical single-winner RCV rules, eliminating one candidate at a time (or batch eliminating candidates) and edistributing votes from those candidates until only candidates with 15%+ of the vote are left. I'm not certain of this but I believe that 15% is set based on the votes cast in the first round, and is not dynamic based on the number of valid votes in each round. (this isn't specified in the rules, unfortunately). 

The Wyoming and Alaska primaries are scheduled for April 4, and Kansas is scheduled for May 2. 

A few things i wanted to raise that need deciding: 
- How should this arbitrary threshold option present itself to the user? Only if they select bottoms up? By selecting something labeled Democratic Presidential Primary? That selective appearance is not a function we've used before, though. Seems best to limit the availability of arbitrary threshold so that people don't accidentally use it, which is why I'm raising it here
- Some states award delegates at the statewide and congressional district levels - I wonder if it would be possible to take statewide CVRs that have the Congressional District noted as the precinct and run both statewide and precinct results (but treating precincts as individual contests, unlike current precinct counts in the Tabulator). happy to be shot down on this if it adds too much complexity to the tabulator. 

Also attaching the delegate selection rules I've gotten so far. Hawaii, which won't be using the tabulator, hasn't yet sent me theirs. Kansas lays out their RCV rules on pages 3-5; Alaska on pages 6-7; and Wyoming on pages 8-9. 
[2020-Kansas-Delegate-Selection-Plan_122019.pdf](https://github.com/BrightSpots/rcv/files/4148730/2020-Kansas-Delegate-Selection-Plan_122019.pdf)
[AK+DSP+12.20.19.pdf](https://github.com/BrightSpots/rcv/files/4148731/AK%2BDSP%2B12.20.19.pdf)
[WY Delegate Selection Plan Oct. 9.19.pdf](https://github.com/BrightSpots/rcv/files/4148732/WY.Delegate.Selection.Plan.Oct.9.19.pdf)

tarheel:
Fixes #403 .


src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -568,43 +570,80 @@ private void validateRules() {
...
-      } else if (isMultiSeatAllowOnlyOneWinnerPerRoundEnabled()) {
-        isValid = false;
-        Logger.log(
-            Level.SEVERE,
-            "winnerElectionMode can't be multiSeatAllowOnlyOneWinnerPerRound in a single-seat "
-                + "contest!");
-      }
+    boolean isMultiSeatBottomsUpPercentageThresholdSpecified =
+        getMultiSeatBottomsUpPercentageThresholdRaw() != null
+            && !getMultiSeatBottomsUpPercentageThresholdRaw().isBlank();
```
HEdingfield: Could you just set this to `getMultiSeatBottomsUpPercentageThreshold() != null`, or is the `.isBlank()` check important as well? If so, could you just put the latter in the former helper function and use it?

src/test/resources/network/brightspots/rcv/test_data/multi_seat_bottoms_up_with_threshold/multi_seat_bottoms_up_with_threshold_config.json
```diff
@@ -10,7 +10,7 @@
     "generateCdfJson" : false
   },
   "cvrFileSources" : [ {
-    "filePath" : "/Users/leisenberg/code/rcv/src/test/resources/network/brightspots/rcv/test_data/multi_seat_bottoms_up_with_threshold/multi_seat_bottoms_up_with_threshold_cvr.xlsx",
+    "filePath" : "test_data/multi_seat_bottoms_up_with_threshold/multi_seat_bottoms_up_with_threshold_cvr.xlsx",
```
HEdingfield: This should just be `multi_seat_bottoms_up_with_threshold_cvr.xlsx`.

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
```diff
@@ -258,7 +258,7 @@
                 <Label text="Undeclared Write-in Label"/>
                 <TextField fx:id="textFieldUndeclaredWriteInLabel" maxWidth="220.0"
                   promptText="Undeclared Write-in Label"/>
-                <Label text="Multi-Seat Bottoms-Up Percentage Threshold"/>
+                <Label text="Multi-Seat Bottoms-up Percentage Threshold"/>
                 <TextField maxWidth="220.0" promptText="Bottoms-Up Percentage Threshold"
```
HEdingfield: Nit: Oops, missed the "Bottoms-up" here too

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -192,7 +197,8 @@
         }
       } else if (winnerToRound.size() < config.getNumberOfWinners()
           || (config.isSingleSeatContinueUntilTwoCandidatesRemainEnabled()
-          && candidateToRoundEliminated.size() < config.getNumCandidates() - 2)) {
+          && candidateToRoundEliminated.size() < config.getNumCandidates() - 2)
+          || config.isMultiSeatBottomsUpWithThresholdEnabled()) {
         // We need to make more eliminations if
         // a) we haven't found all the winners yet, or
         // b) we've found our winner, but we're continuing until we have only two candidates
```
moldover:
```suggestion
        // b) we've found our winner, but we're continuing until we have only two candidates
        // c) not all remaining candidates meet the bottoms up threshold 
```

Issue #415, opened 2020-03-01
Add test(s) for Dominion parsing


HEdingfield:
Per the discussion in PR #411, we'll need to created automated tests (using provided input files) for Dominion once we receive verification of the expected results from @CalebKleppner. I believe #406 and #407 are blockers for this.

HEdingfield:
This should also test the `convert-dominion-to-generic-csv` CLI option.

moldover:
@HEdingfield what do you want to test about the CLI option?  It's pretty simple, right?

HEdingfield:
I think this is fully covered by #425. We should do something similar for the `convert-to-cdf` functionality.

PR #425, opened 2020-03-22
test for Dominion csv conversion routine #415

moldover:
This includes the new test routine for conversion comparison and the first test.  I'll likely add a few more tests before closing the issue, but wanted to get this in the pipeline since we're coming down to the wire here.


src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -97,6 +98,18 @@ private static String getTestFilePath(String stem, String suffix) {
...
   }
 
+  // helper function test cvr conversion routine
+  private static void runDominionCvrConversionTest(String stem) {
+    String dominionDataFolder = getTestFilePath(stem, "_cvr_export");
+    List<String> filesWritten = TabulatorSession
+        .convertDominionCvrJsonToGenericCsv(dominionDataFolder);
+    for (String convertedFile : filesWritten) {
+      String referencePath =
+          convertedFile.substring(0, convertedFile.length() - 4) + "_expected.csv";
```
tarheel: Is there a more robust way to get the part of a file before the extension? This would break if the length of the extension changed, right?

tarheel: e.g. you could write a regex or use something like this: https://stackoverflow.com/questions/924394/how-to-get-the-filename-without-the-extension-in-java

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -64,38 +64,38 @@
...
       ResultsWriter writer = new ResultsWriter();
-      writer.writeGenericCvrCsv(castVoteRecords, dominionCvrReader.getContests().values(),
-          dominionCvrJsonFile);
+      filesWritten = writer
+          .writeGenericCvrCsv(castVoteRecords, dominionCvrReader.getContests().values(),
+              dominionDataFolder);
     } catch (Exception exception) {
       Logger.log(Level.SEVERE, "Failed to convert Dominion CVR to CSV:\n%s", exception.toString());
-      success = false;
+      filesWritten = null;
```
HEdingfield: Is it possible that some files were written even if there was a failure for one of them though?

src/test/java/network/brightspots/rcv/TabulatorTests.java
HEdingfield: Regex is notoriously slow and has some other issues that cause us to avoid using it at Google, but I do agree that we should use a more robust method here. Maybe splitting on `.`? I guess that could fail if there are .s in the folder names (heaven forfend!). Maybe the apache.commons solution would work well, since we're already pulling that library in in `ResultsWriter`?

src/main/java/network/brightspots/rcv/TabulatorSession.java
moldover: Yeah, the idea is a null return value indicates failure in case the caller cares.  More important to fail loudly here (and pretty much everywhere in the Tabulator) than try to recover some partial results and do something with them.  I'll indicate that in the comment.

.idea/misc.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="ExternalStorageConfigurationManager" enabled="true" />
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_13" default="true" project-jdk-name="13.0.2" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_13" default="true" project-jdk-name="13" project-jdk-type="JavaSDK">
```
HEdingfield: Please revert this change. To prevent this, I think you just need to name your local version of the JDK in IntelliJ "13.0.2" instead of "13" (which I thought you'd done previously, so weird this is happening).

moldover: drr... thanks.

Issue #426, opened 2020-03-23
regression tests fail when app version is incremented


moldover:
Currently the validation code rejects any config file whose version does not match the current app version.  This is not correct because we do support older versions.  Also it has resulted in devs updating test asset config files which is incorrect - test assets need to remain stable.

HEdingfield:
We had very detailed discussions about this in #43 and #327. I believe the proper way to address this is to is the latest suggestion I made in #327:

> Related to my original comment here, but distinct: now that the version is in our config files, this requires us to update every test each time we increment the version, which is a pain (since the CLI is blocked if application version doesn't match config version).
> 
> I propose that we create a CLI parameter to migrate the config files to the application version. It would essentially replicate what opening the file and then saving it in the GUI would do, but faster.
> 
> Also wondering: what if someone loads a newer config in an older version of the tabulator? We'll have to think about how to deal with that situation too eventually.

HEdingfield:
Decided instead to make a special version "TEST" that allows tests to bypass this check.

PR #427, opened 2020-03-23
Fix for #426

moldover:
Fix for #426 - this allows test assets to remain stable as app version updates.
The only substantive change is in ContestConfig by removing the requirement that the config version equal the current app version.


HEdingfield:
Replicated in #435 to avoid commit log jank.

PR #436, opened 2020-03-25
Fixes Dominion conversion output filenames and path

HEdingfield:


src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -87,6 +92,17 @@ private static boolean fileCompare(String path1, String path2) {
     } catch (IOException e) {
       Logger.log(Level.SEVERE, "Error reading file!\n%s", e.toString());
       result = false;
+    } finally {
+      try {
```
moldover: Were you seeing some issues with file handles leaking?

src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -98,15 +114,26 @@ private static String getTestFilePath(String stem, String suffix) {
...
-      assertTrue(fileCompare(convertedFile, referencePath));
+      String contestNumber = convertedFile
+          .substring(convertedFile.lastIndexOf('_') + 1, convertedFile.lastIndexOf('.'));
+      String expectedPath = Paths
+          .get(dominionDataFolder, stem + "_contest_" + contestNumber + "_expected.csv")
+          .toAbsolutePath().toString();
+      assertTrue(fileCompare(convertedFile, expectedPath));
+      // Clean up test file(s)
+      try {
+        Files.delete(Paths.get(convertedFile));
```
moldover: Again good catch.

src/test/java/network/brightspots/rcv/TabulatorTests.java
HEdingfield: Yeah, it was making it impossible to delete the Dominion test files. Strange that it wasn't an issue for the normal test files.

HEdingfield:
I thought about doing that, but I realized the "output" folder is actually specified in all of our config files (and could easily be something other than "output"). The Dominion conversion doesn't use a config file (in fact, I feed it `null` instead), so it seems to make more sense to me to put it directly in the Dominion folder.

When we later on build in the ability to go directly from Dominion to tabulated results I'd argue that we should then include the converted generic CVR in the output folder specified by the config. Hopefully this makes sense.

moldover:
Yeah - totally agree.  Good call.

PR #441, opened 2020-03-26
make batch elimination work properly with singleSeatContinueUntilTwoCandidatesRemain

### LINKED ISSUE #396: ensure that batch elimination is compatible with continueUntilTwoCandidatesRemain
Recommendation from Chris Jerdonek and David Cary:

> We agreed that the way to modify the batch elimination rule so that it also works with reduce-to-two is simply to exclude from consideration batches that would eliminate all but one candidate. So one way to state the rule would be as follows:
> 
> Define a "batch" of candidates in a round to be all of the candidates whose vote totals are less than some candidate (you can call this the "next" candidate). Then to do a batch elimination, eliminate the largest batch that satisfies both of the following conditions (1) the sum of the vote totals of the candidates in the batch is strictly less than the vote total of the next candidate, and (2) at least two candidates aren't in the batch.

tarheel:
Fixes #396.

This was kind of tricky; I implemented it the wrong way the first time I tried it. By the end, I felt like I was wrapping up a technical interview that hadn't gone particularly well.


src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -202,7 +202,7 @@
         // We need to make more eliminations if
         // a) we haven't found all the winners yet, or
         // b) we've found our winner, but we're continuing until we have only two candidates
-        // c) not all remaining candidates meet the bottoms up threshold 
+        // c) not all remaining candidates meet the bottoms-up threshold
```
HEdingfield: @tarheel: a man after my own heart.

tarheel:
> Why is the running total 2 for two of them, and 14 for the other two? That doesn't make sense to me.

"Running total" refers to the total number of votes that we've encountered so far when we decide that we should add a candidate to the batch elimination. So first we see the two candidates who got 1 vote each. That makes a running total of 2 votes. The next-lowest (oops, I mistakenly say "next-highest" in the code; I'll update that) has 4, so those first two candidates can't possibly catch up. And then we end up looking at the candidates who got 4 and 5 votes respectively. 4+5+1+1 makes a running total of 11, so we know that those two can't catch up to the next-lowest tally of 14.

In other words: the basic rule for the batch elimination is that whenever we encounter a tally that exceeds the sum of the votes for all the candidates with lower tallies (which I'm calling the "running total" because of how the algorithm proceeds), we know we can batch-eliminate all of those candidates with lower tallies.

Thinking about this a bit more... it would be faster to start the search from the other direction (high tallies -> low tallies), because then as soon as you've found a batch, you can stop. The fact that my approach's output shows different running totals for some candidates vs. others is a symptom of some wasted effort, since the batch of two candidates ended up being subsumed by the batch of four. I don't think it affects the correctness, though, so I'm not going to worry about it for now. I'll file an issue for later.

PR #459, opened 2020-07-09
Update deps

### LINKED ISSUE #283: Test fails if we upgrade to a more recent version of Gradle
As outlined in #272, I tried updating beyond Gradle 4.10.3 (i.e. a number of versions all the way up to 5.4.1), and `test2013MinneapolisMayorScale()` would crap out around 500k votes with the message "TestEngine with ID 'junit-jupiter' failed to execute tests". It makes me think they added some sort of default time or memory limit for each test.

Ended up sticking with 4.10.3 (the IntelliJ default) for the purpose of that PR, but would be nice to be able to update to a more recent version of Gradle, or at least prepare for when IntelliJ does this naturally.

HEdingfield:
* Updates to latest version:
    * Gradle 6.5.1
    * JDK 14.0.1
    * JavaFX 14.0.1
    * Checkstyle google_checks.xml 8.34
    * Checkstyle plugin 8.34
    * org.openjfx.javafxplugin 0.0.9
    * org.beryx.jlink 2.20.0
    * com.fasterxml.jackson.core:jackson-* 2.11.1
    * org.junit.jupiter.junit-jupiter-* 5.6.2
* Gets tests working with Gradle 6.5.1 (fixes #283)
* Updates `APP_VERSION` to temporary value 1.2.0_alpha
* Addresses Checkstyle warnings
* Fixes versions in Hart tests


src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -51,6 +51,10 @@
     this.manifestFolder = manifestFolder;
   }
 
+  Map<Integer, Contest> getContests() {
```
HEdingfield, line 54: @moldover Is it intentional to have two different `getContest` functions?

moldover: Definitely not!  Probably the second one should be named parseContestData or something like that.

build.gradle
```diff
@@ -49,13 +49,19 @@ idea {
...
 compileJava {
     options.compilerArgs << "-Xlint:unchecked" << "-Xlint:deprecation"
 }
 
 test {
+    moduleOptions {
+        addExports = ["org.junit.platform.commons/org.junit.platform.commons.util": "ALL-UNNAMED",
+                      "org.junit.platform.commons/org.junit.platform.commons.logging": "ALL-UNNAMED"]
+    }
+    maxHeapSize = "2056m"
```
moldover, line 64: woah.  what does this mean?

src/main/java/network/brightspots/rcv/DominionCvrReader.java
HEdingfield: Done.

build.gradle
HEdingfield: So the moduleOptions setting was necessary for tests to actually run after doing the upgrades, and the maxHeapSize (I determined through trial and error) is what allows the big tests to actually finish running and not crap out (i.e. the fix for #283).

moldover:
Looks good to me.  It builds and tests pass, and I like the alpha version :)   @tarheel any problems?

Issue #461, opened 2020-07-13
GUI redesign phase 1


HEdingfield:
Implement the simplest changes specified in the Center for Civic Design Universal Tabulator Redesign Final mockups document (`Tabulator (FINAL mockups) May 2020.docx`).

These will mainly consist of moving existing options to new tabs and rearranging the order they appear in.

HEdingfield:
Provider dropdown should change type of dialog box opened up by "select" button for CVR file path (e.g. CDF should only allow JSON, ES&S should only allow .xlsx and .xls, Hart should only allow directory, etc.).

HEdingfield:
Refugees from #152:

* Delete Selected buttons should be disabled if nothing is selected
* Ability to drag and edit rows (e.g. for candidate permutations)

HEdingfield:
Notes after reviewing the final UI mockups doc today:



@hedingfield:
* Some stuff was moved to the Output tab from Contest Info, and I'm definitely not going to restructure the config files at all in light of this, so our .txt description doc will have things out of order which isn't ideal in the short-term unless I can get to adding the hints
* In this phase, I won't be doing the overvote label, treat blank as UWI settings, etc. on a per-source-file level now since that'd necessitate a config structure change
* I've got general concerns about fitting too much stuff in a panel due to resolution issues... should I bump it to 1600x1024 as the minimum required resolution, or even bigger? And if a display doesn't have that much space, I'll need to auto-expand it and figure out how to make scrollbars behave properly.
* I'm also not a huge fan of disappearing the log window, and it likely won't happen in this phase.

@tarheel:
* For winner election mode, they recommended eliminating support for "single seat until 2 candidates remain." But we need to support that somehow. I guess the alternative is to go back to what I believe we had before, which is to make "continue until two remain" a boolean in the config. We could put it under the output settings, since, as they point out, it only affects how the output is presented, not who actually wins.
* When the mode is bottoms up with percentage, number of winners should not be a field that the user sets. (Right now we have a single multiSeatBottomsUp mode, and we differentiate between percentage-based bottoms-up vs. the normal kind based on whether the config sets numberOfWinners to zero or sets a value for multiSeatBottomsUpPercentageThreshold.)

@moldover:
* Contest Info:
  * The hint: "fields do not influence computations..." isn't accurate as Dominion and Hart filter by contest ID, Clear Ballot filters by Contest Name, etc..
* CVR files:
  * Hint text needs re-working here to account for different providers
[...]

Issue #460, opened 2020-07-10
Support Hart CVR type in GUI


moldover:
Allow user to input Hart - preferably as a drop-down.
File browser should allow user to select a folder (not .json / .xlsx)

HEdingfield:
Closed with #466.

PR #462, opened 2020-07-13
Providers dropdown

HEdingfield:
Changes "Provider" field for CVR source to required, bases it on an enum of valid values, and changes it to a ChoiceBox at the start of the "CVR Files" tab in the GUI. Adds log message to help identify which CVR source validation errors pertain to. (Progress on #460 and #461)

Open questions:
* Do we anticipate any problems with making "Provider" a required field?
* If you have a config file with, e.g., "ES&s" it'll say invalid provider because it's not "ES&S". This is desired behavior, right?
* Do we care about re-ordering fields in the RawConfig? i.e. so provider is first in the CVR area?
* If you double click a cell to edit provider in the table do we care enough to try to make that an embedded dropdown?

### DIFF
--- a/src/main/java/network/brightspots/rcv/ContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/ContestConfig.java
@@ -29,6 +29,7 @@ import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
@@ -75,12 +76,17 @@ class ContestConfig {
   private static final int MAX_MULTI_SEAT_BOTTOMS_UP_PERCENTAGE_THRESHOLD = 100;
   private static final long MIN_RANDOM_SEED = -140737488355328L;
   private static final long MAX_RANDOM_SEED = 140737488355327L;
-  private static final String CDF_PROVIDER = "CDF";
-  private static final String HART_PROVIDER = "HART";
   private static final String JSON_EXTENSION = ".json";
   private static final String MAX_SKIPPED_RANKS_ALLOWED_UNLIMITED_OPTION = "unlimited";
   private static final String MAX_RANKINGS_ALLOWED_NUM_CANDIDATES_OPTION = "max";
   static final String SUGGESTED_MAX_RANKINGS_ALLOWED = MAX_RANKINGS_ALLOWED_NUM_CANDIDATES_OPTION;
+
+  static boolean isCdf(CvrSource source) {
+    return source.getProvider() != null
+        && Provider.getByLabel(source.getProvider()) == Provider.CDF
+        && source.getFilePath() != null
+        && source.getFilePath().toLowerCase().endsWith(JSON_EXTENSION);
+  }
 
   // underlying rawConfig object data
   final RawContestConfig rawConfig;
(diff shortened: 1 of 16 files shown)


src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -75,13 +76,18 @@
...
-  private static final String CDF_PROVIDER = "CDF";
-  private static final String HART_PROVIDER = "HART";
   private static final String JSON_EXTENSION = ".json";
   private static final String MAX_SKIPPED_RANKS_ALLOWED_UNLIMITED_OPTION = "unlimited";
   private static final String MAX_RANKINGS_ALLOWED_NUM_CANDIDATES_OPTION = "max";
   static final String SUGGESTED_MAX_RANKINGS_ALLOWED = MAX_RANKINGS_ALLOWED_NUM_CANDIDATES_OPTION;
 
+  static boolean isCdf(CvrSource source) {
+    return source.getProvider() != null
+        && Provider.getByLabel(source.getProvider()) == Provider.CDF
```
tarheel, line 86: Why not use the helper you created?

```
  && getProvider(source) == Provider.CDF
```

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -142,15 +148,73 @@ static ContestConfig loadContestConfig(String configPath) {
     return loadContestConfig(configPath, false);
   }
 
-  static boolean isCdf(CvrSource source) {
+  static boolean isHart(CvrSource source) {
     return source.getProvider() != null
-        && source.getProvider().toUpperCase().equals(CDF_PROVIDER)
-        && source.getFilePath() != null
-        && source.getFilePath().toLowerCase().endsWith(JSON_EXTENSION);
+        && Provider.getByLabel(source.getProvider()) == Provider.HART;
```
tarheel, line 153: ditto

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
```diff
@@ -55,6 +55,13 @@ Config file must be valid JSON format. Examples can be found in the test_data fo
     list of input cast vote record (CVR) file paths and their associated parameters
     each "cvrFileSources" list item contains the following parameters:
 
+      "provider" required
+        text description of the vendor / machine which generated this file
+        if set to "CDF" (and the filePath is ".json") the CVR file will be read as NIST Common Data Format (CDF)
+        example: "ES&S"
+        value: text string of length [1..1000]
+        if not supplied: file will be read as ES&S format
+
```
tarheel, line 64: Can we just explicitly list the accepted values here?

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
```diff
@@ -119,34 +119,29 @@
...
+                    <TextField prefHeight="25.0" prefWidth="396.0" promptText="File Path *"
+                      fx:id="textFieldCvrFilePath"/>
                   </HBox>
                   <HBox alignment="CENTER_LEFT" spacing="4.0">
+                    <TextField prefHeight="25.0" prefWidth="160.0" promptText="First Vote Col Index"
+                      fx:id="textFieldCvrFirstVoteCol"/>
+                    <TextField prefHeight="25.0" prefWidth="160.0" promptText="First Vote Row Index"
+                      fx:id="textFieldCvrFirstVoteRow"/>
+                    <TextField prefHeight="25.0" prefWidth="120.0" promptText="ID Col Index"
+                      fx:id="textFieldCvrIdCol"/>
```
tarheel, line 140: What are these changes for?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -823,6 +838,32 @@ private Integer stringToIntWithOption(String rawInput, String optionFlag, Intege
     return intValue;
   }
 
+  enum Provider {
```
tarheel, line 841: Do we need to define a generic CSV provider corresponding to the output from the Dominion conversion process? Even if it behaves the same as ES&S, it would be weird to call it ES&S when it's actually derived from a rival company's output.

PR #458, opened 2020-07-06
Hart

moldover:
Add Hart xml reading for #457 (The Hart CVR structure is defined in rcv/reference/Hart_Verity_CVR_Structure.pdf)
During CVR reading this code filters CVRs by contest name (instead of contest Id).
I had to refactor contest Id into a String (was Integer) which allows it to be used across Dominion and Hart (supports #456).
I added two Hart unit tests which support RCV: hart_cedar_park_school_board and hart_travis_county_officers

PR #463, opened 2020-07-13
make contestId a real config field

### LINKED ISSUE #456: Support For Multiple Contests
Dominion and Hart CVR data supports the notion of multiple contests.  To support tabulating these data sources we need to:
- add contest ID field to the config (validation logic should require it when ingesting Hart or Dominion, otherwise ignore)
- allow user to specify contest ID in the GUI.  
- during tabulation filter input RCV data sources by contest ID 

It's worth considering if the existing Contest Name field could simply be used as-is.

update call site at // TODO: use contest Id (blocked by #456)

See #438 Dominion tabulation

tarheel:
Fixes #456.

I confirmed that the Hart tests still pass.

I ignored Dominion for now, but if/when we add direct tabulation for that source provider, it should be easy to update my logic here to accommodate it.

### DIFF
--- a/src/main/java/network/brightspots/rcv/ContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/ContestConfig.java
@@ -447,6 +447,18 @@ class ContestConfig {
                 cvrPath);
           }
         }
+
+        if (isNullOrBlank(getContestId()) && isHart(source)) {
+          isValid = false;
+          Logger.log(
+              Level.SEVERE,
+              "contestId is required for Hart files.");
+        } else if (!isNullOrBlank(getContestId()) && !isHart(source)) {
+          isValid = false;
+          Logger.log(
+              Level.SEVERE,
+              "contestId may not be used with this type of CVR file.");
+        }
       }
     }
   }
--- a/src/main/java/network/brightspots/rcv/RawContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/RawContestConfig.java
@@ -46,11 +46,11 @@ public class RawContestConfig {
   public static class OutputSettings {
 
     public String contestName;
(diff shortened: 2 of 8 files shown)


src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -46,11 +46,11 @@
   public static class OutputSettings {
 
     public String contestName;
+    public String contestId;
```
HEdingfield, line 49: I'm kind of baffled why this field was sitting here all along, and just unused / not hooked up to the GUI before. Any historical context on this?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -447,6 +447,18 @@ private void validateCvrFileSources() {
                 cvrPath);
           }
         }
+
+        if (isNullOrBlank(getContestId()) && isHart(source)) {
+          isValid = false;
+          Logger.log(
+              Level.SEVERE,
+              "contestId is required for Hart files.");
+        } else if (!isNullOrBlank(getContestId()) && !isHart(source)) {
```
HEdingfield, line 456: Is there really a reason to enforce this? If so, please update the .txt file to note that it's not valid with anything other than a Hart input.

tarheel: I think it's good to prevent the user from entering a field that can't possibly be relevant. I'll update the .txt.

src/main/java/network/brightspots/rcv/RawContestConfig.java
tarheel: @moldover just added this in #458.

PR #466, opened 2020-07-19
Hart GUI

### LINKED ISSUE #460: Support Hart CVR type in GUI

HEdingfield:
The GUI provider `ComboBox` now dynamically renders subsequent fields and can handle input of Hart CVRs. Fixes #460 and progress on #461 (actually progress on GUI phase 2!).

### DIFF
--- a/src/main/java/network/brightspots/rcv/ContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/ContestConfig.java
@@ -154,58 +154,86 @@ class ContestConfig {
       sourceValid = false;
       Logger.log(Level.SEVERE, "filePath is required for each cast vote record file!");
     } else {
-      if (getProvider(source) == Provider.PROVIDER_UNKNOWN) {
+      Provider provider = getProvider(source);
+      if (provider == Provider.PROVIDER_UNKNOWN) {
         sourceValid = false;
-        Logger.log(Level.SEVERE, "Invalid provider!");
-      }
+        Logger.log(Level.SEVERE, "Invalid provider for source: %s", source.getFilePath());
+      } else if (provider == Provider.ESS) {
+        // ensure valid first vote column value
+        if (fieldOutOfRangeOrNotInteger(
+            source.getFirstVoteColumnIndex(),
+            "firstVoteColumnIndex",
+            MIN_COLUMN_INDEX,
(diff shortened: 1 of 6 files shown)


src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -288,6 +316,19 @@ private static boolean fieldOutOfRangeOrNotInteger(
     return !stringValid;
   }
 
+  private static boolean fieldIsDefinedButShouldNotBeForProvider(String value, String fieldName,
```
moldover, line 319: I can't help wondering if there's a more concise name for this like fieldIsValidForProvider

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -409,13 +414,37 @@ public void buttonClearDatePickerContestDateClicked() {
...
+        openFile = fc.showOpenDialog(GuiContext.getInstance().getMainWindow());
+        break;
+      }
+      case "Hart": {
+        DirectoryChooser dc = new DirectoryChooser();
+        dc.setInitialDirectory(new File(FileUtils.getUserDirectory()));
+        dc.setTitle("Select Hart Cast Vote Record Folder");
+        openFile = dc.showDialog(GuiContext.getInstance().getMainWindow());
+        break;
+      }
```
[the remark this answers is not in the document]
HEdingfield: I considered this, but unfortunately you can't use polymorphism for `FileChooser` and `DirectoryChooser` because they don't have a common class. I spose you could feed an enum or boolean in as a parameter to tell it to use one or the other though. You'd also need to parameterize both the `ExtensionFilter` and the title, so you're really not saving that many lines.

I say go ahead and go for it if you still think it makes sense in your Dominion PR once you've got it laid out there.

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: The only good alternative I could think of was `fieldUnnecessarilyDefinedForProvider`, which doesn't save that much space, so I'll go ahead and leave as-is.

PR #470, opened 2020-07-20
direct Dominion tabulation

### LINKED ISSUE #438: Direct tabulation from Dominion input
Per discussion in #408, it'd be nice to be able to tabulate Dominion inputs directly. This would likely do the conversion to generic .csv behind-the-scenes, and include those intermediary files as additional output.

Relevant discussion from Slack:

> @HEdingfield: 
> Isn't our end goal to make the Dominion conversion process kind of behind-the-scenes? So instead of creating a separate button / area in the GUI for it, wouldn't it be better to just make the normal load function handle Dominion files? Or is this too advanced / too much work?
> 
> @moldover 
> so you’re right that, we do want to support Dominion as a first-class data source (just as we do for ES&S)
> however, we expose the conversion process as a way to verify the conversion process itself.
> we’ll need to do the same for any data source which has non-trivial input logic.
> so to your question, ‘wouldn’t it be better to just support it as a data source’ …
> yes, that would be better.
> we could read from data sources AND export the converted csv files for verification
> best of both worlds
> 
> @HEdingfield 
> Couldn't that just include providing the converted file as another output along with the results?
> 
> @moldover 
> right
> but there’s a few more wrinkles to consider -
> Dominion data sources specify multiple contest Ids
> Dominion data sources specify multiple contests (by Id)
> and tabulation process is done against for a single contest at a time
> (yes, one could tabulate all the contests for an election “at once”, but in actual practice, that is not what election admins do.  at least according to Caleb.)
> so actually getting all that figured out in a reasonable way is I think, a much bigger chunk of work, than just… plugging in a new fileReader to a new data source
> but that said, I do think we should do it
> so to recap: the “convert cvr” functionality is needed on it’s own right.
> 
> @tarheel 
> so if we wanted to support Dominion as a regular input, we would need the user to specify a contest ID param in the config also
> 
> @moldover 
> something like that, yes
> the dem primaires only have one contest, but take SF - it has like 20 or 30 contests
> Maybe it’s not that much work?

tarheel:
Mostly fixes #438.

As expected, the code changes required to enable direct Dominion tabulation were very minor.

I also decided to start using "Undeclared Write-ins" as the label for UWIs in all output files. Without this, we had no good way to replace the UWI candidate code in Dominion and Hart files (which is always an indecipherable ID like "99cefa84-bad0-4d6b-a916-e606bd43f9cd" or the number 12) to a human-readable string. So that necessitated updating a bunch of test files.

One note: unlike with other source types, there's currently nothing enforcing that all of the candidates found in the Dominion CVR are also specified in the JSON config file. If you omit a candidate, you just get a NullPointerException when you tabulate. I (or someone else) should add this check in a follow-up PR.

Oh, and I actually used the new Java 14 switch statement syntax! It truly is life-changing.


src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -269,19 +269,24 @@ private boolean setUpLogging(ContestConfig config) {
...
+        } else if (provider == Provider.ESS) {
           Logger.log(Level.INFO, "Reading ES&S cast vote record file: %s...", cvrPath);
           new StreamingCvrReader(config, source).parseCvrFile(castVoteRecords, precinctIds);
           continue;
-        } else if (ContestConfig.getProvider(source) == Provider.HART) {
-          HartCvrReader reader = new HartCvrReader(cvrPath, config);
+        } else if (provider == Provider.HART) {
           Logger.log(Level.INFO, "Reading Hart cast vote records from folder: %s...", cvrPath);
+          HartCvrReader reader = new HartCvrReader(cvrPath, config);
           reader.readCastVoteRecordsFromFolder(castVoteRecords);
```
HEdingfield, line 290: Nit: mirror the others by merging these two lines together (since you don't need to use `reader` anywhere else).

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -269,19 +269,24 @@ private boolean setUpLogging(ContestConfig config) {
...
       try {
         if (ContestConfig.isCdf(source)) {
           Logger.log(Level.INFO, "Reading CDF cast vote record file: %s...", cvrPath);
           CommonDataFormatReader reader = new CommonDataFormatReader(cvrPath, config);
           reader.parseCvrFile(castVoteRecords);
           continue;
-        } else if (ContestConfig.getProvider(source) == Provider.ESS) {
+        } else if (provider == Provider.DOMINION) {
+          Logger.log(Level.INFO, "Reading Dominion cast vote records from folder: %s...", cvrPath);
+          new DominionCvrReader(cvrPath).readCastVoteRecords(castVoteRecords);
```
HEdingfield, line 281: So this seems to work great, but I don't think it's creating the intermediary CVR .csv file (e.g. like `dominion_cvr_conversion_alaska_contest_1_expected.csv`). Wasn't this one of our requirements? I imagine we'd want it to appear in the output folder too.

tarheel:
Addressed @HEdingfield's comments.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -507,9 +507,16 @@ private void validateCvrFileSources() {
...
+        } else if (
+            !(provider == Provider.DOMINION || provider == Provider.HART) &&
+                fieldIsDefinedButShouldNotBeForProvider(
+                    getContestId(),
+                    "contestId",
+                    provider,
+                    source.getFilePath())
+        ) {
           isValid = false;
           Logger.log(
```
HEdingfield, line 521: Please delete this log message since it's already covered by `fieldIsDefinedButShouldNotBeForProvider`.

src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -155,6 +155,15 @@ private static void runTabulationTest(String stem) {
       compareJsons(config, stem, timestampString, null);
     }
 
+    // If this is a Dominion tabulation test, also check the converted output file.
+    boolean isDominion = config.rawConfig.cvrFileSources.stream().anyMatch(source ->
```
HEdingfield, line 159: Yer a wizard, Harry.

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -496,13 +496,18 @@ void generateOverallSummaryFiles(
...
       Collection<Contest> contests,
-      String csvOutputFolder)
-      throws IOException {
+      String csvOutputFolder,
+      String contestId
+  ) throws IOException {
     List<String> filesWritten = new ArrayList<>();
     try {
       for (Contest contest : contests) {
+        if (!isNullOrBlank(contestId) && !contest.getId().equals(contestId)) {
```
HEdingfield, line 508: I don't understand the idea behind this and would feel more comfortable if @moldover reviewed as well.

tarheel: Oh, it's just that the Dominion source may contain multiple contests, but when we're doing a tabulation, we only care about one of them (indicated by contestId), so I'm skipping all the others.

src/main/java/network/brightspots/rcv/ContestConfig.java
tarheel: oops, right

src/main/java/network/brightspots/rcv/ResultsWriter.java
moldover: @tarheel This is correct logic, but it does point out a (easy to fix) problem: the DominionCvrReader should be filtering on contest Id when the CastVoteRecords are first read.  Then, no additional filtering is needed, either here, or during tabulation.  Do you want me to fix that in your branch?  Or it's simple enough to add around line 237 where we are already validating the contest id is defined in the 'contests' file.

PR #474, opened 2020-07-23
Changes `contestId` so it's specified on a per-CVR basis (fixes #456)

### LINKED ISSUE #456: Support For Multiple Contests

HEdingfield:


src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -796,18 +817,27 @@ public LocalDate fromString(String string) {
...
-        textFieldCvrFirstVoteCol.setDisable(false);
-        textFieldCvrFirstVoteRow.setDisable(false);
-        textFieldCvrIdCol.setDisable(false);
-        textFieldCvrPrecinctCol.setDisable(false);
-      } else if (provider.equals("CDF") || provider.equals("Dominion") || provider.equals("Hart")) {
-        buttonAddCvrFile.setDisable(false);
-        textFieldCvrFilePath.setDisable(false);
-        buttonCvrFilePath.setDisable(false);
+      switch (provider) {
+        case "ES&S" -> {
```
tarheel, line 821: image

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -796,18 +817,27 @@ public LocalDate fromString(String string) {
...
+          buttonCvrFilePath.setDisable(false);
+          textFieldCvrFirstVoteCol.setDisable(false);
+          textFieldCvrFirstVoteRow.setDisable(false);
+          textFieldCvrIdCol.setDisable(false);
+          textFieldCvrPrecinctCol.setDisable(false);
+        }
+        case "CDF" -> {
+          buttonAddCvrFile.setDisable(false);
+          textFieldCvrFilePath.setDisable(false);
+          buttonCvrFilePath.setDisable(false);
```
tarheel, line 833: These three are common to all three cases, so you could just do them outside of the switch.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Haaaaaaaaa!

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: True, I thought about this, but kept it because it causes an unimplemented provider to not affect the GUI, which seems desirable. Plus it forces us to explicitly lay out what we expect to happen for each provider.

Issue #400, opened 2019-12-02
support Clear Ballot format


tarheel:
[Clear_Ballot_BEN_QA_Gen_Res_WITTesting1.cvr.zip](https://github.com/BrightSpots/rcv/files/3909283/Clear_Ballot_BEN_QA_Gen_Res_WITTesting1.cvr.zip)

Louis's notes from the sample:
- Uses a separate CSV file for each different entity type (contests, parties, choices, precincts, and CVRs).
- Is this an example of an RCV contest? I can't find any rankings, so I'm not sure I understand how we would parse this.

tarheel:
More questions from @moldover:

How are entity files named?
Will we need to add config parameteres to specify these files?

Do we need to apply any logic to determine precincts, as multiple precinct Ids (from precincts.csv) seem to map to a single precinct:
 "Precinct 6 (Generic Use with CounterGroup EV) Ballot, Precinct 6 (Generic Use with CounterGroup EV) Standa

What do the different "choices" from choices.csv mean?

csv Column headers look like: "Choice_11_1:Contest 1 (Candidate):Vote For 1:Write-in:Non-Partisan"  How do these encode rankings?
The Choice id seems to be the in the first part.  
Do we need to apply logic to choice Ids?  To determine a ranking?  There are ALL KINDS of interesting choices one can reference.

Do we need to handle multiple contests?  How?
 - User specified contest ID?
 - Multiple contests from a single config?

HEdingfield:
This is now implemented in the GUI. @moldover Any last things we need to address (e.g. from the above comments) or can we safely close this now?

moldover:
closed via #475

PR #475, opened 2020-07-24
First cut at Clear Ballot rcv reader

moldover:
Add ClearBallorCvrReader
Add Kansas Dem Primary regression test
Add Clear Ballot reference files
Partially Addresses #400


src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -280,6 +280,11 @@ private boolean setUpLogging(ContestConfig config) {
           Logger.log(Level.INFO, "Reading CDF cast vote record file: %s...", cvrPath);
           new CommonDataFormatReader(cvrPath, config).parseCvrFile(castVoteRecords);
           continue;
+        } else if (ContestConfig.getProvider(source) == Provider.CLEAR_BALLOT) {
+          ClearBallotCvrReader reader = new ClearBallotCvrReader(cvrPath, config);
+          Logger.log(Level.INFO, "Reading Clear Ballot cast vote records from folder: %s...", cvrPath);
+          reader.readCastVoteRecords(castVoteRecords);
```
HEdingfield, line 286: Here you can just do:
```
new ClearBallotCvrReader(cvrPath, config).readCastVoteRecords(castVoteRecords);
```
No need to defined reader separately above.

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -280,6 +280,11 @@ private boolean setUpLogging(ContestConfig config) {
           Logger.log(Level.INFO, "Reading CDF cast vote record file: %s...", cvrPath);
           new CommonDataFormatReader(cvrPath, config).parseCvrFile(castVoteRecords);
           continue;
+        } else if (ContestConfig.getProvider(source) == Provider.CLEAR_BALLOT) {
```
HEdingfield, line 283: Did you merge in the latest in develop? I might be wrong, but it seems like @tarheel's latest changes may not have been captured. Please go ahead and do that if you haven't already.

src/main/java/network/brightspots/rcv/ClearBallotCvrReader.java
```diff
@@ -0,0 +1,144 @@
...
+        ArrayList<Pair<Integer, String>> rankings = new ArrayList<>();
+        for (int columnIndex = CvrColumnField.ChoicesBegin.ordinal(); columnIndex < cvrData.length; columnIndex++) {
+          if (Integer.parseInt(cvrData[columnIndex]) == 1) {
+            // user marked this column
+            Pair<Integer, String> ranking = columnIndexToRanking.get(columnIndex);
+            rankings.add(ranking);
+          }
+        }
+        // create the cast vote record
+        CastVoteRecord castVoteRecord = new CastVoteRecord(this.contestConfig.getContestId(),
```
HEdingfield, line 95: Nit: There's a comment in CastVoteRecord that says:
```
// Dominion and Hart use this constructor
```
Please update this or probably better yet delete it (and the one about CDF and ES&S on the constructor below it).

src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -214,6 +214,12 @@ static void setup() {
     Logger.setup();
   }
 
+  @Test
+  @DisplayName("Clear Ballot - Kansas Primary")
+  void testClearBallotKansasPrimary() {
```
HEdingfield, line 219: This fails for me... does it work on your end?
```
2020-07-23 23:26:27 PDT INFO: Starting tabulation session...
2020-07-23 23:26:27 PDT INFO: Successfully loaded contest config: C:\Users\UserName\IdeaProjects\rcv\src\test\resources\network\brightspots\rcv\test_data\clear_ballot_kansas_primary\clear_ballot_kansas_primary_config.json
2020-07-23 23:26:27 PDT INFO: Validating contest config...
2020-07-23 23:26:27 PDT SEVERE: Invalid provider for source: clear_ballot_kansas_primary.cvr.csv
2020-07-23 23:26:27 PDT SEVERE: contestId should not be defined for CVR source with provider "Provider unknown": clear_ballot_kansas_primary.cvr.csv
2020-07-23 23:26:27 PDT SEVERE: Contest config validation failed! Please modify the contest config file and try again.
See config_file_documentation.txt for more details.
```

src/test/resources/network/brightspots/rcv/test_data/clear_ballot_kansas_primary/clear_ballot_kansas_primary_config.json
```diff
@@ -0,0 +1,83 @@
...
+    "tabulateByPrecinct" : false,
+    "generateCdfJson" : false
+  },
+  "cvrFileSources" : [ {
+    "filePath" : "clear_ballot_kansas_primary.cvr.csv",
+    "firstVoteColumnIndex" : "",
+    "firstVoteRowIndex" : "",
+    "idColumnIndex" : "",
+    "precinctColumnIndex" : "",
+    "provider" : "Clear Vote"
```
HEdingfield, line 19: I think this is why the test is failing; should be: Clear Ballot

src/main/java/network/brightspots/rcv/ClearBallotCvrReader.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.logging.Level;
+import javafx.util.Pair;
+
+import network.brightspots.rcv.CastVoteRecord.CvrParseException;
+
+
+class ClearBallotCvrReader {
```
HEdingfield, line 33: Hoping @tarheel can review this class.

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -280,6 +280,11 @@ private boolean setUpLogging(ContestConfig config) {
...
+          ClearBallotCvrReader reader = new ClearBallotCvrReader(cvrPath, config);
+          Logger.log(Level.INFO, "Reading Clear Ballot cast vote records from folder: %s...", cvrPath);
```
HEdingfield, line 285: It looks like Clear Ballot is (blessedly) a single .csv file, so why do we specify a folder instead of a specific file like for ES&S? The latter would seem to make more sense to me...

src/main/java/network/brightspots/rcv/TabulatorSession.java
HEdingfield: Just merged in #474, so that'll need to be accounted for.

src/main/java/network/brightspots/rcv/ClearBallotCvrReader.java
```diff
@@ -0,0 +1,144 @@
...
+      // each "choice column" in the input Csv corresponds to a particular ranking: candidate+rank pair
+      // we parse these rankings from the header row into a map for lookup during CVR parsing
+      String firstRow = csvReader.readLine();
+      if (firstRow == null) {
+        Logger.log(Level.SEVERE, "No header row found in cast vote record file: %s", this.cvrPath);
+        throw new CvrParseException();
+      }
+      String[] headerData = firstRow.split(",");
+      if (headerData.length < CvrColumnField.ChoicesBegin.ordinal()) {
+        Logger.log(Level.SEVERE, "No choice columns found in cast vote record file: %s", this.cvrPath);
```
tarheel, line 58: should we throw or otherwise return an error here instead of just logging?

src/main/java/network/brightspots/rcv/ClearBallotCvrReader.java
```diff
@@ -0,0 +1,144 @@
...
+        if (!contestName.equals(this.contestConfig.getContestId())) {
+          continue;
+        }
+        // validate and store the ranking associated with this choice column
+        String choiceName = choiceFields[RcvChoiceHeaderField.CHOICE_NAME.ordinal()];
+        if (!this.contestConfig.getCandidateCodeList().contains(choiceName)) {
+          Logger.log(Level.SEVERE, "Candidate: %s from cast vote record not found in config!", choiceName);
+          throw new CvrParseException();
+        }
+        Integer rank = Integer.parseInt(choiceFields[RcvChoiceHeaderField.RANK.ordinal()]);
```
tarheel, line 75: we should validate that this doesn't exceed the maxRankingsAllowed value set in the config

src/main/java/network/brightspots/rcv/ClearBallotCvrReader.java
```diff
@@ -0,0 +1,144 @@
...
+      // read all remaining rows and create CastVoteRecords for each one
+      while (true) {
+        String row = csvReader.readLine();
+        if (row == null) {
+          break;
+        }
+        // parse rankings
+        String[] cvrData = row.split(",");
+        ArrayList<Pair<Integer, String>> rankings = new ArrayList<>();
+        for (int columnIndex = CvrColumnField.ChoicesBegin.ordinal(); columnIndex < cvrData.length; columnIndex++) {
```
tarheel, line 87: I might be missing something, but I think you need to skip the columns that aren't for the desired contest when you're going through this loop also. Otherwise, when you encounter a column for another contest, you'll do a lookup in `columnIndexToRanking` with a key that it doesn't have, and end up putting a null ranking in your rankings list.

src/main/java/network/brightspots/rcv/ClearBallotCvrReader.java
```diff
@@ -0,0 +1,144 @@
...
+          if (Integer.parseInt(cvrData[columnIndex]) == 1) {
+            // user marked this column
+            Pair<Integer, String> ranking = columnIndexToRanking.get(columnIndex);
+            rankings.add(ranking);
+          }
+        }
+        // create the cast vote record
+        CastVoteRecord castVoteRecord = new CastVoteRecord(this.contestConfig.getContestId(),
+            cvrData[CvrColumnField.ScanComputerName.ordinal()],
```
tarheel, line 96: This corresponds to `tabulatorId` and not `batchId`?

src/main/java/network/brightspots/rcv/ClearBallotCvrReader.java
moldover: Yeah good idea.

src/main/java/network/brightspots/rcv/ClearBallotCvrReader.java
moldover: Good catch.  Updated to use the rankings keySet instead.

src/main/java/network/brightspots/rcv/ClearBallotCvrReader.java
```diff
@@ -0,0 +1,153 @@
...
+            this.cvrPath);
+        throw new CvrParseException();
+      }
+      Map<Integer, Pair<Integer, String>> columnIndexToRanking = new HashMap<>();
+      for (int columnIndex = CvrColumnField.ChoicesBegin.ordinal(); columnIndex < headerData.length;
+          columnIndex++) {
+        String choiceColumnHeader = headerData[columnIndex];
+        String[] choiceFields = choiceColumnHeader.split(":");
+        // filter by contest
+        String contestName = choiceFields[RcvChoiceHeaderField.CONTEST_NAME.ordinal()];
```
tarheel, line 68: If their column headers are malformed, we could get an IndexOutOfBounds here.

src/main/java/network/brightspots/rcv/ClearBallotCvrReader.java
moldover: Yeah, batch usually refers to a group of ballots, while tabulator in this context usually means the ballot counting device which accumulates / aggregates ballots from a polling place.  This doesn't impact the contest tabulation of course, but we should at some point figure out how to reconcile data like this which varies between different vendors in a more comprehensive way.

PR #487, opened 2020-08-03
remove Dominion conversion feature

### LINKED ISSUE #476: Remove "Convert Dominion to Generic Format..." functionality
Because these intermediary human-readable CVR .csv files are now included as part of the tabulation output, there's probably no reason to keep this separate option around (verified that this idea is supported by @tarheel, @moldover, and @chughes297).

Note that we'll have to remove it from both the GUI (might want to get rid of the whole "Conversion" menu... should discuss this) and the CLI (as well as `README.md`).

Also, this comment in `DominionCvrReader` should be worded differently, and we should make sure `null` should no longer be acceptable as a parameter for `contestId`.

```
  // parse Cvr json into CastVoteRecord objects and add them to the input list
  // (If contestId is specified, we'll only load CVRs for that contest.)
  void readCastVoteRecords(List<CastVoteRecord> castVoteRecords, String contestId)
```

tarheel:
Fixes #476


src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -291,9 +295,9 @@ private void parseCvrFile(String filePath, List<CastVoteRecord> castVoteRecords,
   // Candidate data from a Dominion candidate manifest Json
   static class Candidate {
 
-    private String name;
-    private String code;
-    private String contestId;
+    private final String name;
+    private final String code;
+    private final String contestId;
```
tarheel, line 300: IntelliJ did this, for the record.

tarheel:
OK, I rebased.

We may want to rearrange the menu items now that Conversion only contains one item, but let's not try to deal with that in this PR.

src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -123,10 +123,14 @@
     return contests;
   }
 
-  // parse Cvr json into CastVoteRecord objects and add them to the input list
-  // (If contestId is specified, we'll only load CVRs for that contest.)
+  // parse CVR JSON for records matching the specified contestId into CastVoteRecord objects and add
+  // them to the input list
   void readCastVoteRecords(List<CastVoteRecord> castVoteRecords, String contestId)
       throws CvrParseException, UnrecognizedCandidatesException {
+    if (contestId == null) {
```
HEdingfield, line 130: This already covered in `passesBasicCvrSourceValidation()` so I don't think there should be any need to do it here (which I suppose means you'd revert basically everything in this file).

tarheel: I meant this more like an assertion (guarding against a coding error, not a user error), but yeah, I guess we don't really do that elsewhere, so I'll remove it. My other changes in this file still make sense, though.

PR #488, opened 2020-08-05
throw error when candidates are missing from Dominion config

### LINKED ISSUE #473: raise error if we encounter an unrecognized candidate while loading Dominion CVRs during direct tabulation
Just makes sure the config file is in sync with the data, as we already do for other providers.

tarheel:
Just mimicking the logic from `StreamingCvrReader`.

Fixes #473.


tarheel:
> Do we need to do this for Hart or any other providers as well?

For Clear Ballot and Hart, @moldover added similar logic. The difference is that instead of counting the number of unrecognized candidate instances, it just throws an exception as soon as it sees one. It would be nice to be consistent with how we handle this, but it's not a big deal. I'll create a separate (non-urgent) issue.

PR #492, opened 2020-08-06
GUI redesign phase 1

HEdingfield:
I figure it's time to check in after working on this for over a week... it gets a lot done on #461, but there's still quite a bit left to do.


src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -78,6 +80,47 @@
...
   private static final String CONFIG_FILE_DOCUMENTATION_FILENAME =
       "network/brightspots/rcv/config_file_documentation.txt";
+  private static final List WINNER_ELECTION_MODE_VALUES = List.of(
+      "Single-winner majority determines winner",
+      "Multi-winner allow only one winner per round",
+      "Multi-winner allow multiple winners per round",
+      "Bottoms-up",
+      "Bottoms-up using percentage threshold",
+      "Multi-pass IRV"
+  );
```
tarheel, line 90: It's unfortunate that we're defining these strings here and then repeating them for the map definition. Can't we just use the map definition and then pull these strings out when we need them?

Same comment for tiebreak modes.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -203,6 +248,30 @@ private static String getTextOrEmptyString(TextField textField) {
...
+      case "multiSeatAllowOnlyOneWinnerPerRound" -> {
+        return "Multi-winner allow only one winner per round";
+      }
+      case "multiSeatBottomsUp" -> {
+        return config.getNumberOfWinners() == 0
+            || config.getMultiSeatBottomsUpPercentageThreshold() != null
+            ? "Bottoms-up using percentage threshold" : "Bottoms-up";
+      }
+      case "multiSeatSequentialWinnerTakesAll" -> {
+        return "Multi-pass IRV";
```
tarheel, line 269: can we define constants for these human-readable strings?

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -203,6 +248,30 @@ private static String getTextOrEmptyString(TextField textField) {
...
+      }
+      case "singleSeatContinueUntilTwoCandidatesRemain" -> {
+        return "Single-winner majority determines winner";
+      }
+      case "multiSeatAllowOnlyOneWinnerPerRound" -> {
+        return "Multi-winner allow only one winner per round";
+      }
+      case "multiSeatBottomsUp" -> {
+        return config.getNumberOfWinners() == 0
+            || config.getMultiSeatBottomsUpPercentageThreshold() != null
```
tarheel, line 265: if getMultiSeatBottomsUpPercentageThreshold() is not null and getNumberOfWinners is non-zero, something is wrong

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -203,6 +248,30 @@ private static String getTextOrEmptyString(TextField textField) {
     return textField.getText() != null ? textField.getText().trim() : "";
   }
 
+  private static String convertConfigWinnerElectionModeToGuiText(ContestConfig config) {
```
tarheel, line 251: I have a general concern about this approach. Up until this point we've been successful at keeping a tight alignment between the raw config file and the UI, which I think has been really helpful for the maintainability of our code. In this PR you're introducing various pieces of tricky logic to map from the config to the UI and vice versa because they don't represent information in the same way. If we want to make these UI changes, why don't we just update the config file schema accordingly? We can add simple migration logic to convert from an old file to a current one in order to preserve backward compatibility. This might be a bit more work in the short term, but I think it will save us pain in the long run.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -882,12 +994,54 @@ public LocalDate fromString(String string) {
...
 
-    choiceTiebreakMode.getItems().addAll(TieBreakMode.values());
-    choiceTiebreakMode.getItems().remove(TieBreakMode.MODE_UNKNOWN);
+    clearAndDisableWinningRuleFields();
+    choiceTiebreakMode.getItems().addAll(TIEBREAK_MODE_VALUES);
+    choiceTiebreakMode.setOnAction(event -> {
+      clearAndDisableTiebreakFields();
+      String mode = getChoiceElse(choiceTiebreakMode, TieBreakMode.MODE_UNKNOWN);
+      switch (mode) {
+        case "Random", "Previous round counts (then random)", "Generate permutation" -> textFieldRandomSeed
```
tarheel, line 1003: Here and below -- same concern I raised above about having the same string literals repeated several times in the code. We should pull these out into constants.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Agree, I wish we could keep it 1:1, but it just doesn't seem feasible unless we revamp not only the config file, but the logic that parses it. This new list of winner election modes both includes new entries (which are based not only on `winnerElectionMode` but also how other parameters are set), and gets rid of an entry. Please [review this comment](https://github.com/BrightSpots/rcv/issues/461#issuecomment-668365720) and let me know how we could update the config file schema without making serious changes to the other back-end logic (changes that I'd argue are better for a major version increase: 2.0.0, rather than 1.2.0.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Indeed, keep in mind that this logic is for loading a config file, which could indeed be configured incorrectly. Subsequent validation would catch it if so. This logic is just to best decide what should appear in the Winner Election Mode text box based off what it loads. Given that, do you agree that this handling makes sense?

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Totally agree. I tried with just the map at first, but the keys were unsorted so they appeared in a wacky order. Then I tried a sorted map, but there's apparently still no immutable version in vanilla Java. This was the somewhat unhappy compromise I made without needing to bring another library like Guava in. I do think it'll be improved a little by factoring the string literals out at least. The things I do for defensive programming...

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Absolutely, I thought of this immediately after I sent the PR.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Yep yep, will do.

tarheel:
I just added a commit that fixes the winner election mode logic.

One comment: I noticed that the config files now contain the human-readable strings for winner election mode and tiebreak mode. This feels a little unusual and error-prone. Can we keep it so that the files have the camelCasedValues even though we're now showing friendlier strings in the UI?

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -882,12 +934,57 @@ public LocalDate fromString(String string) {
...
+    });
     choiceOvervoteRule.getItems().addAll(OvervoteRule.values());
     choiceOvervoteRule.getItems().remove(OvervoteRule.RULE_UNKNOWN);
     choiceWinnerElectionMode.getItems().addAll(WinnerElectionMode.values());
     choiceWinnerElectionMode.getItems().remove(WinnerElectionMode.MODE_UNKNOWN);
+    choiceWinnerElectionMode.setOnAction(event -> {
+      clearAndDisableWinningRuleFields();
+      setWinningRulesDefaultValues();
+      String mode = getChoiceElse(choiceWinnerElectionMode, WinnerElectionMode.MODE_UNKNOWN);
+      switch (mode) {
```
moldover, line 956: Would it be preferable to switch on the actual enum and avoid using these hard-coded constants here?  I realize this would mean updating getChoiceElse, but this feels icky.

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -1098,13 +1105,13 @@ public String toString() {
...
-    RANDOM("random"),
-    INTERACTIVE("interactive"),
-    PREVIOUS_ROUND_COUNTS_THEN_RANDOM("previousRoundCountsThenRandom"),
-    PREVIOUS_ROUND_COUNTS_THEN_INTERACTIVE("previousRoundCountsThenInteractive"),
-    USE_PERMUTATION_IN_CONFIG("usePermutationInConfig"),
-    GENERATE_PERMUTATION("generatePermutation"),
-    MODE_UNKNOWN("modeUnknown");
+    RANDOM("Random"),
+    INTERACTIVE("Stop counting and ask"),
+    PREVIOUS_ROUND_COUNTS_THEN_RANDOM("Previous round counts (then random)"),
```
moldover, line 1110:
```suggestion
    PREVIOUS_ROUND_COUNTS_THEN_RANDOM("Previous round tallies then random"),
```

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -1098,13 +1105,13 @@ public String toString() {
...
   enum TieBreakMode {
-    RANDOM("random"),
-    INTERACTIVE("interactive"),
-    PREVIOUS_ROUND_COUNTS_THEN_RANDOM("previousRoundCountsThenRandom"),
-    PREVIOUS_ROUND_COUNTS_THEN_INTERACTIVE("previousRoundCountsThenInteractive"),
-    USE_PERMUTATION_IN_CONFIG("usePermutationInConfig"),
-    GENERATE_PERMUTATION("generatePermutation"),
-    MODE_UNKNOWN("modeUnknown");
+    RANDOM("Random"),
+    INTERACTIVE("Stop counting and ask"),
```
moldover, line 1109:
```suggestion
    INTERACTIVE("Pause counting and ask"),
```

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -1098,13 +1105,13 @@ public String toString() {
...
+    PREVIOUS_ROUND_COUNTS_THEN_RANDOM("Previous round counts (then random)"),
+    PREVIOUS_ROUND_COUNTS_THEN_INTERACTIVE("Previous round counts (then stop counting and ask)"),
```
moldover, line 1111:
```suggestion
    PREVIOUS_ROUND_COUNTS_THEN_INTERACTIVE("Previous round tallies then pause counting and ask"),
```

src/main/java/network/brightspots/rcv/Tabulator.java
HEdingfield: Per our conversation, will leave all labels as the UX person recommended for now.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Done!

PR #496, opened 2020-08-11
handle unrecognized candidates in CDF

### LINKED ISSUE #453: incorrect overvote label in CDF leads to NPE

tarheel:
Fixes #453 and related problems. We now handle this situation more gracefully.

### DIFF
--- a/src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
+++ b/src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
@@ -29,11 +29,13 @@ import java.util.List;
 import java.util.Map;
 import java.util.logging.Level;
 import javafx.util.Pair;
+import network.brightspots.rcv.TabulatorSession.UnrecognizedCandidatesException;
 
 class CommonDataFormatReader {
 
   private final String filePath;
   private final ContestConfig config;
+  private Map<String, Integer> unrecognizedCandidateCounts = new HashMap<>();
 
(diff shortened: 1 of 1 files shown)


src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -114,13 +119,15 @@
...
   // parse the given file into a List of CastVoteRecords for tabulation
-  void parseCvrFile(List<CastVoteRecord> castVoteRecords) {
+  void parseCvrFile(List<CastVoteRecord> castVoteRecords) throws UnrecognizedCandidatesException {
     // cvrIndex and fileName are used to generate IDs for cvrs
     int cvrIndex = 0;
     String fileName = new File(filePath).getName();
 
     try {
       HashMap json = JsonParser.readFromFile(filePath, HashMap.class);
+      Map<String, String> candidates = getCandidates();
```
HEdingfield, line 129: Nit: you can just inline this.

tarheel: That would be very inefficient -- it would mean that we were reloading the whole CVR to get the candidates again for each ballot that we process.

HEdingfield: Ah, right you are, I missed the loop due to GitHub review line hiding.

Issue #402, opened 2019-12-02
support Unisyn format


tarheel:
[RCV_WriteIns_1_UNIOVS22585_cvr_1_of_1.zip](https://github.com/BrightSpots/rcv/files/3909287/RCV_WriteIns_1_UNIOVS22585_cvr_1_of_1.zip)

Louis's notes:
- Since it uses the CDF, it should be fairly easy for us to handle, I hope.
- The sample you shared is strange, though, as each ranking is represented as a separate CVR entity... which seems wrong. Based on all the work we did earlier with the CDF, I'm fairly confident that the proper way to represent multiple rankings is to have multiple CVRContestSelection entities (one per ranking) inside the CVRContest, which is inside CVRSnapshot, which is inside CVR.

HEdingfield:
Unisyn output is supposedly equivalent to .xml CDF support, so we should be able to build that in and just expand the CDF provider in the GUI to be able to accept .xml files in addition to .json files.

moldover:
This is ostensibly closed as we can now support reading and tabulating example CDF XML CVR reports from Unisyn.  Woof.

PR #504, opened 2020-08-31
Read and Tabulate Unisyn Xml Cdf

moldover:
This is for #402.  I added an XML CDF parser of the CDF UML spec, as described in:
[https://github.com/BrightSpots/rcv/blob/master/reference/NIST-CommonDataFormat.png](url)
[https://github.com/BrightSpots/rcv/blob/master/reference/NIST%201500-103%20CDF%20Draft%202019-02-08.pdf](url) With additional documentation on XML naming conventions from here: [https://github.com/HiltonRoscoe/CDFPrototype/blob/master/mapping.md#mapping-common-data-formats](url)

Currently I have 2 Unisyn regression tests and will be adding a few more this week.  I wanted to get the ball rolling on the review now, since the core parsing code in CommonDataFormatReader.java now has all the pieces needed to complete a tabulation :)


src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -118,12 +198,204 @@
...
+                    .equals(STATUS_NO)) {
+                  continue;
+                }
+                // skip if not allocable
+                if (selectionPosition.IsAllocable.equals(STATUS_NO)) {
+                  continue;
+                }
+                if (selectionPosition.Rank == null) {
+                  Logger
+                      .severe("No Rank found on CVR %s Contest %s!", cvr.UniqueId,
```
HEdingfield, line 337: Nit: make consistent with other messages by putting quotes around both %s instances.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -118,12 +198,204 @@
...
+    // find current snapshot
+    CVRSnapshot currentSnapshot = null;
+    for (CVRSnapshot cvrSnapshot : cvr.CVRSnapshot) {
+      if (cvrSnapshot.ObjectId.equals(cvr.CurrentSnapshotId)) {
+        currentSnapshot = cvrSnapshot;
+        break;
+      }
+    }
+    // find CVRContest which matches the Contest we are tabulating
+    for (CVRContest cvrContest : currentSnapshot.CVRContest) {
```
HEdingfield, line 213: This can throw an NPE, so should handle safely with an assert, if statement, etc.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -22,28 +22,100 @@
...
+  private final String STATUS_YES = "yes";
+  private final String STATUS_NO = "no";
+  private final String STATUS_UNKNOWN = "unknown";
+  private final String BOOLEAN_TRUE = "true";
+  private final String BOOLEAN_FALSE = "false";
+
   private final String filePath;
   private final ContestConfig config;
+  private final CvrSource source;
   private Map<String, Integer> unrecognizedCandidateCounts = new HashMap<>();
```
HEdingfield, line 53: final

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -118,12 +198,204 @@
...
+    for (CVRContest cvrContest : currentSnapshot.CVRContest) {
+      if (cvrContest.ContestId.equals(contestToTabulate.ObjectId)) {
+        cvrContestToTabulate = cvrContest;
+        break;
+      }
+    }
+    return cvrContestToTabulate;
+  }
+
+  void parseXML(List<CastVoteRecord> castVoteRecords) {
```
HEdingfield, line 222: Nit: parseXml

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -22,28 +22,100 @@
...
 
-  CommonDataFormatReader(String filePath, ContestConfig config) {
+  CommonDataFormatReader(String filePath, ContestConfig config, CvrSource source) {
     this.filePath = filePath;
     this.config = config;
+    this.source = source;
+  }
+
+  // This method will extract candidate data from a CDF file for the contestID specified in
+  // our CvrSource.  It is currently un-used but will be handy for automating
```
HEdingfield, line 62: This comment seems out-of-date because this method _is_ used.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -22,28 +22,100 @@
...
+  // config file creation when we are ready to implement that.
+  Map<String, String> getCandidates() throws CvrParseException {
+    Map<String, String> candidates;
+    if (filePath.endsWith(".xml")) {
+      candidates = getCandidatesXml();
+    } else if (filePath.endsWith(".json")) {
+      candidates = getCandidatesJson();
+    } else {
+      Logger.log(Level.SEVERE,
+          "Unexpected file extension: %s.  CDF source files must be .xml or .json", this.filePath);
```
HEdingfield, line 72: I think we can and should do this check as part of the validation instead of here and in `parseCvrFile`. Recommend removing it from both and putting it in `ContestConfig` where the "// perform CDF checks" comment is.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -22,28 +22,100 @@
...
 import network.brightspots.rcv.TabulatorSession.UnrecognizedCandidatesException;
 
 class CommonDataFormatReader {
 
+  private final String STATUS_NEEDS_ADJUDICATION = "needs-adjudication";
+  private final String STATUS_YES = "yes";
+  private final String STATUS_NO = "no";
+  private final String STATUS_UNKNOWN = "unknown";
+  private final String BOOLEAN_TRUE = "true";
+  private final String BOOLEAN_FALSE = "false";
```
HEdingfield, line 48: Nit: 3 of these are unused so delete, and the others should probably just be inlined.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -22,28 +22,100 @@
...
     this.config = config;
+    this.source = source;
+  }
+
+  // This method will extract candidate data from a CDF file for the contestID specified in
+  // our CvrSource.  It is currently un-used but will be handy for automating
+  // config file creation when we are ready to implement that.
+  Map<String, String> getCandidates() throws CvrParseException {
+    Map<String, String> candidates;
+    if (filePath.endsWith(".xml")) {
```
HEdingfield, line 66: You'll also need to modify `GuiConfigController.buttonCvrFilePathClicked()` so the file loader accepts "JSON and XML files".

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -100,9 +100,7 @@ private ContestConfig(RawContestConfig rawConfig, String sourceDirectory) {
   }
 
   static boolean isCdf(CvrSource source) {
-    return getProvider(source) == Provider.CDF
-        && source.getFilePath() != null
```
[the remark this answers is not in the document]
moldover: The filePath null check is redundant with passesBasicCvrSourceValidation and conflates validation logic with a function which tells you something about the source _type_.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -250,17 +248,19 @@ static boolean passesBasicCvrSourceValidation(CvrSource source) {
...
+      boolean providerRequiesContestId = provider == Provider.DOMINION ||
+          provider == Provider.HART ||
+          provider == Provider.CLEAR_BALLOT ||
+          provider == Provider.CDF;
+
+      if (isNullOrBlank(source.getContestId()) && providerRequiesContestId) {
         sourceValid = false;
         Logger.log(
             Level.SEVERE,
             String.format("contestId must be defined for CVR source with provider \"%s\"!",
```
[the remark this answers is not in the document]
moldover: It should always have been.  CDF reports (xml and json) support multiple contests (multiple entire elections in fact) and so it should have a contest id specified.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1028,19 +1028,8 @@ void setCandidateExclusionStatus(String candidateCode, boolean excluded) {
   // 1) if there are any CDF input sources extract candidates names from them
   // 2) build map of candidate ID to candidate name
   // 3) generate tie-break ordering if needed
-  private void processCandidateData() {
+  private void processCandidateData() throws CvrParseException {
     candidateCodeToNameMap = new HashMap<>();
-
-    for (RawContestConfig.CvrSource source : rawConfig.cvrFileSources) {
```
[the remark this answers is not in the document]
moldover: Oh yeah I missed the comment.  So this was doing the wrong thing.  It was extracting candidate data from the CDF, and then using that to do config validation checks.  In fact, it should be the other way around: use candidate data from the config to validate data in the CDF.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -164,4 +436,324 @@ void parseCvrFile(List<CastVoteRecord> castVoteRecords) throws UnrecognizedCandi
...
+    @JacksonXmlProperty()
+    @JacksonXmlElementWrapper(useWrapping = false)
+    Contest[] Contest;
+    @JacksonXmlProperty(isAttribute = true)
+    String ElectionScopeId;
+    @JacksonXmlProperty(isAttribute = true)
+    String Name;
+  }
+
+  static class CVR {
```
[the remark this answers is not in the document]
moldover: Yeah - this is one of the classes designed to work with Jackson XML parsing.  It needs to match the CDF naming convention exactly :/

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -258,7 +258,7 @@ private boolean setUpLogging(ContestConfig config) {
       try {
         if (ContestConfig.isCdf(source)) {
           Logger.log(Level.INFO, "Reading CDF cast vote record file: %s...", cvrPath);
-          new CommonDataFormatReader(cvrPath, config).parseCvrFile(castVoteRecords);
+          new CommonDataFormatReader(cvrPath, config, source).parseCvrFile(castVoteRecords);
```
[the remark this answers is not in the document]
moldover: +1

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
moldover: +1

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
moldover: Yeah I am kinda on the fence here.  My thinking here was to add some names to self-document that 'needs-adjudication' is actually a contest selection 'status' from the CDF spec, and not a magic in-line comment.  You really think it would be clearer to lose them?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -518,6 +517,12 @@ private void validateCvrFileSources() {
             isValid = false;
             Logger.log(Level.SEVERE, "CDF files must be tabulated individually.");
           }
+          if (!source.getFilePath().endsWith(".xml") && !source.getFilePath().endsWith(".json")) {
+            Logger.log(Level.SEVERE,
+                "Unexpected file extension: %s.  CDF source files must be .xml or .json",
+                source.getFilePath());
```
HEdingfield, line 523: Nit (to avoid long filepaths obscuring the important part of the message):
```
Logger.severe("CDF source files must be .json or .xml! Unexpected file extension for: %s", source.getFilePath());
```

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -203,23 +197,27 @@ CVRContest getCvrContest(CVR cvr, Contest contestToTabulate) {
...
+          break;
+        }
       }
     }
     // find CVRContest which matches the Contest we are tabulating
-    for (CVRContest cvrContest : currentSnapshot.CVRContest) {
-      if (cvrContest.ContestId.equals(contestToTabulate.ObjectId)) {
-        cvrContestToTabulate = cvrContest;
-        break;
+    if (currentSnapshot.CVRContest != null) {
```
HEdingfield, line 209: This still has issues with `currentSnapshot` potentially being `null` at this point. You can address it by replacing with:
```
if (currentSnapshot != null && currentSnapshot.CVRContest != null) {
```

src/main/java/network/brightspots/rcv/ContestConfig.java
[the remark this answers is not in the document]
HEdingfield: Agreed, this looks good to me now that we have the .xml / .json check in the validation function.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
HEdingfield: I see. I think, as a general rule, it's more readable to inline things that are used only once or maybe twice, but I can understand your justification here.

I do still think we should get rid of the three that aren't used at all though (`STATUS_YES`, `STATUS_UNKNOWN`, `BOOLEAN_FALSE`).

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -431,7 +431,7 @@ public void buttonCvrFilePathClicked() {
       case CDF -> {
         FileChooser fc = new FileChooser();
         fc.setInitialDirectory(new File(FileUtils.getUserDirectory()));
-        fc.getExtensionFilters().add(new ExtensionFilter("JSON files", "*.json"));
+        fc.getExtensionFilters().add(new ExtensionFilter("JSON and XML files", "*.json", "*.xml"));
```
HEdingfield, line 434: Oops, one more thing here... you need to un-disable the "Contest ID" field in the GUI when CDF is selected. To do this, look for this line:
```
case CLEAR_BALLOT, DOMINION, HART
```
And add `CDF` in there, and delete the clause above it that currently handles CDF separately.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -118,12 +192,203 @@
...
+      // Find the contest for tabulation
+      Contest contestToTabulate = null;
+      for (Contest contest : cvrReport.Election[0].Contest) {
+        if (contest.Name.equals(this.contestId)) {
+          contestToTabulate = contest;
+          break;
+        }
+      }
+      if (contestToTabulate == null) {
+        Logger.severe("Contest \"%s\" from config file not found!", this.contestId);
```
HEdingfield, line 241: `Logger.severe` (and likely the other new log functions) are not WAI. When I tested triggering this message, it said:

> "SEVERE: Contest "%s" from config file not found!"

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -518,6 +517,12 @@ private void validateCvrFileSources() {
...
+            isValid = false;
+          }
```
HEdingfield, line 525: Actually, could you please move this to `passesBasicCvrSourceValidation`? That way it triggers when attempting to add a CVR in the GUI, instead of just when attempting to tabulate.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
moldover: Are you getting these from some style checker or you just have the eagle eye?

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
moldover: Good catch.

.idea/misc.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="ExternalStorageConfigurationManager" enabled="true" />
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_14" default="true" project-jdk-name="14.0.1" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_X" default="true" project-jdk-name="14.0.1" project-jdk-type="JavaSDK">
```
tarheel, line 4: X?

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -22,28 +22,91 @@
...
+    try {
+      XmlMapper xmlMapper = new XmlMapper();
+      xmlMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+      FileInputStream inputStream = new FileInputStream(new File(filePath));
+      CastVoteRecordReport cvrReport = xmlMapper.readValue(inputStream, CastVoteRecordReport.class);
+      for (Contest contest : cvrReport.Election[0].Contest) {
+        if (!contest.ObjectId.equals(this.contestId)) {
+          continue;
+        }
+        for (ContestSelection contestSelection : contest.ContestSelection) {
```
tarheel, line 80: Say this 5 times fast.

Oh, Java.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -22,28 +22,91 @@
...
+              candidateName = candidate.Name;
+              // fallback to Code.Value
+              if (candidateName == null && candidate.Code != null && candidate.Code.Value != null) {
+                candidateName = candidate.Code.Value;
+              }
+              break;
+            }
+          }
+          if (candidateName == null) {
+            Logger.log(Level.WARNING, "No name found for CandidateId: %s", candidateId);
```
tarheel, line 95: is just a warning sufficient?

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -118,12 +189,205 @@
...
+          cvrContestToTabulate = cvrContest;
+          break;
+        }
+      }
+    }
+    return cvrContestToTabulate;
+  }
+
+  void parseXml(List<CastVoteRecord> castVoteRecords) {
+    try {
```
tarheel, line 218: for readability (so you don't have to indent this entire long method), I would move this `try` up a level in the call stack

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -118,12 +189,205 @@
...
+      // Find the contest for tabulation
+      Contest contestToTabulate = null;
+      for (Election election : cvrReport.Election) {
+        for (Contest contest : election.Contest) {
+          if (contest.Name.equals(this.contestId)) {
+            contestToTabulate = contest;
+            break;
+          }
+        }
+      }
```
tarheel, line 235: Just from reading the code, it's hard for me to figure out what the difference is between this block and the `getCvrContest` method. Could you clarify by adding more verbose comments and/or naming things more descriptively?

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -22,28 +22,91 @@
...
-  CommonDataFormatReader(String filePath, ContestConfig config) {
+  CommonDataFormatReader(String filePath, ContestConfig config, String contestId) {
     this.filePath = filePath;
     this.config = config;
+    this.contestId = contestId;
+  }
+
+  // This method will extract candidate data from a CDF file for the contestID specified in
+  // our CvrSource.
+  Map<String, String> getCandidates() throws CvrParseException {
```
tarheel, line 59: I'm confused. The only place you call this is in `parseJson`, so when would it ever call `getCandidatesXml`? And that method doesn't have any other callsites, so I don't understand how it gets used.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -118,12 +189,205 @@
...
+
+      // build a map of ContestSelections
+      HashMap<String, ContestSelection> contestSelectionById = new HashMap<>();
+      for (ContestSelection contestSelection : contestToTabulate.ContestSelection) {
+        if (contestSelection.CandidateIds.length > 1) {
+          Logger.warning(
+              "Multiple CandidateIds found in ContestSelection \"%s\". "
+                  + " Only the first one will be processed.",
+              contestSelection.ObjectId);
+        }
```
tarheel, line 258: I saw your note below about not supporting this yet. So should this be a fatal error instead of just a warning?

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -118,12 +189,205 @@
...
+          continue;
+        }
+        List<Pair<Integer, String>> rankings = new ArrayList<>();
+        // parse CVRContestSelections into rankings
+        // they will be null for an undervote
+        if (contest.CVRContestSelection != null) {
+          for (CVRContestSelection cvrContestSelection : contest.CVRContestSelection) {
+            if (cvrContestSelection.Status != null && cvrContestSelection.Status
+                .equals(STATUS_NEEDS_ADJUDICATION)) {
+              Logger.info("Contest Selection needs adjudication.  Skipping.");
```
tarheel, line 282: nit: double space

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -118,12 +189,205 @@
...
+            }
+            String contestSelectionId = cvrContestSelection.ContestSelectionId;
+            ContestSelection contestSelection = contestSelectionById
+                .get(contestSelectionId);
+            if (contestSelection == null) {
+              Logger.severe("ContestSelection \"%s\" from CVR not found!", contestSelectionId);
+              throw new CvrParseException();
+            }
+            String candidateId;
+            // check for declared write-in:
```
tarheel, line 293: You say "declared" here, but then you use the undeclared write-in label. Just a typo in your comment?

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -118,12 +189,205 @@
...
+                .equals(BOOLEAN_TRUE)) {
+              candidateId = this.config.getUndeclaredWriteInLabel();
+            } else {
+              Candidate candidate = candidateById
+                  .get(contestSelection.CandidateIds[0]);
+              if (candidate == null) {
+                Logger.severe("CandidateId \"%s\" from ContestSelectionId \"%s\" not found!",
+                    contestSelection.CandidateIds[0], contestSelection.ObjectId);
+                throw new CvrParseException();
+              }
```
tarheel, line 304: use `unrecognizedCandidateCounts` instead of throwing immediately?

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -118,12 +189,205 @@
...
+              if (candidateId.equals(config.getOvervoteLabel())) {
+                candidateId = Tabulator.EXPLICIT_OVERVOTE_LABEL;
+              }
+              if (!config.getCandidateCodeList().contains(candidateId)) {
+                Logger
+                    .severe(
+                        "Contest Selection CandidateId: \"%s\" from CVR is not in the config file!",
+                        candidateId);
+                throw new CvrParseException();
+              }
```
tarheel, line 315: Or I guess this is where we would use `unrecognizedCandidateCounts`. So the previous check would catch an internal inconsistency within the CDF file, whereas this check would catch an inconsistency between the CDF and the config?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -209,6 +206,14 @@ static boolean passesBasicCvrSourceValidation(CvrSource source) {
           Logger.log(Level.SEVERE, "overvoteDelimiter is invalid.");
         }
       } else {
+        if (provider == Provider.CDF) {
+          if (!source.getFilePath().endsWith(".xml") && !source.getFilePath().endsWith(".json")) {
+            Logger.log(Level.SEVERE,
+                "Unexpected file extension: %s.  CDF source files must be .xml or .json",
```
tarheel, line 212: double space

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -209,6 +206,14 @@ static boolean passesBasicCvrSourceValidation(CvrSource source) {
           Logger.log(Level.SEVERE, "overvoteDelimiter is invalid.");
         }
       } else {
+        if (provider == Provider.CDF) {
+          if (!source.getFilePath().endsWith(".xml") && !source.getFilePath().endsWith(".json")) {
```
tarheel, line 210: use `toLowerCase` for these checks to be safe

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -518,6 +525,12 @@ private void validateCvrFileSources() {
             isValid = false;
             Logger.log(Level.SEVERE, "CDF files must be tabulated individually.");
           }
+          if (!source.getFilePath().endsWith(".xml") && !source.getFilePath().endsWith(".json")) {
+            Logger
+                .severe("CDF source files must be .json or .xml! Unexpected file extension for: %s",
+                    source.getFilePath());
+            isValid = false;
+          }
```
tarheel, line 533: Why are we checking this again here? Isn't it redundant with lines 209-216?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -250,17 +255,19 @@ static boolean passesBasicCvrSourceValidation(CvrSource source) {
         }
       }
 
-      if (isNullOrBlank(source.getContestId()) &&
-          (provider == Provider.DOMINION || provider == Provider.HART
-              || provider == Provider.CLEAR_BALLOT)) {
+      boolean providerRequiresContestId = provider == Provider.DOMINION ||
+          provider == Provider.HART ||
+          provider == Provider.CLEAR_BALLOT ||
+          provider == Provider.CDF;
```
tarheel, line 261: Apologies if I missed the earlier discussion on this point: why are we adding this requirement? Is it necessary for both JSON and XML? Why didn't we need to have it before?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1025,22 +1038,11 @@ void setCandidateExclusionStatus(String candidateCode, boolean excluded) {
   }
 
   // perform pre-processing on candidates:
-  // 1) if there are any CDF input sources extract candidates names from them
-  // 2) build map of candidate ID to candidate name
-  // 3) generate tie-break ordering if needed
+  // 1) build map of candidate ID to candidate name
+  // 2) generate tie-break ordering if needed
+  // 3) add uwi candidate if needed
```
tarheel, line 1043: Does this mean now we're requiring CDF configs to include candidates? Why did you have to make this change?

tarheel (commented):
Apologies if I'm repeating questions that @HEdingfield already asked. I haven't dug through all of the existing discussion in this PR.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
HEdingfield: This one was from `Analyze => Inspect Code...` but the rest were all eagle eyes, baybee!

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: Yeah, I was secretly trying to get him to fix this by suggesting this haha:
```
Logger.severe("CDF source files must be .json or .xml! Unexpected file extension for: %s", source.getFilePath());
```

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -518,9 +526,9 @@ private void validateCvrFileSources() {
             Logger.log(Level.SEVERE, "CDF files must be tabulated individually.");
           }
           if (!source.getFilePath().endsWith(".xml") && !source.getFilePath().endsWith(".json")) {
-            Logger.log(Level.SEVERE,
-                "Unexpected file extension: %s.  CDF source files must be .xml or .json",
-                source.getFilePath());
+            Logger
```
HEdingfield, line 529: You can delete this now (since you moved it to the "basic" check, though I recommend using the message you changed it to here for that one.

src/main/java/network/brightspots/rcv/ContestConfig.java
moldover: @tarheel  Right - the main reason is, our tabulation architecture relies on the config file defining valid candidates for a contest to be tabulated.  This change treats all data sources "equally" in that respect.  It generalizes validation code across source types, and generalizes the feature set, which enables usage of candidate codes and excluding candidates from tabulation of CDF sources.

To be sure, simply using the embedded CDF candidate data is easier in that you don't have to bother putting it in the config (which I have spent a lot of time doing and not done yet) but I think -- and would like to get your feedback -- the better solution is to eventually provide a feature to extract the candidate data from the CDF into a config.

tarheel: Yeah, that all sounds fine to me. I agree that the ideal end state would be to require the data in the config, but make it easy to copy it over from the source.

.idea/misc.xml
moldover: For some weird reason intelliJ does this sometimes.  I'll fix it.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
moldover: Technically you can have Candidates without Name, so it's still possible to tabulate.  However, this entire code block (and the getCandidatesJson one) are going away in the next PR, so this can simply be ignored for now.  If you want a preview, check out the branch json_cdf which builds on this one.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
moldover: I think this is the first time you've asked me to add comments.  Very exciting :)

src/main/java/network/brightspots/rcv/ContestConfig.java
moldover: Yeah it does seem redundant.  @HEdingfield is that right?

moldover: Yeah, Hylton mentions this below.

src/main/java/network/brightspots/rcv/ContestConfig.java
moldover: It should have been there all along.  CDF supports multiple contests, therefore we need to specify which contest to tabulate.  Our CDF json examples created by George only contained one contest each, so it may have seemed appropriate at the time to just assume that the one contest is 'the one to tabulate' but that is not the case.

HEdingfield: Please update the `config_file_documentation.txt` section on `contestId` to reflect this. It currently only mentions Hart and Dominion. It'd also be great if you could make a quick pass on that file with the new validation stuff you're adding here to make sure there aren't any other notes that should be added.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -118,15 +140,209 @@
...
+            candidateId = candidate.Name;
+            if (candidateId.equals(config.getOvervoteLabel())) {
+              candidateId = Tabulator.EXPLICIT_OVERVOTE_LABEL;
+            }
+            if (!config.getCandidateCodeList().contains(candidateId)) {
+              Logger
+                  .severe(
+                      "Contest Selection CandidateId: \"%s\" from CVR is not in the config file!",
+                      candidateId);
+              throw new CvrParseException();
```
tarheel, line 268: I'm still wondering whether we should be using unrecognizedCandidateCounts here instead of throwing exceptions.

tarheel (approved):
Just one more question, but go ahead and merge when you're ready.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
moldover: @tarheel I agree, in fact I already made that change in both xml and json parsers in the other open PR #506 which builds on this code.

Issue #505, opened 2020-09-13
CDF JSON reader / writer and tests are broken 


moldover:
Our current implementation of CDF json reading (CommonDataFormatReader.java) and writing (ResultsWriter.java) are broken.  
 - They do not create top-level Candidate objects.  
 - They do not link Contest Selection objects to Candidate objects.  
 - They do not filter by contestID 
 - Regression test assets mistakenly reference CDF internal ids (contest Id, and candidate selection ids).

See https://github.com/BrightSpots/rcv/blob/master/reference/NIST-CommonDataFormat.png
https://github.com/BrightSpots/rcv/blob/master/reference and #504 for additional details on the correct format and implementation.

Issue #251, opened 2019-04-06
Refactor CommonDataFormatReader and StreamingCVRReader into common base class


moldover:
And come up with some clever names for them...

PR #506, opened 2020-09-13
Cdf Json Fixes

moldover:
This PR addresses #505 by fixing our Cdf Json reader and writer implementation.  This is essentially the analogous work to #504 but in Json.  It also updates all existing regression test assets to reflect the fixes.

Might be best to start by looking at how one of the regression test input cdf cvr files has changed.  That will demonstrate the linkage between Candidate objects and ContestSelection objects.

There are a couple small things remaining to do, but I wanted to get this PR going as the core tabulation code, and test assets are ready to go.


src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -336,51 +261,183 @@ void parseCvrFile(List<CastVoteRecord> castVoteRecords)
     }
   }
 
-  void parseJson(List<CastVoteRecord> castVoteRecords) throws UnrecognizedCandidatesException {
-    // cvrIndex and fileName are used to generate IDs for cvrs
+  HashMap getCvrContestJson(HashMap cvr, String contestIdToTabulate) {
```
tarheel, line 264: can this be private?

tarheel (commented):
Could you explain a bit what the changes are to CommonDataFormatReader? I'm having trouble looking at this PR and figuring out how much is just moving code around and how much is substantive changes to the logic.

Also, would it be possible to make it an abstract class and then create a separate concrete child class for each format? I'm finding it hard to understand which logic is shared and which is format-specific. You're already sort of doing this just with the way you're naming the internal methods, so maybe this won't be a difficult refactor.

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -136,18 +137,27 @@ private static String generateCvrSnapshotId(String cvrId, Integer round) {
         : String.format("ballot-%s", cvrId);
   }
 
-  private static String getCdfIdForCandidateCode(String code) {
-    String id = candidateCodeToCdfId.get(code);
+  // generates an internal ContestSelectionId based on a candidate code
+  private static String getContestSelectionIdForCandidateCode(String code) {
```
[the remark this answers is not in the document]
moldover: Yeah sorry about that @tarheel!  I did not give you much explanation.

Among the important things:
1) All static election objects (Candidates, Contests, ContestIds) are parsed at beginning of cvr processing.  This simplifies and clarifies the parsing somewhat.
2) **CVRContestSelections now contain links to Candidate objects.**  During cvr processing these links are used to look up the linked Candidate.  From there, Candidate Name field is used to assign the rankings and create the CastVoteRecord for tabulation.  Previously, Candidate names were embedded directly in the CVRContestSelection objects using a "Code" object with a 'vendor-label' field on it.  This is non-standard, and while it's conceivable a vendor might do this (Code objects allow flexibility) it's not the primary way CDF encodes information about a Candidate.  The primary way is by linking to a Candidate object.
3) **Candidate names are stored in config** - not pulled from CDF file at tabulation time.  This normalizes candidate logic across formats.
4) **Configs do not use CDF internal ids to as candidate codes.**  It's important to understand that CDF internal object links (e.g. CandidateIds contained in a CVRContestSelection in CDF should be used only _within_ the context of the CDF.  They are generated by the cvr report generating program, and should not appear in our config files.
5) **Filter contest tabulation by contest ID as** specified in the config file

So to your second question: yes it would be possible for these methods to share an interface.  I think the real move is for all reader objects to share an interface.  I don't think what you propose would be difficult, nevertheless it's outside the scope of this PR.

src/test/resources/network/brightspots/rcv/test_data/test_set_1_exhaust_at_overvote/test_set_1_exhaust_at_overvote_config.json
```diff
@@ -1,7 +1,7 @@
 {
   "tabulatorVersion": "TEST",
   "outputSettings": {
-    "contestName": "Test Set 1",
+    "contestName": "test set 1",
```
moldover, line 4: @tarheel this is still just for output.

src/test/resources/network/brightspots/rcv/test_data/test_set_1_exhaust_at_overvote/test_set_1_exhaust_at_overvote_config.json
```diff
@@ -15,23 +15,23 @@
     "idColumnIndex" : "",
     "precinctColumnIndex" : "",
     "provider" : "CDF",
-    "contestId": "contest-001"
+    "contestId": "test set 1"
```
moldover, line 18: Tabulator will filter Contest to tabulate based on this.

src/test/resources/network/brightspots/rcv/test_data/test_set_1_exhaust_at_overvote/test_set_1_exhaust_at_overvote_config.json
```diff
@@ -15,23 +15,23 @@
...
   } ],
   "candidates" : [ {
-    "name" : "A",
-    "code" : "cs-a",
+    "name" : "Candidate A Name",
```
moldover, line 21: Here I removed the internal cdf CandidateId usage, and changed the name to something that is more clearly a Name.

src/test/resources/network/brightspots/rcv/test_data/test_set_1_exhaust_at_overvote/test_set_1_exhaust_at_overvote_cvr.json
```diff
@@ -799,45 +799,44 @@
   "Election" : [ {
     "@id" : "election-001",
     "@type" : "CVR.Election",
+    "Candidate" : [
```
moldover, line 802: Here you see a very simple example of top-level "static" Candidate objects.  One could attach more Candidate info here e.g. a PartyId.

src/test/resources/network/brightspots/rcv/test_data/test_set_1_exhaust_at_overvote/test_set_1_exhaust_at_overvote_cvr.json
```diff
@@ -799,45 +799,44 @@
...
         "@id" : "cs-a",
-        "@type" : "CVR.ContestSelection",
-        "Code" : [ {
-          "@type" : "CVR.Code",
-          "OtherType" : "vendor-label",
-          "Type" : "other",
-          "Value" : "A"
-        } ]
+        "@type" : "CandidateSelection",
+        "CandidateIds" : [ "candidate-a" ]
```
moldover, line 827: @tarheel Here we replace a 'Code' object with a link to a Candidate object.  Note that confusingly it is actually an array, and is correspondingly named CandidateIds.  This is to support "party ticket" options, and is a rare thing, but we all have to pay the price.  It's important to not the _multiplicity_ of links in the CDF UML spec.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -336,51 +261,183 @@ void parseCvrFile(List<CastVoteRecord> castVoteRecords)
...
+          if (candidateIds == null || candidateIds.size() == 0) {
+            Logger.severe("CandidateSelection \"%s\" has no CandidateIds!", contestSelectionId);
+            throw new CvrParseException();
+          }
+          if (candidateIds.size() > 1) {
+            Logger.warning("CandidateSelection \"%s\" has multiple CandidateIds.  "
+                + "Only the first one will be processed.", contestSelectionId);
+          }
+          String candidateObjectId = (String) candidateIds.get(0);
+          HashMap Candidate = (HashMap) candidates.get(candidateObjectId);
```
tarheel, line 380: why is it Candidate instead of candidate here?

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -336,51 +261,183 @@ void parseCvrFile(List<CastVoteRecord> castVoteRecords)
...
+          candidateId = (String) Candidate.get("Name");
+          if (candidateId.equals(config.getOvervoteLabel())) {
+            candidateId = Tabulator.EXPLICIT_OVERVOTE_LABEL;
+          }
+          if (!this.config.getCandidateCodeList().contains(candidateId)) {
+            Logger.severe("Unrecognized candidate found in CVR: %s", candidateId);
+            unrecognizedCandidateCounts.merge(candidateId, 1, Integer::sum);
+          }
```
tarheel, line 388: If you set candidateId to EXPLICIT_OVERVOTE_LABEL, isn't the check in line 385 guaranteed to fail? I think maybe 385 should be `else if` instead of `if`.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -336,51 +261,183 @@ void parseCvrFile(List<CastVoteRecord> castVoteRecords)
...
+          Integer rank = (Integer) (cvrContestSelection.get("Rank"));
+          rankings.add(new Pair<>(rank, candidateId));
+        } else {
+          // extract all the SelectionPositions (ranks) which this selection has been assigned
+          ArrayList selectionPositions = (ArrayList) cvrContestSelection.get("SelectionPosition");
+          for (Object selectionPositionObject : selectionPositions) {
+            HashMap selectionPosition = (HashMap) selectionPositionObject;
+            // WriteIn can be linked at the selection position level
+            if (selectionPosition.containsKey("CVRWriteIn")) {
+              candidateId = this.config.getUndeclaredWriteInLabel();
```
tarheel, line 404: what if `getUndeclaredWriteInLabel()` returns null?

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -136,18 +137,27 @@ private static String generateCvrSnapshotId(String cvrId, Integer round) {
         : String.format("ballot-%s", cvrId);
   }
 
-  private static String getCdfIdForCandidateCode(String code) {
-    String id = candidateCodeToCdfId.get(code);
+  // generates an internal ContestSelectionId based on a candidate code
+  private static String getCvrContestSelectionIdForCandidateCode(String code) {
```
tarheel, line 141: I thought I commented on this before, but I don't see my comment. This and the next method (and the maps they use) should be labeled so that it's clear they're CDF-specific.

src/main/java/network/brightspots/rcv/ResultsWriter.java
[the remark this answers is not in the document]
tarheel: Could you create a follow-up issue (not for this deadline) to refactor this class?

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -336,8 +336,11 @@ private boolean setUpLogging(ContestConfig config) {
         encounteredSourceProblem = true;
       } catch (CvrParseException e) {
         encounteredSourceProblem = true;
+      } catch (Exception exception) {
+        Logger.severe("Unexpected error parsing source file %s", cvrPath);
```
HEdingfield, line 340: If you're going to catch a general exception like this (which is typically not recommended), I'd at least make sure you include the details in the error message, e.g.:

```
} catch (Exception e) {
  Logger.severe("Unexpected error parsing source file: %s\n%s", cvrPath, e.toString());
```

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -336,51 +261,183 @@ void parseCvrFile(List<CastVoteRecord> castVoteRecords)
...
+    HashMap currentSnapshot = null;
+    for (Object snapshotObject : cvrSnapshots) {
+      HashMap snapshot = (HashMap) snapshotObject;
+      if (snapshot.get("@id").equals(currentSnapshotId)) {
+        currentSnapshot = snapshot;
+        break;
+      }
+    }
+    // find the cvr contest in this snapshot
+    ArrayList cvrContests = (ArrayList) currentSnapshot.get("CVRContest");
```
HEdingfield, line 277: Account for potential NPE with `currentSnapshot.get()`.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -336,51 +261,183 @@ void parseCvrFile(List<CastVoteRecord> castVoteRecords)
...
+    return cvrContestToTabulate;
+  }
+
+  // parse cdf json CastVoteRecordReport into CastVoteRecords and append them to input list
+  void parseJson(List<CastVoteRecord> castVoteRecords)
+      throws CvrParseException, UnrecognizedCandidatesException {
+
+    // static election data
+    HashMap<Object, Object> candidates = new HashMap<>();
+    HashMap<Object, Object> gpuUnits = new HashMap<>();
```
HEdingfield, line 296: This doesn't seem like it's actually used for anything.

src/main/java/network/brightspots/rcv/ResultsWriter.java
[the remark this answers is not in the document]
moldover: Sure, we already have such a ticket: https://github.com/BrightSpots/rcv/issues/251

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
moldover: oh you're right - I just missed that.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
moldover: Good catch!

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
moldover: @tarheel what do you think we should do?  Require that CDF tabulations have an UWI label?

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
moldover: True - these are for precincts - I just added the parsing code.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
tarheel: I would just throw an error here if the UWI label is null.

tarheel (approved):

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -336,51 +274,214 @@ void parseCvrFile(List<CastVoteRecord> castVoteRecords)
...
+            HashMap selectionPosition = (HashMap) selectionPositionObject;
+            // WriteIn can be linked at the selection position level
+            if (selectionPosition.containsKey("CVRWriteIn")) {
+              candidateId = this.config.getUndeclaredWriteInLabel();
+              if (candidateId == null || candidateId.isEmpty()) {
+                Logger.severe(
+                    "Undeclared write-in candidate found while parsing CVR, however undeclared "
+                        + "write-in label has not been defined in config.");
+                throw new CvrParseException();
+              }
```
tarheel, line 433: @moldover I added an inline comment here but GitHub's crappy UI seems to have discarded it...? Anyway, the comment was that you added the same six lines in four different places, so please factor it out into a helper. Also, you can use `isNullOrBlank`.

Issue #501, opened 2020-08-17
Redesign `nonIntegerWinningThreshold` and `hareQuota` in GUI and add new validation rule


HEdingfield:
* Per mock-up, these settings will only be enabled for multi-winner, and the two check boxes will be replaced by a 3-way radio button.
* Non-Integer Winning Threshold = false and Hare Quota = false:
  * Compute using most common threshold formula
  = ((Votes/(Seats+1))+1, disregarding fractions
* Non-Integer Winning Threshold = true and Hare Quota = false:
  * Compute using HB Quota
  \> (Votes / (Seats +1))
* Non-Integer Winning Threshold = false and Hare Quota = true:
  * Compute using Hare Quota
  = (Votes / Seats)
* Non-Integer Winning Threshold = true and Hare Quota = true:
  * **This will become an invalid combination that needs to be rejected during validation.**

Relevant discussion:

@chughes297:
> We made this a radio button because you can technically run an election using both Hare and HB Quotas but that doesn't exist in law anywhere and it's (as far as I know) never been used in elections because Hare is outmoded as a threshold. So we wanted to foreclose the ability to combine those as thresholds. happy to discuss further, though. 

@HEdingfield:
> Louis / Jon, do you agree that we should get rid of the possibility of both Hare Quota and Non-Integer Winning Threshold both being set to true? If so, do we add validation to prevent this, potentially invalidating past configs, and also change the docs in the .txt? Or do we leave it as a possibility via manually-edited config files and not via the GUI?

@tarheel:
> Fine with me. Yeah, let's just add validation for it.

@HEdingfield:
> Our documentation (see below) doesn't seem to 100% match the formulas listed in the radio button options here.The formulas differ subtly, and it also seems that there should be a fourth option (where both values are true), right? If so, what's the wording you want? And how do you want to correct the wording for the other ones?
> 
>  "nonIntegerWinningThreshold" optional
the vote threshold used to determine winners can be a non-integer
if true, threshold = V/(S+1) + 10^-d
if false, threshold = floor(V/(S+1)) + 1
where V = total number of votes (in the current round for single-seat or in the first round for multi-seat); S = numberOfWinners; and d = decimalPlacesForVoteArithmetic
(note that S+1 in the formulas above becomes just S if hareQuota is set to true)
value: true | false
if not supplied: false 
>
>  "hareQuota" optional
[...]

PR #512, opened 2020-09-26
Redesigns handling of `nonIntegerWinningThreshold` and `hareQuota` in the GUI

HEdingfield:
* Replaces `checkBoxNonIntegerWinningThreshold` and `checkBoxHareQuota` with a radio button array in the FXML file (addressing #501)
* Implements radio button array in GUI to determine `nonIntegerWinningThreshold` and `hareQuota` values, and adds new validation rules for those settings.
* GUI now disables `numberOfWinners` field and sets it to 1 only when `winnerElectionMode` is "Single-winner majority determines winner".
* GUI now disables `numberOfWinners` field and sets it to 0 when `winnerElectionMode` is "Bottoms-up using percentage threshold".
* Fixes bugs in validation error messages when `numberOfWinners` is 0
* Suppresses `rawtypes` warnings in `GuiConfigController`.

### DIFF
--- a/src/main/java/network/brightspots/rcv/ContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/ContestConfig.java
@@ -61,7 +61,6 @@ class ContestConfig {
   static final int SUGGESTED_CVR_FIRST_VOTE_ROW = 2;
   static final int SUGGESTED_CVR_ID_COLUMN = 1;
   static final int SUGGESTED_CVR_PRECINCT_COLUMN = 2;
-  static final int SUGGESTED_NUMBER_OF_WINNERS = 1;
   static final int SUGGESTED_DECIMAL_PLACES_FOR_VOTE_ARITHMETIC = 4;
   static final int SUGGESTED_MAX_SKIPPED_RANKS_ALLOWED = 1;
   static final String SUGGESTED_OVERVOTE_LABEL = "overvote";
@@ -722,21 +721,13 @@ class ContestConfig {
             Logger.log(Level.SEVERE, "batchElimination can't be true in a multi-seat contest!");
           }
         } else { // numberOfWinners == 1
-          if (isMultiSeatSequentialWinnerTakesAllEnabled() ||
-              isMultiSeatBottomsUpUntilNWinnersEnabled() ||
-              isMultiSeatAllowOnlyOneWinnerPerRoundEnabled() ||
-              isMultiSeatAllowMultipleWinnersPerRoundEnabled()) {
+          if (!isSingleWinnerEnabled()) {
             isValid = false;
             Logger.log(
                 Level.SEVERE,
                 "winnerElectionMode can't be \"%s\" in a single-seat contest!",
                 winnerMode.toString()
             );
-          }
-
-          if (isHareQuotaEnabled()) {
-            isValid = false;
-            Logger.log(Level.SEVERE, "hareQuota can only be true in a multi-seat contest!");
           }
         }
       } else { // numberOfWinners == 0
@@ -760,6 +751,27 @@ class ContestConfig {
       isValid = false;
       Logger.log(Level.SEVERE, "batchElimination can't be true when winnerElectionMode is \"%s\"!",
           winnerMode.toString());
+    }
+
+    if (!isMultiSeatAllowOnlyOneWinnerPerRoundEnabled()
+        && !isMultiSeatAllowMultipleWinnersPerRoundEnabled()) {
+      if (isNonIntegerWinningThresholdEnabled()) {
+        isValid = false;
+        Logger.log(Level.SEVERE,
+            "nonIntegerWinningThreshold can't be true when winnerElectionMode is \"%s\"!",
+            winnerMode.toString());
+      }
+      if (isHareQuotaEnabled()) {
+        isValid = false;
+        Logger.log(Level.SEVERE, "hareQuota can't be true when winnerElectionMode is \"%s\"!",
+            winnerMode.toString());
+      }
+    }
+
+    if (isNonIntegerWinningThresholdEnabled() && isHareQuotaEnabled()) {
+      isValid = false;
+      Logger.log(Level.SEVERE,
+          "nonIntegerWinningThreshold and hareQuota can't both be true at the same time!");
     }
 
     if (!isNullOrBlank(getOvervoteLabel())
@@ -825,6 +837,10 @@ class ContestConfig {
   WinnerElectionMode getWinnerElectionMode() {
     WinnerElectionMode mode = WinnerElectionMode.getByLabel(rawConfig.rules.winnerElectionMode);
     return mode == null ? WinnerElectionMode.MODE_UNKNOWN : mode;
+  }
+
+  boolean isSingleWinnerEnabled() {
+    return getWinnerElectionMode() == WinnerElectionMode.STANDARD;
   }
 
   boolean isMultiSeatAllowOnlyOneWinnerPerRoundEnabled() {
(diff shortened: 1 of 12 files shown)


src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -756,6 +753,27 @@ private void validateRules() {
           winnerMode.toString());
     }
 
+    if (!isMultiSeatAllowOnlyOneWinnerPerRoundEnabled()
+        && !isMultiSeatAllowMultipleWinnersPerRoundEnabled()) {
```
moldover, line 757: This condition is pretty confusing - can you add a comment on what we're checking for in this block?

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -76,7 +76,7 @@
 import network.brightspots.rcv.Tabulator.TieBreakMode;
 import network.brightspots.rcv.Tabulator.WinnerElectionMode;
 
-@SuppressWarnings("WeakerAccess")
+@SuppressWarnings({"WeakerAccess", "rawtypes"})
```
moldover, line 79: This is it?  Nice find!

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -722,30 +721,28 @@ private void validateRules() {
...
+        } else {
+          if (getMultiSeatBottomsUpPercentageThreshold() == null) {
+            isValid = false;
+            Logger.log(Level.SEVERE,
+                "If winnerElectionMode is \"%s\", multiSeatBottomsUpPercentageThreshold must be specified!",
+                winnerMode.toString());
+          }
         }
       }
     }
```
moldover, line 748: This block here:  if (Utils.isInt(getNumberOfWinnersRaw())) { ... }
Is it doing anything useful?  Should there be an else?

moldover (commented):
LGTM.  Will let @tarheel have a pass.

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: If the user provides a non-integer value for this, it will display the message:

> numberOfWinners must be an integer from 0 to 2147483647!

Per a check that's done above. This needs to be here so the remaining validation around that field doesn't occur (and thus throw `NumberFormatException`s) when that's the case.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -821,6 +840,10 @@ WinnerElectionMode getWinnerElectionMode() {
     return mode == null ? WinnerElectionMode.MODE_UNKNOWN : mode;
   }
 
+  boolean isSingleWinnerEnabled() {
+    return getWinnerElectionMode() == WinnerElectionMode.STANDARD;
```
tarheel, line 844: Can we rename `STANDARD` to `STANDARD_SINGLE_WINNER` or something? Just calling it "standard" is too ambiguous.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -722,30 +721,28 @@ private void validateRules() {
...
-        if (!isMultiSeatBottomsUpWithThresholdEnabled()
-            || getMultiSeatBottomsUpPercentageThreshold() == null) {
+        if (!isMultiSeatBottomsUpWithThresholdEnabled()) {
           isValid = false;
           Logger.log(Level.SEVERE,
               "If numberOfWinners is zero, winnerElectionMode must be \"%s\" and multiSeatBottomsUpPercentageThreshold must be specified!",
-              winnerMode.toString());
+              WinnerElectionMode.MULTI_SEAT_BOTTOMS_UP_USING_PERCENTAGE_THRESHOLD);
+        } else {
+          if (getMultiSeatBottomsUpPercentageThreshold() == null) {
```
tarheel, line 740: combine these two lines

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -756,6 +753,28 @@ private void validateRules() {
           winnerMode.toString());
     }
 
+    // nonIntegerWinningThreshold and hareQuota are only allowed for multi-seat elections
+    if (!isMultiSeatAllowOnlyOneWinnerPerRoundEnabled()
+        && !isMultiSeatAllowMultipleWinnersPerRoundEnabled()) {
```
tarheel, line 758: perhaps it would be useful to have a helper like `isStandardMultiSeatEnabled()` that matches these two modes?

src/main/java/network/brightspots/rcv/GuiConfigController.java
tarheel: it's a Christmas miracle!

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -1063,33 +1065,32 @@ public LocalDate fromString(String string) {
...
-          checkBoxHareQuota.setDisable(false);
           textFieldNumberOfWinners.setDisable(false);
         }
         case MULTI_SEAT_BOTTOMS_UP_USING_PERCENTAGE_THRESHOLD -> {
           textFieldMaxRankingsAllowed.setDisable(false);
           textFieldMinimumVoteThreshold.setDisable(false);
           choiceTiebreakMode.setDisable(false);
-          checkBoxHareQuota.setDisable(false);
-          textFieldNumberOfWinners.setDisable(false);
+          textFieldNumberOfWinners.setText("0");
```
tarheel, line 1093: can we hide it entirely in this case?

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
```diff
@@ -451,27 +451,27 @@
...
+                      </fx:define>
+                      <Label text="Threshold Calculation Method *"/>
+                      <RadioButton alignment="TOP_LEFT" mnemonicParsing="false"
+                        toggleGroup="$thresholdCalculationMethodToggleGroup"
+                        fx:id="radioThresholdMostCommon"
+                        text="Compute using most common threshold formula&#xD;  = ((Votes / (Seats + 1)) + 1, disregarding fractions"/>
+                      <RadioButton alignment="TOP_LEFT" mnemonicParsing="false"
+                        toggleGroup="$thresholdCalculationMethodToggleGroup"
+                        fx:id="radioThresholdHbQuota"
+                        text="Compute using HB Quota&#xD;  > (Votes / (Seats +1))"/>
```
tarheel, line 466: space between + and 1?

tarheel (approved):
just some minor things inline

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: If we end up doing the same check in the future, I'd be on board with this, but will leave as-is for now.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: The answer is yes, but it wouldn't move any of the options around, it'd just appear as a blank space there, which is awkward. I also think we've had this conversation a number of times and I've been pretty firmly in the camp of keeping the user as aware of what ends up in the config file as possible. Happy to reassess this in the future for the entire GUI if you'd like to file an issue for that. Chris would probably be excited about this idea.

PR #514, opened 2020-09-27
Adds more hints and makes a fix to config_file_documentation.txt

HEdingfield:


src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
```diff
@@ -1,13 +1,57 @@
...
+
   * Multi-pass IRV: Elects multiple winners. Eliminate candidates one-by-one or using batch elimination until only two candidates remain. Candidate with the most votes at the end wins. Run a new set of rounds with any winning candidates ignored.
 
+Maximum Number of Candidates That Can Be Ranked: How many rankings each voter has in this contest.
+
+Minimum Vote Threshold: The number of first-choice votes a candidate must receive in order to remain in the race. Most jurisdictions do not set a minimum vote threshold.
+
+Use Batch Elimination: Batch elimination, or simultaneous elimination of all candidates for whom it is mathematically impossible to be elected, eliminates all candidates who cannot receive enough votes to surpass the candidate with the next highest number of votes. Example: in a six candidate contest with 200 votes, Candidate A has 80 votes, Candidate B has 70, and the other four combined have 50. Because those four candidates can never combine their votes to surpass Candidate B, they can be batch eliminated.
```
tarheel, line 21: This should note that it only applies to single-winner contests.

src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
```diff
@@ -1,13 +1,57 @@
...
+
+Continue until Two Candidates Remain: Single-winner ranked-choice voting elections can stop as soon as a candidate receives a majority of votes, even though 3 or more candidates may still be in the race. Selecting this option will run the round-by-round count until only two candidates remain, regardless of when a candidate wins a majority of votes.
+
+Tiebreak mode (required): Tiebreak procedures are set in law, either in the ranked-choice voting law used in your jurisdiction or in the elections code more generally. Select the option from this list that complies with law and procedure in your jurisdiction.
```
tarheel, line 25: Mode

src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
```diff
@@ -1,13 +1,57 @@
...
+
+  * Stop counting and ask: Pause count when a tie is reached. User is prompted to select any tied candidate to eliminate.
+
+  * Previous round counts (then random): The tied candidate with the least votes in the previous round loses the tie. If there is a tie in the previous round, the tie is broken randomly. Requires a random seed.
+
+  * Previous round counts (then stop counting and ask): The tied candidate with the least votes in the previous round loses the tie. If there is a tie in the previous round, user is prompted to select any tied candidate to eliminate.
+
+  * Use candidate order in the config file: Use the order of candidates in the config file to determine ties. Candidates lower in the list are eliminated in case of a tie.
+
+  * Generate permutation: Generate a randomly ordered list of candidates in the contest. Candidates lower in the permutation are eliminated in case of a tie. Requires a random seed.
```
tarheel, line 37: @chughes297 we need to rephrase these to reflect the fact that the tiebreak mode is also used to break ties when multiple candidates win at the same time with the same vote total. (That's even more of an edge case than a loser tiebreak, of course, so I'd still recommend putting most of the emphasis on the losing case.)

src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
```diff
@@ -1,13 +1,57 @@
...
+
+Random seed (required if Tiebreak Mode is "Random", "Previous round counts (then random)", or "Generate permutation"): Enter a positive or negative integer to generate random orders.
+
+Number of Winners: The number of seats to be filled in the contest.
+
+Percentage Threshold: The share of votes a candidate must have in order to win. Candidates falling below this threshold are eliminated one-by-one beginning with the candidate with the fewest votes.
```
tarheel, line 43: Note that this only relevant for `MULTI_SEAT_BOTTOMS_UP_USING_PERCENTAGE_THRESHOLD` (whatever we're calling that now).

src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
```diff
@@ -1,13 +1,57 @@
...
+
+Threshold Calculation Method: The threshold of election is the number of votes a candidate must receive in order to win election. There are three primary ways to calculate the threshold of election in multi-winner RCV contests. This will be set in law (either by statute or regulation) in your jurisdiction.
```
tarheel, line 45: And this is specifically *not* applicable when the mode is `MULTI_SEAT_BOTTOMS_UP_USING_PERCENTAGE_THRESHOLD`.

src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
chughes297: Added brief qualifiers or re-wrote sentences to capture this! Thanks for catching. 

Also thought it may help to add a couple sentences to the intro to the tiebreak mode hint, would add these two sentences to the end?

Note that ties in ranked-choice voting contests can occur when eliminating candidates or when electing candidates. Multi-winner contests can have ties between candidates who have both crossed the threshold of election; in that case ties are broken to determine whose surplus vote value transfers first. 

  * Random: Randomly select a tied candidate to eliminate or, in multi-winner contests only, elect. Requires a random seed.

  * Stop counting and ask: Pause count when a tie is reached. User is prompted to select any tied candidate to eliminate or, in multi-winner contests only, elect.

  * Previous round counts (then random): The tied candidate with the least votes in the previous round loses the tie. If there is a tie in the previous round, the tie is broken randomly. Requires a random seed.

  * Previous round counts (then stop counting and ask): The tied candidate with the least votes in the previous round loses the tie. If there is a tie in the previous round, user is prompted to select any tied candidate to eliminate or, in multi-winner contests only, elect.

  * Use candidate order in the config file: Use the order of candidates in the config file to determine tiebreak results. Candidates lower in the list lose the tiebreaker.

  * Generate permutation: Generate a randomly ordered list of candidates in the contest. Candidates lower in the permutation lose the tiebreaker. Requires a random seed.

tarheel (approved):

PR #515, opened 2020-09-28
Creates combo text box / check box input for `maxSkippedRanksAllowed` and `maxRankingsAllowed`

### LINKED ISSUE #498: Combo text box / check box input for `maxSkippedRanksAllowed` and `maxRankingsAllowed`
* `maxRankingsAllowed` should default to having the "Maximum" check box selected
* `maxSkippedRanksAllowed` should default to having the "Unlimited" check box unselected (and 1 in text box)
* Checking either check box should disable and clear the text box
* More to consider from @chughes297:
  > I vaguely remember discussing a fill-in + checkbox option with Lynn but don't remember where we wound up. The more I think about number of rankings the more I think we should require people to put in the number of rankings, because to some places "max" would be all candidates + a write-in (or write-ins) while in others "max" may mean just all candidates. We could clear that up with instructions/hints/etc but it seems simpler to me to just ask people how many rankings are permitted in the contest and require numeric entry. 

  > I'm neutral on splitting up the skipped rankings question, it's just a matter of asking the questions in the right order . Probably would first ask about whether voters can skip as many rankings as they want, then if the box is left unchecked have them indicate how many rankings a voter can skip? I'm having a hard time thinking of how to phrase these questions... Maybe "Can voters skip unlimited rankings in a row and still have later rankings count" for checkbox and "How many consecutive skipped ranks are allowed" for the entry field?

HEdingfield:
Adds check boxes next to `textFieldMaxSkippedRanksAllowed` and `textFieldMaxRankingsAllowed` to substitute "unlimited" and "max" values in, respectively (fixes #498).

### DIFF
--- a/src/main/java/network/brightspots/rcv/GuiConfigController.java
+++ b/src/main/java/network/brightspots/rcv/GuiConfigController.java
@@ -256,6 +256,29 @@ public class GuiConfigController implements Initializable {
     return textField.getText() != null ? textField.getText().trim() : "";
   }
 
+  private static String loadTxtFileIntoString(String configFileDocumentationFilename) {
+    String text;
+    try {
+      text =
+          new BufferedReader(
+              new InputStreamReader(
+                  Objects.requireNonNull(
+                      ClassLoader.getSystemResourceAsStream(configFileDocumentationFilename))))
+              .lines()
+              .collect(Collectors.joining("\n"));
+    } catch (Exception exception) {
+      Logger.log(
+          Level.SEVERE,
+          "Error loading text file: %s\n%s",
+          configFileDocumentationFilename,
+          exception.toString());
+      text =
+          String.format(
+              "<Error loading text file: %s>", configFileDocumentationFilename);
+    }
+    return text;
+  }
+
   private String getOvervoteRuleChoice() {
     String overvoteRuleString = OvervoteRule.RULE_UNKNOWN.toString();
     if (radioOvervoteAlwaysSkip.isSelected()) {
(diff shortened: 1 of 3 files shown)


src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -1063,15 +1073,19 @@ public LocalDate fromString(String string) {
       setWinningRulesDefaultValues();
       switch (getWinnerElectionModeChoice(choiceWinnerElectionMode)) {
         case STANDARD_SINGLE_WINNER -> {
-          textFieldMaxRankingsAllowed.setDisable(false);
+          checkBoxMaxRankingsAllowedMax.setDisable(false);
+          checkBoxMaxRankingsAllowedMax
+              .setSelected(ContestConfig.SUGGESTED_MAX_RANKINGS_ALLOWED_MAXIMUM);
```
tarheel, line 1078: I think it would be good practice to factor out all of the parts (like these two lines that you've added) that apply to all of the cases instead of repeating them across each case. You could just put them before the switch block.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -1253,8 +1281,27 @@ private void loadConfig(ContestConfig config) {
         .setText(rules.multiSeatBottomsUpPercentageThreshold);
     textFieldDecimalPlacesForVoteArithmetic.setText(rules.decimalPlacesForVoteArithmetic);
     textFieldMinimumVoteThreshold.setText(rules.minimumVoteThreshold);
-    textFieldMaxSkippedRanksAllowed.setText(rules.maxSkippedRanksAllowed);
-    textFieldMaxRankingsAllowed.setText(rules.maxRankingsAllowed);
+    if (rules.maxSkippedRanksAllowed
+        .equalsIgnoreCase(ContestConfig.MAX_SKIPPED_RANKS_ALLOWED_UNLIMITED_OPTION)) {
+      checkBoxMaxSkippedRanksAllowedUnlimited.setSelected(true);
```
tarheel, line 1286: Do you need `checkBoxMaxSkippedRanksAllowedUnlimited.setDisable(false)` here also? Not sure why this is different from what you're doing for `checkBoxMaxRankingsAllowedMax` on line 1297.

tarheel (approved):

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Ahh yeah, good point, I've been wanting to do this for a while but overlooked the fact that the switch wasn't driving the whole `onAction` thing. Done.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: No, I don't think it's needed. The difference is this field will never have the check box disabled, whereas the other one you mentioned is disabled when no `WinnerElectionMode` is selected.

Issue #508, opened 2020-09-21
Move UWI, Overvote, Undervote labels and "treat blank as UWI" boolean to a per-CVR level


HEdingfield:
Per GUI mock-ups [here](https://a89u6z.axshare.com/#g=1&p=cvr_files_es_s__fields_showing). This will require some version migration logic too.

PR #518, opened 2020-09-30
update validation and documentation for 4 fields that moved to CVR level

tarheel:
Following up on @HEdingfield's requests.

This should complete the (known) work for #508.


HEdingfield (approved):
Pushed a couple fixes. LGTM now!

Issue #494, opened 2020-08-11
try converting config strings back to camelCaseValues


tarheel:
See whether there's a clean way to use camelCaseValues for winner election mode, overvote rule, and tie-break mode while still showing friendlier strings in the UI.

PR #519, opened 2020-09-30
camelCase strings for enums in config files

tarheel:
This addresses #494.

It's mostly transparent to the user, but the one minor regression in the UI is that the table of existing CVR sources shows the internal label for each source's provider instead of the UI label. Maybe there's an easy way to have it show the UI label instead and also make this field read-only?


src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1096,29 +1097,35 @@ private boolean undeclaredWriteInsEnabled() {
...
+    ESS("ess", "ES&S"),
+    HART("hart", "Hart"),
+    PROVIDER_UNKNOWN("providerUnknown", "Provider unknown");
+
+    private final String internalLabel;
+    private final String uiLabel;
+
+    Provider(String internalLabel, String uiLabel) {
+      this.internalLabel = internalLabel;
+      this.uiLabel = uiLabel;
```
HEdingfield, line 1112: Nit: might be better to call these `guiLabel`, since the CLI is also technically a UI.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -245,18 +245,21 @@
...
+        .getByInternalLabel(choiceBox.getValue().getInternalLabel())
         : Provider.PROVIDER_UNKNOWN;
   }
 
   private static WinnerElectionMode getWinnerElectionModeChoice(
       ChoiceBox<WinnerElectionMode> choiceBox) {
     return choiceBox.getValue() != null ? WinnerElectionMode
-        .getByLabel(choiceBox.getValue().toString()) : WinnerElectionMode.MODE_UNKNOWN;
+        .getByInternalLabel(choiceBox.getValue().getInternalLabel())
+        : WinnerElectionMode.MODE_UNKNOWN;
```
HEdingfield, line 257: Because this calls `.toString()` it'd make it "Mode unknown" in the config file instead of "modeUnknown". Related to my other comment.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -585,7 +588,7 @@ public void buttonAddCvrFileClicked() {
             getTextOrEmptyString(textFieldCvrIdCol),
             getTextOrEmptyString(textFieldCvrPrecinctCol),
             getTextOrEmptyString(textFieldCvrOvervoteDelimiter),
-            getProviderChoice(choiceCvrProvider).toString(),
+            getProviderChoice(choiceCvrProvider).getInternalLabel(),
```
HEdingfield, line 591: Agree that it'd be nice to make this display the GUI label instead, and also disable the field for editing (or even all rows in the CVR table in general).

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
```diff
@@ -228,11 +228,11 @@ Config file must be valid JSON format. Examples can be found in the test_data fo
         example: 15
         values: [max, 1..1000000]
 
-      "randomSeed" required if tiebreakMode is "Random", "Previous round counts (then random)", or "Generate permutation"
+      "randomSeed" required if tiebreakMode is "random", "previousRoundCountsThenRandom", or "generatePermutation"
         the integer seed for the application's pseudorandom number generator
         value: [-140737488355328..140737488355327]
 
-      "multiSeatBottomsUpPercentageThreshold" required if winnerElectionMode is "Bottoms-up using percentage threshold" and numberOfWinners is 0
+      "multiSeatBottomsUpPercentageThreshold" required if winnerElectionMode is "bottomsUpUsingPercentageThreshold" and numberOfWinners is 0
```
HEdingfield, line 235: Can you please Ctrl-Shift-F through the entire project (including comments and this .txt file; with the exception of the hint .txt files since those should show the GUI labels) and replace all instances of every single enum value you're modifying in this PR? Looks like you missed at least one instance of "Bottoms-up using percentage threshold" here.

src/main/java/network/brightspots/rcv/TieBreak.java
```diff
@@ -295,8 +295,8 @@ private String doPreviousRounds(List<String> tiedCandidates) throws TabulationCa
               candidatesInContention.size() > 2 ? "among" : "between",
               Utils.listToSentenceWithQuotes(candidatesInContention),
               tieBreakMode == TieBreakMode.PREVIOUS_ROUND_COUNTS_THEN_RANDOM
-                  ? "random"
-                  : "interactive");
+                  ? TieBreakMode.RANDOM.getInternalLabel()
+                  : TieBreakMode.INTERACTIVE.getInternalLabel());
```
HEdingfield, line 299: Good catch! I wonder if there are any more of these hard-coded strings lurking around in the codebase?

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
tarheel: Hmm. I already tried to search through everything and update it. I'm confused by your comment -- I _did_ catch this and update it, didn't I?

src/main/java/network/brightspots/rcv/ContestConfig.java
tarheel: that's fair

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1096,29 +1097,35 @@ private boolean undeclaredWriteInsEnabled() {
...
-          .filter(v -> v.label.equals(labelLookup))
+          .filter(v -> v.internalLabel.equals(labelLookup))
           .findAny()
           .orElse(null);
     }
 
     @Override
     public String toString() {
-      return label;
+      return uiLabel;
```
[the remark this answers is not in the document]
tarheel: yeah, I think you're right

src/main/java/network/brightspots/rcv/GuiConfigController.java
tarheel: will disable all editing

src/main/java/network/brightspots/rcv/TieBreak.java
tarheel: I believe not, but I'll check a bit more...

src/main/resources/network/brightspots/rcv/config_file_documentation.txt
HEdingfield: There's one below it, but GH acts dumb if I comment on lines outside the diff.

src/main/java/network/brightspots/rcv/ContestConfig.java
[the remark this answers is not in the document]
tarheel: Actually, I'm going to punt on this because Java sucks and doing this would significantly complicate the ChoiceBox code.

src/main/java/network/brightspots/rcv/GuiConfigController.java
tarheel: Reading this comment again, I realize I don't understand it. This method returns a `WinnerElectionMode`, and then in `createRawContestConfig`, we call `.getInternalLabel()` on that object. So what do you mean when you say that it calls `.toString()`?

tarheel:
OK, updated. Do you think we should disable editing for the candidate rows also?

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Good point, thought it was returning a String.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Sounds good, please note this clearly in the squash commit log!

HEdingfield (approved):
LGTM!

Please note clearly in the squash commit log that we disabled editing the CVR Files table in this PR.

And yes, I think we should disable that functionality for the Candidate table as well, but in a separate PR.

PR #520, opened 2020-10-01
don't log config file contents

tarheel:


HEdingfield:
Can we run this by @chughes297 or @gngilbert before merging?

PR #522, opened 2020-10-01
make candidate rows not editable

tarheel:


HEdingfield (approved):
Please get rid of the note about editing in the FXML file as well.

Issue #451, opened 2020-05-04
better handling of overvotes/UWIs on conversion to CDF


tarheel:
Two things:
1. The documentation should note that we change the labels, e.g. "overvote" -> "cs-overvote."
2. We should try to match the overvote and UWI labels against both the candidate name and the ID.
3. We need to register these as valid selection options so that the generated file is still technically valid.

PR #523, opened 2020-10-01
register explicit overvote as valid candidate/contest selection in CDF output when needed

tarheel:
I decided this was all that's worth doing for #451.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Or, again, if you're compiling and running using Gradle:
 
 `$ gradlew run --args="-cli path/to/config  convert-to-cdf"`
 
-Note: if you convert a source to CDF and that source uses an overvoteLabel or an undeclaredWriteInLabel, you'll need to update each label in your new CDF-based config to match the ContestSelection ID assigned to the label in the generated CDF file.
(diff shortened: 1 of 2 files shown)


moldover (approved):
Good catch!  I did not think to do this when I was working on the ResultsWriter before.

README.md
```diff
@@ -95,6 +95,8 @@ Or, again, if you're compiling and running using Gradle:
 
 `$ gradlew run --args="-cli path/to/config  convert-to-cdf"`
 
+Note: if you convert a source to CDF and that source uses an overvoteLabel or an undeclaredWriteInLabel, you'll need to update each label in your new CDF-based config to match the ContestSelection ID assigned to the label in the generated CDF file.
```
HEdingfield, line 98: Kind of hard for me to follow this as someone who is coming in with uneducated eyes. So this means, if they set overvoteLabel to "overvote" and UWI label to "UWI", they need to manually go through the resulting .json file and Ctrl-R to replace each instance with "overvote-mycontest" and "UWI-mycontest", or what exactly? Could you maybe give an example in the doc here?

HEdingfield: Also, if this is the case, isn't it something we could do programmatically during the conversion process?

tarheel: No, what I'm saying here is they'll need to set a different label in the _config_ that they use with the new CDF source file. They shouldn't try to modify the generated CDF source file. I'll clarify this text a bit.

PR #524, opened 2020-10-02
fix precinct test

tarheel:


HEdingfield:
This is super quick, so I'll just roll it into my current PR to avoid clutter. STEALSIES!

PR #525, opened 2020-10-02
fix user and computer name logging

### LINKED ISSUE #521: computer name and user name aren't logged properly on macOS
Two issues:
- It should check USER in addition to USERNAME
- Java doesn't seem to be getting the env vars anyway

tarheel:
fixes #521


HEdingfield (approved):
Yes, oh yes indeed!

PR #526, opened 2020-10-02
Final 1.2.0 polish

HEdingfield:
Recommend reviewing one commit at a time, lest ye dissolve into a Swirling Miasma of Madness.


src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -187,7 +186,7 @@ void parseXml(List<CastVoteRecord> castVoteRecords)
       if (contest.CVRContestSelection != null) {
         for (CVRContestSelection cvrContestSelection : contest.CVRContestSelection) {
           if (cvrContestSelection.Status != null && cvrContestSelection.Status
-              .equals(STATUS_NEEDS_ADJUDICATION)) {
+              .equals("needs-adjudication")) {
```
tarheel: what warning is this addressing? you can't define a constant unless it's used in multiple places?

tarheel (approved):
Looks good! Thanks for dealing with all these nits. I know @moldover was really hoping to do them himself, though.

moldover (commented):
Nice.  There's another VVSG requirement on variable names (5.2.5 paragraph c) that only index loops are allowed to have single-character variable names.  So that's why we have 'exception' everywhere.  If you can put those back, this looks good.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
HEdingfield: Yeah, the warning was recommending inlining a value that's only referenced once, which is consistent with Google's best practices as well (since it improves readability).

PR #527, opened 2020-10-03
Minor cleanups

moldover:
A few small things here:
Correct Readme.md regarding usage of overvote and uwi strings.
Throw if a candidate object can't be found during CVR parsing (this caused a NPE when parsing a CDF file which had been exported before (#523) was fixed.  @gngilbert encountered this problem
Finally, the CDF class (type) names have been fixed.  I verified and updated regressions tests impacted by this.  I just noticed this was wrong.

### DIFF
--- a/src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
+++ b/src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
@@ -363,6 +363,10 @@ class CommonDataFormatReader {
       for (Object cvrContestSelectionObject : cvrContestSelections) {
         HashMap cvrContestSelection = (HashMap) cvrContestSelectionObject;
         String contestSelectionId = (String) cvrContestSelection.get("ContestSelectionId");
+        if (!contestSelections.containsKey(contestSelectionId)) {
+          Logger.severe("ContestSelection \"%s\" from CVR not found!", contestSelectionId);
+          throw new CvrParseException();
+        }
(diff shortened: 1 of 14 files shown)


src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -363,6 +363,10 @@ void parseJson(List<CastVoteRecord> castVoteRecords)
       for (Object cvrContestSelectionObject : cvrContestSelections) {
         HashMap cvrContestSelection = (HashMap) cvrContestSelectionObject;
         String contestSelectionId = (String) cvrContestSelection.get("ContestSelectionId");
+        if (!contestSelections.containsKey(contestSelectionId)) {
+          Logger.severe("ContestSelection \"%s\" from CVR not found!", contestSelectionId);
```
HEdingfield, line 367: Will the user know what a "ContestSelection" is? Cause I don't...

HEdingfield (approved):

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
moldover: They "should".  ContestSelection is CDF-speak for "one of the things you can pick for a particular contest".  I think ClearVote calls it "option" and more often we just say "candidate" because that's the typical domain.  CDF can also describe ballot measure selections, retention contest selections, and party contest selections.  It is truly a kitchen sink format.

Issue #528, opened 2020-10-04
Open Help Menu does not work on Mac


moldover:
When running from the distribution build using the GUI, the Open Help Menu option fails with an error.
We need to convert the internal resource URL into a full file path, and specify the full path to the open utility (we are not running via a shell, so PATH is not defined, and simply calling 'open' fails to find the open utility)

HEdingfield:
Doesn't work in Windows either :(

![image](https://user-images.githubusercontent.com/8259446/95025211-f699a180-063c-11eb-9f19-7b8f0a857668.png)

moldover:
Turns out we are storing the file as a resource (packed with other resources) which is not available as a file without unpacking or some other strategy.

moldover:
I pushed updates to halp_menu: it now simply shows a message directing the user to the config_file_documentation file.  Not a great user experience, but it’s an improvement.  IMO the big structural change we need would involve writing a comprehensive help document with context links from various UI elements.

PR #529, opened 2020-10-04
fix for help menu not opening on mac #528

moldover:
Convert help.txt file resource URL to file _path_ and use full path to 'open' util does the trick for me.  Pretty sure this should work on other platforms, but worth checking when launching from the OS i.e. not using intelliJ.


src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -307,20 +307,21 @@ private String getOvervoteRuleChoice() {
...
     } else if (Utils.IS_OS_MAC) {
-      command = String.format("open %s", helpFileUrl);
+      command = String.format("/usr/bin/open %s", helpFile);
     } else if (Utils.IS_OS_LINUX) {
-      command = String.format("xdg-open \"%s\"", helpFileUrl);
+      command = String.format("xdg-open \"%s\"", helpFile);
     } else {
       Logger.info("Unable to determine operating system. Try opening the documentation "
-          + "manually at: %s", helpFileUrl);
+          + "manually at: %s", helpFile);
```
tarheel, line 320: So this won't work on my Amiga?

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -307,20 +307,21 @@ private String getOvervoteRuleChoice() {
...
     }
     if (command != null) {
       try {
+        Logger.info("Running: %s", command);
```
tarheel, line 324: I don't think we need this to be logged.

tarheel (approved):

tarheel:
LGTM, but will let @HEdingfield weigh in also.

HEdingfield (approved):

src/main/java/network/brightspots/rcv/GuiConfigController.java
moldover: Ok, yeah this is removed.

src/main/java/network/brightspots/rcv/GuiConfigController.java
moldover: No Newton either :/

tarheel: https://www.youtube.com/watch?v=u6qxixgQJ4M

PR #530, opened 2020-10-04
UI tweaks

tarheel:
image

### DIFF
--- a/src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
+++ b/src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
@@ -539,7 +539,7 @@
             <VBox>
               <VBox maxWidth="460.0" styleClass="bordered-box">
                 <HBox spacing="4.0">
-                  <Label prefWidth="100.0" text="Overvote Rule *"/>
+                  <Label prefWidth="120.0" text="Overvote Rule *"/>
                   <VBox spacing="4.0">
                     <fx:define>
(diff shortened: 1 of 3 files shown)


src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
```diff
@@ -539,7 +539,7 @@
             <VBox>
               <VBox maxWidth="460.0" styleClass="bordered-box">
                 <HBox spacing="4.0">
-                  <Label prefWidth="212.0" text="Overvote Rule *"/>
+                  <Label prefWidth="100.0" text="Overvote Rule *"/>
```
HEdingfield, line 542: I'll need to check this in Linux tomorrow before approving.

HEdingfield: Yeah, this cuts things off in Linux. Please revert to what it was before.

HEdingfield: 120.0 also works ok, which is closer, but isn't in line with the other input box. Either that or 212 is ok with me.

HEdingfield (approved):

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
tarheel: OK, going with 120.

PR #532, opened 2020-10-05
change getByInternalLabel methods to return unknown value instead of null

### LINKED ISSUE #531: Config file with bad provider value fails with NPE and logging doesn't catch it
If you load the attached config file in the latest build, the console shows an NPE, but the log file doesn't record it (I've included in the .zip both the log file and the console output for comparison).

Furthermore, the GUI seems to behave erratically after this has happened (unresponsive CVR Files table, etc.).

Note: The config file is clearly bad because it has "ES&S" as a provider, when it should now be "ess", as well as a few other differences. It also has the version "1.2.0_alpha", so the migration logic isn't triggered. However, even if I change the version to "TEST" or "1.1.0" and the migration logic is triggered, it appears to just freak out when encountering the "ES&S" value.

[2013_minneapolis_mayor_config.zip](https://github.com/BrightSpots/rcv/files/5325654/2013_minneapolis_mayor_config.zip)

tarheel:
Fixes #531.

### DIFF
--- a/src/main/java/network/brightspots/rcv/Tabulator.java
+++ b/src/main/java/network/brightspots/rcv/Tabulator.java
@@ -1059,7 +1059,7 @@ class Tabulator {
       return Arrays.stream(OvervoteRule.values())
           .filter(v -> v.internalLabel.equals(labelLookup))
           .findAny()
-          .orElse(null);
+          .orElse(RULE_UNKNOWN);
     }
 
     @Override
@@ -1103,7 +1103,7 @@ class Tabulator {
       return Arrays.stream(TiebreakMode.values())
           .filter(v -> v.internalLabel.equals(labelLookup))
           .findAny()
-          .orElse(null);
+          .orElse(MODE_UNKNOWN);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -1103,7 +1103,7 @@ static TiebreakMode getByInternalLabel(String labelLookup) {
       return Arrays.stream(TiebreakMode.values())
           .filter(v -> v.internalLabel.equals(labelLookup))
           .findAny()
-          .orElse(null);
```
HEdingfield, line 1106: Nit: this might be a good time to name these in a more clear fashion: `OVERVOTE_RULE_UNKNOWN`, `TIEBREAK_MODE_UNKNOWN`, `WINNER_ELECTION_MODE_UNKNOWN`.

HEdingfield (approved):

src/main/java/network/brightspots/rcv/Tabulator.java
tarheel: Eh, let's live dangerously.

PR #534, opened 2020-10-06
Handle missing precinct data for older Dominion data sets #533

### LINKED ISSUE #533: Support older Dominion data sets
Older (pre 2020) Dominion data sets might not have precinct data.  For example, SF 2019 consolidated: https://sfelections.sfgov.org/november-5-2019-election-results-detailed-reports

In this case we need to handle missing precinct data gracefully.

moldover:
Fixes #533


src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -208,13 +215,17 @@ private void parseCvrFile(
...
+        if (precinctId != null && (this.precincts == null || !this.precincts
+            .containsKey(precinctId))) {
           Logger.severe("Precinct ID \"%d\" from CVR not found in manifest data!", precinctId);
           throw new CvrParseException();
         }
-        String precinct = this.precincts.get(precinctId);
+        String precinct = null;
+        if (this.precincts != null) {
+          precinct = this.precincts.get(precinctId);
+        }
```
HEdingfield, line 228: Nit: you could simplify this a bit:
`String precinct = this.precincts != null ? this.precincts.get(precinctId) : null;`

HEdingfield (commented):
Would you or the person who provided you with the data be able to cut 99% of the data out of the data set so it's closer to 5 MB? As it stands, it's not an official regression test because there's no inclusion in `TabulatorTests.java`, and it seems like it should be.

src/test/resources/network/brightspots/rcv/test_data/dominion_no_precinct_data/dominion_no_precinct_data_summary.json
```diff
@@ -0,0 +1,163 @@
+{
```
HEdingfield, line 1: I don't think this file is necessary, right?

src/test/resources/network/brightspots/rcv/test_data/dominion_no_precinct_data/dominion_no_precinct_data/BallotTypeContestManifest.json
```diff
@@ -0,0 +1,9 @@
+{
```
HEdingfield, line 1: Nit: going off the directory structure pattern for the other dominion tests, it'd be more consistent to name the folders:
`/dominion_no_precinct/dominion_no_precinct_input_data`

HEdingfield (approved):
Great, glad that worked out!

src/test/resources/network/brightspots/rcv/test_data/dominion_no_precinct_data/dominion_no_precinct_data_summary.json
moldover: ack!

Issue #536, opened 2020-10-08
Can't tabulate multiple CDF sources


moldover:

moldover:
closed via #537

PR #537, opened 2020-10-08
Fix for #536 - allow multiple CDF files

moldover:
Allow multiple CDF sources
Add regression test with multiple CDF files


tarheel (approved):
image

PR #543, opened 2020-12-15
fix minor bug with excluded candidates

### LINKED ISSUE #540: Eliminating too many candidates when excluding candidates in "eliminate down to two" version of single-winner RCV
This is a bug with the "eliminate down to two" version of single-winner RCV.

If you exclude one or more candidate(s) using the "exclude candidates" option, the tabulator counts all the way down to a single candidate - as in it eliminates every candidate but the winner in the final round. 

My guess is the problem is a mismatch/miscommunication between the total number of candidates in the Candidates tab, the number of candidates excluded, and the number of candidates number used by the "eliminate down to two" function. Attaching configs, CVR, and results from tests displaying this bug.

[excluded candidate bug configs & results](https://github.com/BrightSpots/rcv/files/5489841/excluded.candidate.bug.zip)

[Eastpointe CVR Cleaned.xlsx](https://github.com/BrightSpots/rcv/files/5489846/Eastpointe.CVR.Cleaned.xlsx)

tarheel:
This fixes #540.

I also updated one of the test cases so that it would have failed without this fix.

### DIFF
--- a/src/main/java/network/brightspots/rcv/ContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/ContestConfig.java
@@ -954,7 +954,7 @@ class ContestConfig {
   }
 
   int getNumCandidates() {
-    return getCandidateCodeList().size();
+    return getCandidateCodeList().size() - excludedCandidates.size();
   }
 
   boolean candidateIsExcluded(String candidate) {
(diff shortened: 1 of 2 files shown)


HEdingfield (approved):
This guy! 👍

PR #560, opened 2021-04-13
Add floorWinningThreshold option

NCGThompson:
Some jurisdictions, such as Portland, define "majority" as >= 1/2 + 1, and analogously for multi-winner threshold. This can be accommodated by configuring floorWinningThreshold to false.

As of writing, I haven't tested this with Gradle so it probably doesn't work correctly.


NCGThompson:
I don't actually understand understand where in the source code the raw.rules data are parsed from the config file. Can someone explain that to me?

NCGThompson:
For technical reasons related to git, I am closing this pull request.

Issue #565, opened 2021-04-21
build doesn't work with Java 16


tarheel:
According to SLI...

NCGThompson:
That is a Gradle problem. [Gradle 7.0](https://docs.gradle.org/7.0/release-notes.html) should fix this.

dddvision:
Not sure if this is the same error, but it seems related. I'm on macOS 11.6 with java version 1.8.0_301

---
./gradlew run

FAILURE: Build failed with an exception.

* What went wrong:
java.lang.UnsupportedClassVersionError: org/openjfx/gradle/JavaFXPlugin has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
> org/openjfx/gradle/JavaFXPlugin has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
---

NCGThompson:
@dddvision I think I know the problem. One fix is upgrading the Java on your computer. It's ancient (1.8). It's been months since I last did that on macOS many months ago but you can Google it.

dddvision:
Hi @NCGThompson, I've gone to the Java Control Panel and upgraded to the recommended version for macOS Big Sur, which is Java 8 Update 311. That version is not new enough to solve the problem. I understand that it is possible to install a newer version of Java, but instead I am just poking around with the release version. It seems that the purpose of this project is more about delivering a production-level tabulator than it is about making a tool for the public to play with to discover how RCV works. I'm interested in the latter, so I will probably take a different approach. I think that the algorithm and its demonstration needs to be simplified in order to sell the public on it.

HEdingfield:
Agreed, closed by #580.

PR #568, opened 2021-05-04
Specify JDK 14 or 15 in README

NCGThompson:
Currently, the Tabulator cannot be built with JDK 16 or higher, so it makes sense to say "14 or 15" rather than "14 or higher".

See issue #565.


tarheel (approved):
Good catch. We just ran into this recently ourselves.

PR #569, opened 2021-05-18
Add support for multi-file Dominion format

catrope:
Starting with the March 2020 election, San Francisco has started
publishing their Dominion CVR data as many small files named
CvrExport_1.json, CvrExport_2.json, etc. rather than one big
CvrExport.json file.

To support this new format, first look for CvrExport.json, but if that
doesn't exist, look for CvrExport_1.json. If that does exist, read all
CvrExport_%d.json files in numerical sequence, until we hit one that
doesn't exist.


moldover (commented):
Hi @catrope.  Thanks for the PR!  I've been traveling so I apologize I'm late responding.  Looks pretty good and it will be great to support the new SF format.  It would be really good to have a regression test which parses the new format.  Is it possible to add one?  How large would the data be to use actual published SF cvr data?

src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -173,9 +174,8 @@ void readCastVoteRecords(List<CastVoteRecord> castVoteRecords, String contestId)
     }
   }
 
-  // parse the given file into a List of CastVoteRecords for tabulation
-  private void parseCvrFile(
-      String filePath, List<CastVoteRecord> castVoteRecords, String contestIdToLoad) {
+  // parse the CVR file or files into a List of CastVoteRecords for tabulation
+  private void parseCvrFile(List<CastVoteRecord> castVoteRecords, String contestIdToLoad) {
```
moldover, line 178: I would swap the function names parseCvrFile / gatherCvrsForContest since 'gather is really handling per-file logic now.

src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -190,125 +190,156 @@ private void parseCvrFile(
...
+        HashMap json = JsonParser.readFromFile(singleCvrPath.toString(), HashMap.class);
+        gatherCvrsForContest(json, castVoteRecords, contestIdToLoad, contestIdToCandidateCodes);
+      } else if (firstCvrPath.toFile().exists()) {
+        int recordsParsed = 0, recordsParsedAtLastlog = 0;
+        int cvrSequence = 1;
+        Path cvrPath = Paths.get(manifestFolder, String.format(CVR_EXPORT_PATTERN, cvrSequence));
+        while (cvrPath.toFile().exists()) {
+          HashMap json = JsonParser.readFromFile(cvrPath.toString(), HashMap.class);
+          recordsParsed += gatherCvrsForContest(json, castVoteRecords, contestIdToLoad, contestIdToCandidateCodes);
+          if (recordsParsed - recordsParsedAtLastlog > 50000) {
```
moldover, line 205: Doesn't the existing feedback logging (now moved to gatherCvrsForContest) handle this already?

src/main/java/network/brightspots/rcv/DominionCvrReader.java
moldover, line 178: I would swap the function names parseCvrFile / gatherCvrsForContest since 'gather is really handling per-file logic now.

src/main/java/network/brightspots/rcv/DominionCvrReader.java
moldover, line 205: I think cvr count logging logic should stay as it was (use the length of the castVoteRecords list) and only exist in one place - bottom of the gatherCvrsForContest function.  Is there a reason this won't work anymore?

src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -190,125 +190,156 @@ private void parseCvrFile(
...
-                    + "Tabulator ID: %s Batch ID: %s Record ID: %d",
-                tabulatorId, batchId, recordId);
-            continue;
+      Path singleCvrPath = Paths.get(manifestFolder, CVR_EXPORT);
+      Path firstCvrPath = Paths.get(manifestFolder, String.format(CVR_EXPORT_PATTERN, 1));
+      if (singleCvrPath.toFile().exists()) {
+        HashMap json = JsonParser.readFromFile(singleCvrPath.toString(), HashMap.class);
+        gatherCvrsForContest(json, castVoteRecords, contestIdToLoad, contestIdToCandidateCodes);
+      } else if (firstCvrPath.toFile().exists()) {
+        int recordsParsed = 0, recordsParsedAtLastlog = 0;
```
moldover, line 199: There are a bunch of... pedantic coding standards from the VVSG which we need to follow.  So we are not allowed to initialize multiple variables in a single statement like this.  I think these can go away though if the cvr count logging logic is re-consolidated.

moldover (changes requested):

tarheel:
Hi @catrope, we haven't forgotten about this PR! Are you interested in finishing it, or do you want us to update it? We're actually hoping to make use of this functionality so that we can use our application to verify some SF results.

catrope:
Hi! So sorry for forgetting about this last year. Yes, I am interested in finishing this, probably within the next week or so.

src/main/java/network/brightspots/rcv/DominionCvrReader.java
catrope: It's difficult to consolidate this the way you asked, because we only want to log every 50k records. In the multi-file case, the individual files are relatively small and contain fewer than 50k records, so logging just in parseCvrFile (the method that parses one individual file; renamed from gatherCvrsForContest per your other comment) would result in nothing ever being logged (or in the log being noisy if we log for every file parsed file). But in the single-file case, the file will contain more than 50k records, so we have to log in parseCvrFile, otherwise we'll have no progress logging and we'll just log once at the end.

src/main/java/network/brightspots/rcv/DominionCvrReader.java
catrope: Done

src/main/java/network/brightspots/rcv/DominionCvrReader.java
catrope: Done

catrope:
> Looks pretty good and it will be great to support the new SF format. It would be really good to have a regression test which parses the new format. Is it possible to add one? How large would the data be to use actual published SF cvr data?
The SF data is huge. The Nov 2020 election data was split over 24,940 JSON files, adding up to 4.2 GB. The Feb 2022 special election was "only" 1902 files totaling 200 MB, and that one didn't have a meaningful RCV contest on the ballot.

Instead, I'll add tests with a manually created example of a multi-file election with only 3 files, including one where there are multiple Cards in one Session.

catrope:
I've added a test case based on 4 files from the SF Nov 2020 election, 2 of which have the multi-Card format. Using only four files limited the damage to 3.6 MB. All pre-existing test data combined was 60 MB (now 64 MB with this addition).

HEdingfield (changes requested):
Heya @catrope, thanks for working on this, and for adding the tests! Unfortunately, it looks like you were working off a fairly old version. Could you pull the latest changes from develop into your branch and sort out the merging? The biggest change that will affect you is the fact that we've moved all our test data into a separate repo. Check out PR #576, in particular [here](https://github.com/BrightSpots/rcv/pull/576#issuecomment-945015788).

HEdingfield:
@catrope would you also be able to add an automated test that actually uses the data you provided in the process?

src/main/java/network/brightspots/rcv/DominionCvrReader.java
moldover: Got it - thanks for the explanation.

moldover:
@catrope - Code changes look solid and the new test is great.   Thank you.  As @HEdingfield mentions, we work off of `develop` branch now, and the latest develop code moves all test_data into a submodule -- in part so that we can accommodate larger data sets.  But it also means more commits to get things where they need to be.  But I can totally help with that if you'd like.  Sorry about the confusion.  We need to get better about communicating and onboarding.  I think you are the first outside contributor!   

I believe something like this should happen:

1) delete new test data from dominion-multi-file (make a copy first)
2) add test data to the test_data repo: https://github.com/artoonie/rcvtests via another PR which we can quickly approve
3) merge latest develop -> dominion-multi-file
4) pin test_data to your new 
5) add lines into TabulatorTests.java to execute your test there
6) we approve and merge dominion-multi-file -> develop

moldover:
@HEdingfield I went ahead and merged this with the latest develop, including the new test files.  Should be good to go but just wanted to get your eyeballs on this too :)

HEdingfield (approved):
Awesome!

moldover:
@catrope we finally pushed your work here.  Thanks so much for your contribution!!  

We are also putting together a contributors guide which will hopefully help make things cleaner in the future: https://github.com/BrightSpots/rcv/blob/license_updates_551/LICENSE.html

PR #575, opened 2021-08-10
Use submodules and run CI on each push

artoonie:
1. Move all the tests to a separate submodule. (I would recommend forking the rcvtests repo so brightspots owns that, too, rather than my account).
2. Run the existing tests, including the linting, using Github Actions on each push


HEdingfield:
Hi Armin, thanks for making a pass on this, but would appreciate extra context on this. Could you file an issue outlining the benefits / these proposed changes?

My concern is this would break our current workflow where we can also run our tests locally, but maybe that's unfounded?

Also, I just changed the merge into `develop` because we never push to `master` directly, and it appears there are some merge conflicts you'll need to resolve.

artoonie:
Sure, the goals of splitting the tests into a separate repo is to:
1. Allow other RCV tabulators to test against the same test set (which helps those projects)
2. Allow other tabulators to add to the test set (which helps this project)

Here's a diagram of the overall vision: 
image

(I discussed this briefly over email a few months back.)

As for running the tests - it doesn't change anything you do locally, it just runs those same tests on github. That way, if you forget to run locally you can still guarantee passing tests on `develop` or `main`. (In particular, I see another PR which looks like it failed linting. This would have notified the author before the author made the PR.)

I'll fix the merge conflicts shortly.

artoonie:
#576 has removed the conflicts

PR #578, opened 2021-11-07
Support for Java 17

kouellette:
Added support to build with java 17, the most recent LTS version of java. Gradle 7.3 introduces support for java 17 but doesn't have an official release out just yet. This uses a release candidate for the time being but I would suggest waiting for the official release and updating the PR if you choose to merge. 

~ Kyle


HEdingfield:
Thanks @kouellette! I don't think a reviewer was ever assigned so this slipped off our radar. I'll merge it into a feature branch and update the other dependencies as well.

PR #587, opened 2022-05-02
Add spotbugs and correct found problems

tamird:
See individual commits.


HEdingfield:
Hi @tamird, thanks for working on this stuff! So I assume you ran all the tests locally and they passed, but did you also run the GUI locally and try loading up some test files and tabulating them? I ask because you did some things that I believe might break the GUI (i.e. changing classes from `public` to package-private.

tamird:
Right you are; reverted some of those changes (and added doc comments instead to appease the style checker).

.github/workflows/test.yml
```diff
@@ -16,4 +16,4 @@ jobs:
           java-version: '17'
           distribution: 'temurin'
       - name: Test with Gradle
-        run: ./gradlew build
+        run: ./gradlew check
```
tarheel, line 19: I'm curious why this was `build` previously and what the implications are of this change.

src/main/java/network/brightspots/rcv/ClearBallotCvrReader.java
```diff
@@ -48,7 +49,7 @@ void readCastVoteRecords(List<CastVoteRecord> castVoteRecords, String contestId)
     // map for tracking unrecognized candidates during parsing
     Map<String, Integer> unrecognizedCandidateCounts = new HashMap<>();
     try {
-      csvReader = new BufferedReader(new FileReader(this.cvrPath));
+      csvReader = new BufferedReader(new FileReader(this.cvrPath, StandardCharsets.UTF_8));
```
tarheel, line 52: just curious: what is the benefit of this?

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -129,165 +129,168 @@ void parseXml(List<CastVoteRecord> castVoteRecords)
...
-
-    // Find the Contest we are tabulating:
-    // Note: Contest is different from CVRContest objects which appear in CVRSnapshots
-    Contest contestToTabulate = null;
-    for (Election election : cvrReport.Election) {
-      for (Contest contest : election.Contest) {
-        if (contest.Name.equals(this.contestId)) {
-          contestToTabulate = contest;
-          break;
+    try (FileInputStream inputStream = new FileInputStream(filePath)) {
```
tarheel, line 132: Oh, interesting. I never knew about this feature (try with resources).

src/main/java/network/brightspots/rcv/HartCvrReader.java
```diff
@@ -125,9 +124,6 @@ void readCastVoteRecord(List<CastVoteRecord> castVoteRecords, Path path) throws
           Logger.info("Parsed %d cast vote records.", castVoteRecords.size());
         }
       }
-    } catch (Exception exception) {
-      Logger.severe("Error parsing cast vote record:\n%s", exception);
-      throw exception;
     }
```
tarheel, line 127: Did this error logging get moved elsewhere?

src/main/java/network/brightspots/rcv/JsonParser.java
```diff
@@ -37,31 +38,32 @@ class JsonParser {
   // param: logsEnabled display log messages
   // returns: instance of the object parsed from json or null if there was a problem
   private static <T> T readFromFile(String jsonFilePath, Class<T> valueType, boolean logsEnabled) {
-    T createdObject;
-    try {
-      FileReader fileReader = new FileReader(jsonFilePath);
+    try (FileReader fileReader = new FileReader(jsonFilePath, StandardCharsets.UTF_8)) {
       ObjectMapper objectMapper = new ObjectMapper();
-      createdObject = objectMapper.readValue(fileReader, valueType);
+      return objectMapper.readValue(fileReader, valueType);
```
tarheel, line 43: as discussed in the issue, we're not allowed to have multiple returns :(

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -137,37 +137,29 @@ private static String generateCvrSnapshotId(String cvrId, Integer round) {
...
-    return id;
+    return cdfCandidateCodeToCandidateId.computeIfAbsent(code,
+        c -> String.format("c-%s", sanitizeStringForOutput(c).toLowerCase()));
   }
 
   // Instead of a map from rank to list of candidates, we need a sorted list of candidates
   // with the ranks they were given. (Ordinarily a candidate will have only a single rank, but they
   // could have multiple ranks if the ballot duplicates the candidate, i.e. assigns them multiple
-  // ranks.
+  // ranks).
```
tarheel, line 153: super nit: close parens should be after the period

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -280,17 +270,17 @@ void generatePrecinctSummaryFiles(
       Map<String, Integer> numBallotsByPrecinct)
       throws IOException {
     Set<String> filenames = new HashSet<>();
-    for (String precinct : precinctRoundTallies.keySet()) {
-      String precinctFileString = getPrecinctFileString(precinct, filenames);
+    for (var entry : precinctRoundTallies.entrySet()) {
```
tarheel, line 273: I'm learning a lot in this PR! I didn't know about `var` either. Seems like a nice way to reduce the verbosity a bit. But it's not clear to me why you're using it here but not in some other places, e.g. a bit below here in `updateCandidateNamesInTally`.

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -142,14 +143,14 @@ void tabulate() {
     ContestConfig config = ContestConfig.loadContestConfig(configPath);
     checkConfigVersionMatchesApp(config);
     boolean tabulationSuccess = false;
-    //noinspection ConstantConditions
-    if (config != null && config.validate() && setUpLogging(config)) {
+    if (config.validate() && setUpLogging(config)) {
```
tarheel, line 146: Why is the null check unnecessary here? It looks like it's possible for `loadContestConfig` to return a null value.

tarheel (commented):
Thanks for working on this! I just left a handful of questions inline.

.github/workflows/test.yml
tamird: See the commit message where this changed: https://github.com/BrightSpots/rcv/pull/587/commits/f9026074f8da97f70fe0692c13b5d17c7ab8353d.

src/main/java/network/brightspots/rcv/ClearBallotCvrReader.java
tamird: See the commit message where this changed: https://github.com/BrightSpots/rcv/pull/587/commits/b07a23b3193aa1a9ab43b27eb63a7a344464c217.

In particular, this is:
> H I Dm: Found reliance on default encoding in network.brightspots.rcv.ClearBallotCvrReader.readCastVoteRecords(List, String): new java.io.FileReader(String)  At ClearBallotCvrReader.java:[line 51]

src/main/java/network/brightspots/rcv/ResultsWriter.java
tamird: Done.

src/main/java/network/brightspots/rcv/HartCvrReader.java
tamird: Oops, incorrect removal. Restored.

src/main/java/network/brightspots/rcv/ResultsWriter.java
tamird: I didn't know about `var` either! I only went back and changed this to `var` in those places where naming the full type produced a line that was too long.

src/main/java/network/brightspots/rcv/TabulatorSession.java
tamird: Yes, but checkConfigVersionMatchesApp already dereferences `config`, so it can't possibly be null by the time execution reaches this point.

src/main/java/network/brightspots/rcv/JsonParser.java
tamird: Reverted.

src/main/java/network/brightspots/rcv/HartCvrReader.java
tamird: Actually, I take that back: this is from SpotBugs:

> M D REC: Exception is caught when Exception is not thrown in network.brightspots.rcv.HartCvrReader.readCastVoteRecord(List, Path)  At HartCvrReader.java:[line 127]

src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -29,7 +29,7 @@
 
 @SuppressWarnings("WeakerAccess")
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class RawContestConfig {
+class RawContestConfig {
```
HEdingfield, line 32: Removing the `public` keywords throughout this class breaks loading config files. Please revert all changes in this file.

I recommend doing a full (manual) end-to-end test each time you're putting together a PR with big changes like this, in addition to all the automated tests. This entails booting up the GUI, loading a config file, running the tabulation, and checking that the results look as expected.

.idea/misc.xml
```diff
@@ -6,7 +6,7 @@
     </list>
   </component>
   <component name="ExternalStorageConfigurationManager" enabled="true" />
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_17" default="true" project-jdk-name="17" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_17" project-jdk-name="17" project-jdk-type="JavaSDK">
```
HEdingfield, line 9: Not sure what the significance of deleting this is?

config/spotbugs/exclude.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
HEdingfield, line 1: Cool to know when can do exclusions here!

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-bin.zip
```
HEdingfield, line 3: Nice catch! That should slim the package down a bit.

build.gradle
```diff
@@ -1,25 +1,26 @@
 plugins {
     id "application"
     id "checkstyle"
+    id "com.github.spotbugs" version "5.0.6"
```
HEdingfield, line 4: Cool integration. So how do we actually use this in practice? Before launching a new release we run both checkstyleMain / Test and spotbugsMain / Test from the IntelliJ Gradle console and fix whatever pops up?

build.gradle
```diff
@@ -35,15 +36,22 @@ application {
 
 // ### Checkstyle plugin settings
 checkstyle {
-    toolVersion = "8.36.2"
+    toolVersion = '10.2'
```
HEdingfield, line 39: Big thanks for updating all these, and the `google_checks` too!

rcv.iml
```diff
@@ -1,2 +1,370 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<module external.linked.project.id="rcv" external.linked.project.path="$MODULE_DIR$" external.root.project.path="$MODULE_DIR$" external.system.id="GRADLE" external.system.module.group="" external.system.module.version="unspecified" type="JAVA_MODULE" version="4" />
\ No newline at end of file
+<module relativePaths="true" type="JAVA_MODULE" version="4">
```
HEdingfield, line 2: What are the implications of this change? Is this just an IntelliJ file that sets the project up in a certain way?

src/main/java/network/brightspots/rcv/GuiApplication.java
```diff
@@ -27,7 +27,7 @@
 import javafx.stage.Stage;
 
 @SuppressWarnings("WeakerAccess")
```
HEdingfield, line 29: You might be able to get rid of this suppression now too.

src/main/java/network/brightspots/rcv/HartCvrReader.java
HEdingfield: Weird... if an `Exception` isn't thrown, why do we need to `try` at all? I'm a little suspicious of that Spotbugs warning.

src/main/java/network/brightspots/rcv/JsonParser.java
```diff
@@ -38,25 +39,28 @@ class JsonParser {
...
       ObjectMapper objectMapper = new ObjectMapper();
       createdObject = objectMapper.readValue(fileReader, valueType);
     } catch (JsonParseException | JsonMappingException exception) {
       if (logsEnabled) {
         Logger.severe(
-            "Error parsing JSON file: %s\n%s\n"
-                + "Check file formatting and values and make sure they are correct!\n"
-                + "It might help to try surrounding values causing problems with quotes (e.g. "
-                + " \"value\").\nSee config_file_documentation.txt for more details.",
+            """
```
HEdingfield, line 48: Cool, didn't know about this.

src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -25,6 +25,9 @@
 import java.util.ArrayList;
 import java.util.List;
 
+/**
+ * Main entry point to the RCV tabulator program.
```
HEdingfield, line 29: Main entry point to the RCV tabulator GUI.

src/main/java/network/brightspots/rcv/ResultsWriter.java
HEdingfield: This was introduced in JDK 10, but we never got around to using it much. One downside is that it can decrease readability, but I think tamird@'s use is good. I'd probably lean towards consistently doing it throughout each class it's introduced though (e.g. here and in `Tabulator`).

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -280,17 +270,17 @@ void generatePrecinctSummaryFiles(
...
+      String precinctFileString = getPrecinctFileString(entry.getKey(), filenames);
```
HEdingfield, line 274: Nit: Might improve readability to factor out `String precinct = entry.getKey()` (and similarly for `entry.getValue()`) and inline `precinct` in each of these spots.

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -136,9 +136,9 @@ Set<String> tabulate() throws TabulationCancelledException {
     // - If winnerElectionMode is "Bottoms-up using percentage threshold", we loop until all
     // remaining candidates have vote shares that meet or exceed that threshold.
     //
-    // At each iteration, we'll either a) identify one or more
+    // At each iteration, we'll either a. identify one or more
     // winners and transfer their votes to the remaining candidates (if we still need to find more
-    // winners), or b) eliminate one or more candidates and gradually transfer votes to the
+    // winners), or b. eliminate one or more candidates and gradually transfer votes to the
```
HEdingfield, line 141: Super nit: I'd revert these two back to "a)" and "b)".

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -366,7 +365,7 @@ private void setWinningThreshold(Map<String, BigDecimal> currentRoundCandidateTo
                   ? config.getNumberOfWinners()
                   : config.getNumberOfWinners() + 1);
       // If we use integers, we shouldn't use any decimal places.
-      // Otherwise we use the amount of decimal places specified by the user
+      // Otherwise, we use the amount of decimal places specified by the user
```
HEdingfield, line 368: Nit: please add . at end while you're already here.

src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -51,13 +52,10 @@ class TabulatorTests {
...
   private static boolean fileCompare(String path1, String path2) {
     boolean result = true;
-    FileReader fr1 = null;
-    FileReader fr2 = null;
-    try {
-      fr1 = new FileReader(path1);
-      BufferedReader br1 = new BufferedReader(fr1);
-      fr2 = new FileReader(path2);
-      BufferedReader br2 = new BufferedReader(fr2);
+    try (
```
HEdingfield, line 55: Nice!

HEdingfield (changes requested):

.idea/misc.xml
tamird: I'm not sure either; Intellij just did it, and it seemed reasonable.

gradle/wrapper/gradle-wrapper.properties
tamird: Wish I could take credit; one of the gradle commands did that.

build.gradle
tamird: See commit #2 - `./gradlew :check` will run all the relevant things, and as of this PR, that will be run on every PR as well.

src/main/java/network/brightspots/rcv/Tabulator.java
tamird: Ack; prefer to leave this way because grammar checking plugins in intellij consider these to be unclosed parentheses.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
tamird: It's a good one!

src/main/java/network/brightspots/rcv/HartCvrReader.java
tamird: I think it's complaining that the thing thrown is IOException, rather than Exception. Restored with that change and it's happy now.

src/main/java/network/brightspots/rcv/ResultsWriter.java
tamird: Done.

src/main/java/network/brightspots/rcv/RawContestConfig.java
tamird: Reverted this one, but I'd appreciate a manual test if you have the time, or else a pointer to instructions on exactly how to run it.

src/main/java/network/brightspots/rcv/ResultsWriter.java
tamird: Done.

src/main/java/network/brightspots/rcv/Tabulator.java
tamird: Done.

build.gradle
tamird: You're welcome!

rcv.iml
tamird: Yeah, I believe this is just Intellij configuration.

src/main/java/network/brightspots/rcv/GuiApplication.java
tamird: I'll try and tackle all these suppressions in a later change.

src/main/java/network/brightspots/rcv/Main.java
tamird: Done.

tamird:
Removed the warning suppressions as well; could not get that to produce a failure locally. Will revert some/all if CI is unhappy.

src/main/java/network/brightspots/rcv/RawContestConfig.java
HEdingfield: It looks like you only reverted this one line, but the whole file needs to be reverted (just verified via a manual test that it's still broken).

Test manually by running the GUI via Gradle (I just click "Tasks => application => run" on the right panel in IntelliJ). Then load one of the test config files (e.g. `...\rcv\src\test\resources\network\brightspots\rcv\test_data\2017_minneapolis_mayor\2017_minneapolis_mayor_config.json`), save a copy via the Tabulator to make sure that functionality works. Re-load it. Then tabulate with it and verifying the summary JSON looks like the expected one in the same file path.

rcv.iml
HEdingfield: I see. The problem is it's referencing your local environment (see "tamird" appearing quite a bit in there). I'd say either figure out a way to make this more general / relative paths, or revert these changes.

src/main/java/network/brightspots/rcv/Main.java
HEdingfield: Ugh, sorry, just realized this was `Main`. Your original way was better.

src/main/java/network/brightspots/rcv/Tabulator.java
HEdingfield: Got it, sounds good then. Funny it doesn't pick up on uncapitalized letters after periods, haha.

build.gradle
HEdingfield: Oh nice, so it's run automatically via CI? Where should we keep an eye on for the output if these find anything?

I ran `check` locally using Gradle and it seems to run through all these. However, for `checkstyleMain` I'm now getting:
`warning: [options] module name in --add-exports option not found: org.junit.platform.engine`

And for `spotbugsTest` and `spotbugsMain` I'm getting:
```
WARNING: A terminally deprecated method in java.lang.System has been called
WARNING: System::setSecurityManager has been called by edu.umd.cs.findbugs.ba.jsr305.TypeQualifierValue (file:/C:/Users/...n/.gradle/caches/modules-2/files-2.1/com.github.spotbugs/spotbugs/4.6.0/9b88f95eba7b8de50339cd844e251bb7c6eca68a/spotbugs-4.6.0.jar)
WARNING: Please consider reporting this to the maintainers of edu.umd.cs.findbugs.ba.jsr305.TypeQualifierValue
WARNING: System::setSecurityManager will be removed in a future release
```

Should I file a new bug for these, or easy to resolve via this PR?

src/main/java/network/brightspots/rcv/GuiApplication.java
HEdingfield: Thanks for removing this suppression. I think you may have gone overboard removing the other ones though. They're in place to keep IntelliJ from complaining in its analysis.

HEdingfield:
@tamird not exactly sure how you're force pushing all these changes, but would you mind submitting a commit for each round of changes instead? We always squash before the merge anyway, and it would make things easier to review.

tamird:
May I suggest an alternative review tool which deals with multi-commit PRs more sanely? https://reviewable.io/ does a great job of this without forcing a workflow that produces a messy git history and without forcing single-commit PRs. This PR intentionally contains multiple commits, each of which is semantically separate, so I would prefer this not to be squashed on merge.

This PR can be reviewed [here](https://reviewable.io/reviews/BrightSpots/rcv/587#-).

build.gradle
tamird: That's [this bug](https://github.com/spotbugs/spotbugs/issues/1579).

src/main/java/network/brightspots/rcv/Main.java
tamird: Done.

src/main/java/network/brightspots/rcv/RawContestConfig.java
tamird: Reverted the rest and tested manuall.y

HEdingfield:
> May I suggest an alternative review tool which deals with multi-commit PRs more sanely? https://reviewable.io/ does a great job of this without forcing a workflow that produces a messy git history and without forcing single-commit PRs.

Totally open to new workflows! Anything to make things more like Critique would be great IMO. Maybe you can give us a quick walkthrough of reviewable.io sometime? Though, this may sound dumb, but the lack of dark mode makes it really hard for me to use as-is. Still interested in learning more about it though.

HEdingfield (approved):
Great work; thank you!

HEdingfield:
@tamird I didn't do the squash and merge per your request; I did a merge commit. If you prefer it done with a rebase merge, let me know and we can revert. Not used to doing it this way, so not sure what the best practice is.

tamird:
Merge commit is the right thing to do! Thanks.

> Maybe you can give us a quick walkthrough of reviewable.io sometime?

Happy to!

> Though, this may sound dumb, but the lack of dark mode makes it really hard for me to use as-is.

Hadn't considered this, looks like it's a [known issue](https://github.com/Reviewable/Reviewable/issues/903).

moldover:
@tamird just wanted to say thanks for this work and the suggestions.  I'm curious about reviewable.io as well.  Looks like they offer a free version for open repos which is nifty.

Issue #567, opened 2021-05-01
Multiple return statements


moldover:
VVSG coding standards forbid multiple return statements.  Remove from:

ContestConfig.java:
function ContestConfig.loadContestConfig
line 123

Tabulator.java: 
function Tabulator.shouldContinueTabulating
lines 392, 396

Logger.java
line 141
Logger.addGuiLogging

rrojas350:
According the SLI, who reported the previous issue, [#581](https://github.com/BrightSpots/rcv/issues/581), this is considered a "Moderate Risk." SLI is one of two EAC accredited Voting System Test Labs (VSTL), and responsible for certification of RCTab, so I would consider it a hard requirement as they may not let it slide in future certifications. 

On the flip side, I would hate to fix something only to have it cause more issues.

tarheel:
Yeah, it's a counterproductive/silly requirement, but it seems to be non-negotiable, so we just have to hold our nose and do it.

andyanderson:
Too bad Java doesn’t have a goto statement, it would make it so much easier 😂

tamird:
Also too bad that Java is not expression-oriented; this would be a natural pattern in e.g. Kotlin.

tarheel:
> Too bad Java doesn’t have a goto statement, it would make it so much easier 😂

I should have listened to my original instinct to write this application in QuickBASIC.

moldover:
Closed via PR #600

PR #600, opened 2022-06-06
fix for #567 multiple return statements

moldover:
Fix for #567 multiple return statements.

### DIFF
--- a/src/main/java/network/brightspots/rcv/Tabulator.java
+++ b/src/main/java/network/brightspots/rcv/Tabulator.java
@@ -394,27 +394,29 @@ class Tabulator {
   // determine if we should continue tabulating based on how many winners have been
   // selected and if continueUntilTwoCandidatesRemain is true.
   private boolean shouldContinueTabulating() {
+    boolean continueTabulating;
     int numEliminatedCandidates = candidateToRoundEliminated.keySet().size();
     int numWinnersDeclared = winnerToRound.size();
     // apply config setting if specified
     if (config.isContinueUntilTwoCandidatesRemainEnabled()) {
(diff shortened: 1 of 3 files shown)


src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -394,27 +394,29 @@ private void setWinningThreshold(Map<String, BigDecimal> currentRoundCandidateTo
...
       // in this mode, we're done as soon as we've declared any winners
-      return numWinnersDeclared == 0;
+      keepTabulating = numWinnersDeclared == 0;
     } else {
       // If there are more seats to fill, we should keep going, of course.
       // But also: if we've selected all the winners in a multi-seat contest, we should tabulate one
       // extra round in order to show the effect of redistributing the final surpluses... unless
       // bottoms-up is enabled, in which case we can stop as soon as we've declared the winners.
-      return numWinnersDeclared < config.getNumberOfWinners()
+      keepTabulating = numWinnersDeclared < config.getNumberOfWinners()
```
HEdingfield, line 414: It seems like you could just initially set it to this and get rid of the else statement, right? I guess it's more readable this way?

HEdingfield (approved):

src/main/java/network/brightspots/rcv/Tabulator.java
tarheel: I think it's more readable this way.

tarheel (approved):
Thanks! Did you confirm that these are the only offending methods via a manual audit, or did you use an automated check?

moldover:
> Thanks! Did you confirm that these are the only offending methods via a manual audit, or did you use an automated check?

Manual audit.  It's not bad.

Issue #561, opened 2021-04-13
Using % in config crashes (it is technically a valid windows path name)


moldover:

moldover:
When config contents are initially logged passing a path with % causes UnknownFormatConversionException

andyanderson:
Issue [#564](https://github.com/BrightSpots/rcv/issues/564#issuecomment-822174275) provides more detail about the source of this bug.

moldover:
Fixed with PR #601

Issue #564, opened 2021-04-18
Input CVR file name containing % fails tabulation: java.util.MissingFormatArgumentException: Format specifier '%2C'


andyanderson:
I’m working with output files from Qualtrics, which URL-encodes contest names (I assume to ensure cross-platform file name compatibility). An example is:

What+Are+Your+Favorite+Fruits+and+Vegetables%3F_April+13%2C+2021_23.28_Fruits.csv

The %3F is '?’ and %2C is ','. 

I’ve converted this to ES&S input format and resaved as Excel. When I validate, all is well. But when I tabulate, I get the following error:

java.util.UnknownFormatConversionException: Conversion = 'F'
Tabulation failed!

If I shorten the file name to just “April+13%2C+2021_23.28_Fruits.csv”, I get this error:

java.util.MissingFormatArgumentException: Format specifier '%2C'
Tabulation failed!

If I shorten the file name to just “Fruits.csv”, tabulation succeeds.

Apparently the file is being read by a different routine for tabulation than for validation, and the former isn’t taking care to protect the file name from attempted string formatting.

Attaching sample input files.

[What+Are+Your+Favorite+Fruits+and+Vegetables%3F_April+13%2C+2021_23.28_Fruits.zip](https://github.com/BrightSpots/rcv/files/6331672/What%2BAre%2BYour%2BFavorite%2BFruits%2Band%2BVegetables.3F_April%2B13.2C%2B2021_23.28_Fruits.zip)

tarheel:
Thanks for flagging. We actually got the same report just recently from another source. Closing this because it's a duplicate of #561.

moldover:
@HEdingfield + @andyanderson yep - was able to tabulate:

2022-06-06 18:53:47 PDT INFO: Round: 4
2022-06-06 18:53:47 PDT INFO: Winning threshold set to 14.
2022-06-06 18:53:47 PDT INFO: Candidate "Strawberry" got 15 vote(s).
2022-06-06 18:53:47 PDT INFO: Candidate "Grapefruit" got 11 vote(s).
2022-06-06 18:53:47 PDT INFO: Candidate "Strawberry" was elected in round 4 with 15 votes.

Congratulations to Strawberry.

PR #601, opened 2022-06-06
Allow percent % in path and file names #561

moldover:


src/main/java/network/brightspots/rcv/Logger.java
```diff
@@ -116,19 +118,24 @@ static void removeTabulationFileLogging() {
...
-    logger.log(Level.WARNING, String.format(message, obj));
+    log(Level.WARNING, message, obj);
   }
 
   static void severe(String message, Object... obj) {
-    logger.log(Level.SEVERE, String.format(message, obj));
+    log(Level.SEVERE, message, obj);
+  }
+
+  static void log(Level level, String message, Object... obj) {
```
HEdingfield, line 136: Recommend keeping this `private` until it's actually needed outside this class.

src/main/java/network/brightspots/rcv/Logger.java
```diff
@@ -99,9 +99,11 @@ static void setup() {
 
   // adds file logging for a tabulation run
   static void addTabulationFileLogging(String outputPath) throws IOException {
-    // use Level.FINE to capture audit info
-    tabulationHandler =
-        new FileHandler(outputPath, LOG_FILE_MAX_SIZE_BYTES, TABULATION_LOG_FILE_COUNT, true);
+    // replace any instances of % with %% because the FileHandler pattern generator requires this
+    tabulationHandler = new FileHandler(outputPath.replace("%", "%%"),
```
HEdingfield, line 103: Would this work as intended if a file name has a double percentage in the name?

HEdingfield (commented):
Were you able to test this with the files included in [this issue](https://github.com/BrightSpots/rcv/issues/564#issue-860701500)?

src/main/java/network/brightspots/rcv/Logger.java
moldover: Yep, I tried this with multiple percent characters and that part works, but as I'm checking again I realize I broke the %g functionality here so it's not quite correct yet.  Will push a fix for that.

moldover: Just pushed a fix for this bit.  Go ahead and get hog-wild with your path names.

moldover:
@HEdingfield + @andyanderson yep - was able to tabulate:

2022-06-06 18:53:47 PDT INFO: Round: 4
2022-06-06 18:53:47 PDT INFO: Winning threshold set to 14.
2022-06-06 18:53:47 PDT INFO: Candidate "Strawberry" got 15 vote(s).
2022-06-06 18:53:47 PDT INFO: Candidate "Grapefruit" got 11 vote(s).
2022-06-06 18:53:47 PDT INFO: Candidate "Strawberry" was elected in round 4 with 15 votes.

Congratulations to Strawberry.

src/main/java/network/brightspots/rcv/Logger.java
```diff
@@ -98,14 +98,24 @@ static void setup() {
...
+
+    tabulationHandler = new FileHandler(tabulationLogPattern,
+            LOG_FILE_MAX_SIZE_BYTES,
+            TABULATION_LOG_FILE_COUNT,
+            true);
     tabulationHandler.setFormatter(formatter);
     tabulationHandler.setLevel(Level.FINE);
     logger.addHandler(tabulationHandler);
-    info("Tabulation logging to: %s", outputPath.replace("%g", "*"));
+    info("Tabulation logging to: %s", tabulationLogPattern.replace("%g", "0"));
```
HEdingfield, line 118: Why this change? It seems like it makes it look like it's always pointing to log index 0 when that may not actually be the case.

HEdingfield (commented):
Big up the strawberry.

src/main/java/network/brightspots/rcv/Logger.java
moldover: I believe this is an improvement over the current message.  At the time it's printed it's accurate and it is less confusing than the current asterisk message, which might mean something to folks acquainted with the notion of wildcards, but not to our users.

HEdingfield: I guess this begs the question of why we even include the log index at all if we know it's always going to be 0? I assume it serves some purpose and, if so, I think we should keep this as-is. We've already got an established precedent of using * in a log name for the execution log message, and I think it's a pretty well-understood concept.

In any case, this PR looks good, thanks for fixing this!

HEdingfield (approved):

Issue #584, opened 2022-03-24
Rebrand Universal RCV Tabulator to RCTab


rrojas350:
To keep documentation and the software consistent, mentions of "Universal RCV Tabulator" will have to be updated to "RCTab."

HEdingfield:
![s-l500](https://user-images.githubusercontent.com/8259446/167310459-a3c051d8-082d-4249-a9bf-b6b52b67b5f6.jpg)
![16markets-brf-tab-mediumSquareAt3X](https://user-images.githubusercontent.com/8259446/167310511-e861d355-708f-4763-851f-09c1f5c4d587.jpg)

rrojas350:
![giphy](https://user-images.githubusercontent.com/98929622/167455964-6364fe63-12b6-471b-9559-0bc49e6e38ce.gif)

moldover:
Closed by #603, but PibbXtra is open, in our heart forever.

PR #603, opened 2022-06-13
rebranding

moldover:
Fix for #584 rebranding.


src/main/java/network/brightspots/rcv/CastVoteRecord.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Universal RCV Tabulator
+ * RCTab
  * Copyright (c) 2017-2020 Bright Spots Developers.
```
HEdingfield, line 3: Can you please update the copyright date to 2022 for all these files while you're at it?

HEdingfield (commented):
I found two other locations that should probably be updated too:
* [BrightSpotsCopyrightProfile.xml](https://github.com/BrightSpots/rcv/blob/26ae62fdd9402d9fd0d3f435771fb90553843dd1/.idea/copyright/BrightSpotsCopyrightProfile.xml#L3)
* [Clear Ballot 2.1 RCV Format Specification.rtf](https://github.com/BrightSpots/rcv/blob/26ae62fdd9402d9fd0d3f435771fb90553843dd1/reference/clear_ballot/RCV/Clear%20Ballot%202.1%20RCV%20Format%20Specification.rtf#L74)

rrojas350 (approved):
Main.java has a few lines that refer to the "RCV tabulator" but they're comments, on line 29 and 38.

It's not a major change or show stopper but that's all I found besides the ones Hylton pointed out.

src/main/java/network/brightspots/rcv/CastVoteRecord.java
moldover: I think it's better to do this along with the license updates in a different PR.

moldover: I fixed the copyright template, and updated the comments in Main.java.  The .rtf here is a document we received from Clear Ballot, so not something we want to touch.

Issue #551, opened 2021-04-02
What is the rationale for making this project AGPL?


jlumbroso:
I noticed the initial license for this project was Apache, but early in the commit logs this was replaced by the AGPL (I think by @tarheel at fe95903).

I am an RCV activist as well, and I am sincerely confused by the approach.

Strong copyleft licenses are activism for open-source. Here it seems like this project is both advocating for RCV but also for strong open-source, and that seems like a mix of concerns.

If your goal is to ensure that RCV comes to the US, wouldn't you welcome private vendors (who cannot be cut out of the process since a large number of election-related contracts are private) having access to this software and being able to use it themselves — this would not be great for open-source, but it would be great for RCV, as it would reduce the friction necessary for private actors to move to RCV.

I can understand someone considering that Apache is too permissive, that's why my favorite license is LGPLv3 — both it and the Mozilla Public License v2 are good trade-offs between promoting open-source but not making it impossible for private companies to use the artefacts. It states that if a vendor uses your library, they can do so without making their work open-source; but if they make any changes to your library, they must make these publicly accessible. I believe this is a good compromise between encouraging adoption and protecting the rights of the codebase.

tarheel:
Hi @jlumbroso, thanks for the question! You make good points. Our original decision to adopt AGPL was in response to strong encouragement from open-source proponents, as you may have guessed. You're right that a license like LGPL would make it much easier for a vendor to incorporate our code into their products. In practice, though, we've found that it's unlikely that a vendor would want to engage with our work in this way; they're more likely to a) merely look to our code as inspiration for their own proprietary implementation, or b) keep their focus on the voting machines and cast vote record generation and let our software handle the RCV tabulation directly (as is now happening in multiple jurisdictions). If we did encounter a situation in which a vendor wanted to pull our work into their own codebase and we saw potentially significant benefit for the broader RCV movement from enabling that adoption, we would give serious consideration to modifying our license. But, as far as we're aware, this is only a hypothetical at the moment.

HEdingfield:
Since we're going to be modifying the license in the near future (CC: @chughes297, @moldover), let's also make sure to update all the copyright blocks to say 2022 (they currently say 2020) and fix [this typo](https://github.com/BrightSpots/rcv/pull/603/commits/ece5a579d9aafefb290e751a81b467e2079826be) where it says "Universal RCTab" instead of "RCTab".

moldover:
Good call 👍 
[...]

PR #604, opened 2022-07-11
License updates

moldover:
This updates our license to MPL 2.0  #551 

License: https://github.com/BrightSpots/rcv/blob/license_updates_551/LICENSE.html

Added Contributors Guide:  https://github.com/BrightSpots/rcv/blob/license_updates_551/contributor_guide.md


moldover:
@rrojas350 see my question here: #551 ... the copyright notice needs to be entirely replaced on every source file (and the template)

tarheel (approved):

moldover:
@artoonie I realized there is something we should consider - I updated the license file in the test_data repo, that might conflict with how you are using rcvis.  are you ok with updating the rcvis license if needed?  check out more background in issue #551

artoonie:
Thanks for the heads up!

RCVis doesn't use any of the test_data, actually - that was for a separate project. Some files in RCVis may have been generated by the tabulator, but not as part of test_data.

moldover:
Ok cool, in that case we can leave test_data with the new license and we're gg.

PR #605, opened 2022-08-07
Upgrade deps

HEdingfield:
Upgrades Gradle to 7.5.1, spotbugs-gradle-plugin to 5.0.9, spotbugs to 4.7.1, checkstyle to 10.3.2, jlink to 2.25.0, jackson to 2.13.3, jupiter to 5.9.0, and javafx to 18. Also fixes `unchecked` Java compiler warning and issue where copyright was being inserted in non-source files.

NOTE:
* I decided not to update to JDK 18 (yet) because JDK 17 is LTS
* I wanted to update `org.openjfx.javafxplugin` to 0.0.13, but was blocked by [this issue](https://github.com/openjfx/javafx-gradle-plugin/issues/117)
* Suppressions were initially ignored after updating Gradle to 7.5.1, so I had to manually apply the fix mentioned [here](https://github.com/gradle/gradle/issues/21409) in `google_checks.xml`


moldover:
Love it.  Would you mind throwing one more in?  Update javafx to version 18 in build.gradle:

// ### JavaFX plugin settings
javafx {
    version = "18"
    modules = ["javafx.base", "javafx.controls", "javafx.fxml", "javafx.graphics"]
}

This is actually a =problem for me because the GUI crashes on M1 :/  Fixed in version 18.  
[Discussion: ](https://youtrack.jetbrains.com/issue/JBR-4116)

HEdingfield:
Updated!

build.gradle
```diff
@@ -36,10 +37,13 @@ application {
...
 // ### Checkstyle plugin settings
 checkstyle {
-    toolVersion = '10.2'
+    // Latest version at https://github.com/checkstyle/checkstyle/tags
+    toolVersion = '10.3.2'
     // Keep the below file updated along with subsequent versions of Checkstyle (make sure to choose
     // the tag matching the toolVersion above):
-    // https://github.com/checkstyle/checkstyle/blob/checkstyle-10.2/src/main/resources/google_checks.xml
+    // https://github.com/checkstyle/checkstyle/blob/checkstyle-10.3.2/src/main/resources/google_checks.xml
+    // NOTE: It may also be necessary to change "${org.checkstyle.google.suppressionfilter.config}"
```
tamird, line 45: I don't understand this note.

EDIT: ah, I see that this changed in the file itself. Can you clarify this comment to:
- cite something
- indicate that the edit takes place in that file?

Also, was that file updated by just downloading a new one? It might be good to write down a `curl` command here; you could even write `curl ... | sed s|org.checkstyle...|\${config_log}/...|` to automate this away entirely.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,16 @@
+#
+# RCTab
```
tamird, line 2: isn't there a gradle command that generates these files? if yes, it may not make sense to put these license headers there.

tamird (approved):

gradle/wrapper/gradle-wrapper.properties
HEdingfield: Good catch, this definitely shouldn't be appearing in here. I'll fix the auto-copyright rules and clean these up.

build.gradle
HEdingfield: Added more context per your request. Yeah, I just copy and paste it from GH usually, and I like the idea of automating it with a script like that except the address would need to be changed each time to the match most recent version posted to the Maven repo, and `sed` is a no-go in my Windows environment. :/

HEdingfield: Ack, sorry, missed this PR with this change. Will include it in my next.

PR #607, opened 2022-08-08
ignore .idea/misc.xml

tarheel:
This file keeps getting updated in my local checkout as follows:
```
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_17" project-jdk-name="17" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_17" default="true" project-jdk-name="17" project-jdk-type="JavaSDK">
```
Presumably it's not something we want in the repo anyway?


HEdingfield:
I see arguments on both sides saying to ignore and not to ignore. [This one](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360005054360-Get-misc-xml-from-VCS-but-do-not-track-it-by-default) seems semi-convincing.

I think @tamird had flipped that `default="true"` bit off in one of his recent PRs, but throughout the project the three of us have been fine with having it set that way, so I guess I'm in favor of keeping it (with default true), but don't feel super strongly.

tamird (approved):
I don't have an opinion on this.

PR #608, opened 2022-08-08
Exit gracefully if all declared candidates fall beneath minimum vote threshold

### LINKED ISSUE #553: silent failure if no candidate meets minimum threshold
We should handle this situation gracefully.

tarheel:
Fixes #553.

I didn't add a test case because I don't believe we have any support (yet) for tests that confirm a failed tabulation.


src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -173,7 +172,7 @@ void tabulate() {
           try {
             newWinnerSet = runTabulationForConfig(config, castVoteRecords);
           } catch (TabulationCancelledException exception) {
-            Logger.severe("Tabulation was cancelled by the user!");
+            Logger.severe(exception.getMessage());
```
HEdingfield, line 175: Here and below: I think you might be better off just logging `exception`, since that should include the message and additional useful info as well.

HEdingfield (approved):
Naisu.

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -1189,5 +1196,21 @@ static class BatchElimination {
 
   static class TabulationCancelledException extends Exception {
 
+    final boolean cancelledByUser;
```
tamird, line 1199: this doesn't appear to be used

tamird (commented):
Seems we need a test.

src/main/java/network/brightspots/rcv/TabulatorSession.java
tarheel: I don't think we want other info; the purpose of this exception class is to communicate a specific message to the user in the output.

src/main/java/network/brightspots/rcv/Tabulator.java
tarheel: Am I misunderstanding your comment? It's used just below.

tarheel (commented):
> Seems we need a test.

Not sure whether you saw my comment in the PR description: "I didn't add a test case because I don't believe we have any support (yet) for tests that confirm a failed tabulation."

Ideally, yes, we would have support for that kind of test, but I don't think it's worth doing at the moment. (Or at least I can't prioritize it right now.)

src/main/java/network/brightspots/rcv/Tabulator.java
tamird: Ah, I see - you did it this way so that you didn't have to update existing usages. Can we please remove the default constructor so that all usages must be explicit about this?

tamird:
> 

Yep, I saw. I'm not sure how to review this in the absence of a test.

tarheel:
OK, I removed the default constructor.

Regarding review: I did test it manually and confirmed that it aborts with the correct message if this situation occurs. So I think that suffices for now.

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -206,6 +206,13 @@ Set<String> tabulate() throws TabulationCancelledException {
         // 2. If there's a minimum vote threshold, drop all candidates below that threshold.
         if (eliminated.isEmpty()) {
           eliminated = dropCandidatesBelowThreshold(currentRoundTallyToCandidates);
+          // One edge case: if everyone is below the threshold, we can't proceed. This would only
+          // happen in the first or (if we drop undeclared write-ins first) second round.
+          if (eliminated.size() == config.getNumDeclaredCandidates()) {
+            Logger.severe("Tabulation can't proceed because all declared candidates are below "
```
moldover, line 212: Maybe "Tabulation complete because..." is more descriptive.

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -206,6 +206,13 @@ Set<String> tabulate() throws TabulationCancelledException {
...
+                + "the minimum vote threshold.");
+            throw new TabulationCancelledException(false);
```
moldover, line 214: It's not really cancelled, right?  It's just ending sooner than we thought.

src/main/java/network/brightspots/rcv/Tabulator.java
tarheel: I don't think we should consider this to be a "normal" case. Instead of producing winners and generating output files, we're aborting.

src/main/java/network/brightspots/rcv/Tabulator.java
tarheel: Would you be more on board with "aborted" instead of "cancelled"? I think that captures both situations where this exception is used (giving up because no one meets the threshold or exiting because the user clicked a Cancel button). Just `TabulationException` is also OK, but I'd prefer to be more specific.

moldover: Sure, sounds good.

HEdingfield (approved):
> Not sure whether you saw my comment in the PR description: "I didn't add a test case because I don't believe we have any support (yet) for tests that confirm a failed tabulation."
>
> Ideally, yes, we would have support for that kind of test, but I don't think it's worth doing at the moment. (Or at least I can't prioritize it right now.)

Would you mind adding an issue for this, pointing to this PR?

tarheel:
> Would you mind adding an issue for this, pointing to this PR?

Upon further reflection... it might not be a big deal to add a test for this. I'll follow up in this PR.

tarheel:
Added a test here also. It was awkward to find a way to make the test work, but I think the solution I came up with isn't... terrible.

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -131,6 +131,10 @@ void convertToCdf() {
   }
 
   void tabulate() {
+    tabulate(null);
+  }
+
+  void tabulate(List<String> exceptionsEncountered) {
```
moldover, line 137: Why not just have this return the List?  That's a more typical api for error codes and then you don't need to do all the checks for if (exceptionsEncountered != null) here.

moldover: You can also get rid of the wrapper methods

tarheel: I'm OK with that approach also. I was thinking it would be weird, but upon further reflection, I guess it's fine.

PR #610, opened 2022-08-08
fix bug in logic for exhaustIfMultipleContinuing overvote rule

### LINKED ISSUE #554: "exhaust if multiple continuing" not working properly
From NY testing:

*Expected Behavior:*
When using a CSV that contains one or more CVRs with overvotes in the first rank; where 3 or more candidates are selected within a single selection. When only one candidate has been eliminated during the first round, this will leave the remaining two more remaining candidates within the overvote as active and eligible. The system will thus consider these CVR selections as exhausted immediately and are tabulated as inactive.

*Observed Behavior:*
Within an overvoted contest when one candidate was eliminated and the remaining two or more candidates are eligible within the contest still. The system would count the votes towards one of these candidates instead of immediately exhausting the CVR record.

Short term: document this bug.

Next release: fix it.

tarheel:
This is a fun one. It definitely calls for a new test. Just waiting for guidance on the best way to do that.

Fixes #554.


src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -796,9 +796,9 @@ private OvervoteDecision getOvervoteDecision(Set<String> candidateSet) {
           if (continuingCandidate != null) { // at least two continuing
             decision = OvervoteDecision.EXHAUST;
             break;
+          } else {
+            continuingCandidate = candidate;
```
moldover, line 800: :facepalm:

tarheel: Indeed.

tarheel:
OK, test added.

moldover (approved):
Same ask here: remove the 'test_set' prefix from test files and names

tarheel:
Will do. Please accept the PRs preemptively so I can merge them as soon as I make these trivial updates (which might be at a weird time late tonight).

PR #611, opened 2022-08-08
allow batch elimination and "continue until two candidates remain" to be enabled in multi-pass IRV mode

### LINKED ISSUE #556: batch elimination and "continue until two candidates remain" should be permitted under multi-pass IRV
From NY:

*Relevant Documentation:*
v1.2.0-440.1 - Universal RCV Tabulator System Tab Hints.pdf

*Description:*
The bMulti-Pass IRVb setting is available for selection within the bWinner Election Modeb setting. bUse Batch Eliminationb is an available option under the Winning Rules tab. This feature eliminates candidates using one-by-one or batch eliminations until only two candidates remain. Candidates with the most votes at the end of a round will be considered winners, then a new set of rounds is run with the winning candidates from the previous round being disregarded.

*Expected Behavior:*
Within the Winning Rules tab, when the "Multi-Pass IRV" is selected for the Winner Election Mode setting, the operator contains the ability to determine if candidates will be eliminated either one-by-one or in a batch.

*Observed Behavior:*
Within the Winning Rules tab, when the "Multi-Pass IRV" is selected for the Winner Election Mode setting, the options bUse Batch Elimination" and bContinue until Two Candidates Remainb are unselected and greyed out. These settings cannot be changed by the operator and only one-by-one elimination may occur.

Short term: document this limitation. (I wouldn't really call it a bug since it doesn't lead to incorrect outcomes.)

Next release: allow the config to enable these options.

tarheel:
To refresh everyone's memory: "multi-pass IRV" is an especially silly form of multi-winner RCV in which you run a series of single-winner elections, excluding the candidates who have already won from each subsequent pass. So this filed issue is correct that there's no reason you shouldn't be able to enable these options for this election mode.

I'll also add a test for this one when I figure out that process.

Fixes #556.


moldover (dismissed):
Long live multi-pass IRV.

tarheel:
test added

moldover:
This looks like a good test and code change.  
Can I convince you to remove the 'test_set' prefix from your test/file names?
What is up with the force-pushes?

tarheel:
Yes, I can rename them.

I have to force push after I rebase an open PR on main. I know doing a merge commit is also an option, but I prefer this workflow, though I admit it's not perfect.

moldover (approved):

moldover:
> I have to force push after I rebase an open PR on main. I know doing a merge commit is also an option, but I prefer this workflow, though I admit it's not perfect.

Don't let perfect be the enemy of the good.  They should be friends.

Issue #612, opened 2022-08-10
Bring test data back into main RCTab repo


HEdingfield:
Per discussion in #576, we should consider moving the data from https://github.com/BrightSpots/rcvtests back into the main repo.

tarheel:
@moldover @HEdingfield OK, how about we go ahead and do this now? I wouldn't mind having a simpler process for finishing my open PRs...

moldover:
@tarheel you sure you dont want to learn more about submodules first?

HEdingfield:
The sass is off the charts. @moldover do you have any availability to do this? I'm working on a few other issues at the moment and I think @tarheel is too.

moldover:
Twould be my pleasure.

HEdingfield:
Before closing this issue, could you please let us know if there are any Git commands we need to run locally to undo [these steps]()?

Also, let's delete the repos [here](https://github.com/BrightSpots/rcvtests) and [here](https://github.com/artoonie/rcvtests) (CC @artoonie) as soon as we verify that the tests are running successfully locally.

artoonie:
I've archived the repo, effectively marking it as stale/out-of-date. LMK if you prefer a full deletion.

HEdingfield:
@artoonie Ok, sounds good. I think it's safe to delete on your end; I'll archive our fork of it and plan to delete it in a month or so if everything is still going smoothly.

@tarheel noted, after the above PR:

>  I reverted any changes I had made inside the test dir, then checked out develop, ran `git submodule deinit --all`, then `git pull`, and it appears to have worked

HEdingfield:
Just deleted BrightSpots/rcvtests.

tarheel:
Good night, sweet prince,
And flights of angels sing thee to thy rest.

PR #615, opened 2022-08-12
Remove submodule per #612

moldover:
Removes test_data submodule and re-add test_data files directly into the main repo as a subtree which preserves history.
#612 

@artoonie tagging to keep you in the loop here.


HEdingfield (changes requested):
Hmm, not sure how you went about doing this, but there are TONS of commits listed in the PR, which is kind of concerning. Any idea why?

HEdingfield:
I'm guessing related to this?
> re-add test_data files directly into the main repo as a subtree which preserves history

Is there a reason that we can't just re-add the files and make the changes as a new commit? I'd prefer not to erase the history of having done this.

moldover:
> Hmm, not sure how you went about doing this, but there are TONS of commits listed in the PR, which is kind of concerning. Any idea why?

That's the history being re-created.  When the submodule was first created those files became unparented from the original tree and history.  This is reapplying that history onto the new base.  I don't understand all this in detail, but the history is all there (although it's a bit wonky now) but that's what matters.  I pulled this from:

https://stackoverflow.com/questions/52224200/convert-a-git-submodule-to-a-regular-directory-and-preserve-the-history-in-the-m

HEdingfield:
Thanks for that context, I think I understand a bit. We basically want to make sure the history of the [3-5 commits](https://github.com/BrightSpots/rcvtests/commits/main) made in `rcvtests` since its inception are re-integrated into `rcv`'s history?

I think that makes sense, but I'd feel most comfortable if @artoonie gives this his close review and approval.

moldover:
Sure.  I just invited him to the group.

artoonie (approved):
nice git-fu gettting the history back! Looks like `git blame` does what we want it to do. 👍

moldover:
Sweet.  Just waiting on Hylton then.

moldover:
Sweet.  Just waiting on @HEdingfield then.

.idea/vcs.xml
```diff
@@ -2,6 +2,5 @@
 <project version="4">
   <component name="VcsDirectoryMappings">
     <mapping directory="" vcs="Git" />
-    <mapping directory="$PROJECT_DIR$/src/test/resources/network/brightspots/rcv/test_data" vcs="Git" />
```
HEdingfield, line 5: Do you know why this change is necessary? Not a blocker, but I'm curious.

.gitmodules
```diff
@@ -1,3 +0,0 @@
-[submodule "src/test/resources/network/brightspots/rcv/test_data"]
-	path = src/test/resources/network/brightspots/rcv/test_data
-	url = https://github.com/brightspots/rcvtests.git
```
HEdingfield, line 3: There's also mention of "submodules" in the contributor guide [here](https://github.com/BrightSpots/rcv/blob/88afa608a615b4229e6e758b0672f673b3b98697/contributor_guide.md#L22) and in the `test.yaml` file [here](https://github.com/BrightSpots/rcv/blob/26ae62fdd9402d9fd0d3f435771fb90553843dd1/.github/workflows/test.yml#L12) that might need to be addressed.

HEdingfield (approved):

.idea/vcs.xml
moldover: This tells intelliJ that the submodule is a submodule which enables Intellij vcs integration.

.gitmodules
moldover: I remove the reference from the guide.  

The yaml I imagine is just a generic configuration to configure submodules recursively if there are any.  It doesn't seem to have impacted the test from running and completing which is the important thing here.  @artoonie any thoughts on this?

.idea/vcs.xml
moldover: it's pretty solid btw - worth checking out even if you have a git client you like. @

PR #616, opened 2022-08-12
Changes audit logs to include validation outcome

### LINKED ISSUE #566: Errors during config validation are logged to the execution logging file, instead of the tabulation audit log.
**Description:**

The electronic records shall be capable of being exported in a non-proprietary format. This
export may be used for auditing or analysis purposes, in order to determine if all information
within is correct or in order to help troubleshoot any errors that may occur. The error messages
shall be displayed within this audit log to allow a trained technician to diagnose and resolve the
error.

**Expected Behavior:**

All electronic logs shall be capable of being exported in order to audit or analysis the
information within the technology computing platform. This audit log shall contain all error
messages and events that have occurred in order to allow a trained technician to identify and
resolve error messages that have occurred on the system.

**Observed Behavior:**

The Audit log is only capable of being exported when the tabulation process has been
completed successfully. There are no addition settings or options to export electronic logs. Thus
the error logs will not be included within the audit log unless tabulation has been completed.

Jon notes: 

audit logging does not get hooked up until after config validation completes. So an error during validation, such as a bad file path name, would result in no audit log which seems like it should be improved.

As folks point out, the "execution logging" does capture validation errors, but it is in a separate log file which is not really the tabulation audit. For Mac this file is in the user's HOME directory. I'm not sure what it is on Windows, but when you launch via rcv.bat it should tell you in the console window: 'execution logging to ....'
So maybe we call this another known bug:
Errors during config validation are logged to the execution logging file, instead of the tabulation audit log.

HEdingfield:
Also improves comment in `build.gradle`

Fixes #566.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,8 +42,13 @@ checkstyle {
     // Keep the below file updated along with subsequent versions of Checkstyle (make sure to choose
     // the tag matching the toolVersion above):
     // https://github.com/checkstyle/checkstyle/blob/checkstyle-10.3.2/src/main/resources/google_checks.xml
-    // NOTE: It may also be necessary to change "${org.checkstyle.google.suppressionfilter.config}"
-    // to "${config_loc}/checkstyle-suppressions.xml" or else the suppressions might be ignored
+    // NOTE: After updating the local version of config/checkstyle/google_checks.xml again in the
+    // future, it may be necessary to subsequently replace the following line in the local copy to
+    // prevent suppressions from being ignored:
+    // "${org.checkstyle.google.suppressionfilter.config}"
+    // with:
+    // "${config_loc}/checkstyle-suppressions.xml"
+    // Additional context: https://github.com/gradle/gradle/issues/21409
     configFile = file("$projectDir/config/checkstyle/google_checks.xml")
     maxWarnings = 0
     ignoreFailures = false
--- a/src/main/java/network/brightspots/rcv/TabulatorSession.java
+++ b/src/main/java/network/brightspots/rcv/TabulatorSession.java
@@ -26,7 +26,6 @@ import java.io.BufferedReader;
 import java.io.FileReader;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Paths;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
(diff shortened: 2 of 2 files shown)


src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -136,7 +135,8 @@ void tabulate() {
     ContestConfig config = ContestConfig.loadContestConfig(configPath);
     checkConfigVersionMatchesApp(config);
     boolean tabulationSuccess = false;
-    if (config.validate() && setUpLogging(config)) {
+    boolean isLoggingSetUp = setUpLogging(config);
+    if (config.validate() && isLoggingSetUp) {
```
tamird, line 139: this is change in behavior. `setUpLogging` is now called with `confi` even if `config.validate()` is false. That doesn't seem right.

build.gradle
```diff
@@ -42,8 +42,13 @@ checkstyle {
     // Keep the below file updated along with subsequent versions of Checkstyle (make sure to choose
     // the tag matching the toolVersion above):
     // https://github.com/checkstyle/checkstyle/blob/checkstyle-10.3.2/src/main/resources/google_checks.xml
-    // NOTE: It may also be necessary to change "${org.checkstyle.google.suppressionfilter.config}"
-    // to "${config_loc}/checkstyle-suppressions.xml" or else the suppressions might be ignored
+    // NOTE: After updating the local version of config/checkstyle/google_checks.xml again in the
```
tamird, line 45: Let's please not lump things together like this. At the very least this should be in its own commit.

src/main/java/network/brightspots/rcv/TabulatorSession.java
HEdingfield: Actually this is exactly the goal of the PR: we want the audit logs to now include failed validation messages.

build.gradle
HEdingfield: Noted for the future. We generally squash all our commits within a PR though, so it won't matter in practice.

src/main/java/network/brightspots/rcv/TabulatorSession.java
tamird: This API is pretty confusing. It reads to me like `config` may not be valid, yet we pass it as input to a function (which presumably expects it to be valid?). Does `setUpLogging` need the entire config object? Can we clarify that this is an acceptable usage?

Also, do we need a test?

HEdingfield: Good call, I made an attempt to clarify this more.

Re: the test, I don't think it's necessary -- all our tests so far revolve around the actual tabulation process, whereas this is just a slight change to what's included in the logs. I verified locally that the original issue identified in #566 is solved by this change.

tamird (approved):

tarheel:
@HEdingfield sorry, I hadn't paid close attention to this PR previously. I just noticed that it violates our favorite arbitrary requirement: you added multiple return statements to `TabulatorSession.tabulate`. To avoid having to add an extra level of indentation to like 100 lines, you might want to structure it something like this:
```
if (!isLoggingSetUp) {
  ...
} else if (!config.validate()) {
  ...
} else {
  callHelperFunction();
}
```

tarheel:
Actually, just pushed a commit to #608 that addresses this.

HEdingfield:
Arghhh, painfully stupid! Thanks for "fixing" it!

Issue #592, opened 2022-05-09
Static CVR Provider and File Path


rrojas350:
Related to #589 

Having CVR Provider and File Path saved and carried over by a JSON config file. This will help jurisdictions create a complete JSON config file for easy election night tabulation and the addition of new CVR files. Could be solved by #589.

HEdingfield:
To clarify this: the ask is that, when a user loads an existing config file and then they attempt to add a path to a new CVR file, it will attempt to open the dialog box at the location of the last CVR file that's already in the list? And also pre-populate provider with the same one as the last one in the list?

rrojas350:
Correct Hylton.

HEdingfield:
Given the resolution of #589, I don't think this should necessary any longer. Feel free to re-open if anyone disagrees.

PR #617, opened 2022-08-12
Allows users to specify multiple CVR files at once in the GUI

### LINKED ISSUE #589: Uploading Multiple CVRs
The current process can get tedious when uploading multiple CVRs from the same vendor. Rather than selecting individual CVRs, users would like to option/ability to select a folder containing multiple CVRs.

HEdingfield:
Fixes #589 and #592.

### DIFF
--- a/src/main/java/network/brightspots/rcv/GuiConfigController.java
+++ b/src/main/java/network/brightspots/rcv/GuiConfigController.java
@@ -98,7 +98,8 @@ public class GuiConfigController implements Initializable {
   private static final String HINTS_VOTER_ERROR_RULES_FILENAME =
       "network/brightspots/rcv/hints_voter_error_rules.txt";
   private static final String HINTS_OUTPUT_FILENAME = "network/brightspots/rcv/hints_output.txt";
-  private static final String CVR_FILE_PATH_DELIMITER = ";";
+  // It's possible for file paths to legitimately have consecutive semicolons, but unlikely
+  private static final String CVR_FILE_PATH_DELIMITER = ";;";
 
   // Used to check if changes have been made to a new config
   private String emptyConfigString;
--- a/src/main/java/network/brightspots/rcv/GuiConfigController.java
+++ b/src/main/java/network/brightspots/rcv/GuiConfigController.java
@@ -522,6 +522,7 @@ public class GuiConfigController implements Initializable {
    */
   public void buttonCvrFilePathClicked() {
     List<File> selectedFiles = null;
+    File selectedDirectory = null;
 
     Provider provider = getProviderChoice(choiceCvrProvider);
     switch (provider) {
@@ -543,8 +544,7 @@ public class GuiConfigController implements Initializable {
         DirectoryChooser dc = new DirectoryChooser();
         dc.setInitialDirectory(new File(FileUtils.getUserDirectory()));
         dc.setTitle("Select " + provider + " Cast Vote Record Folder");
-        // TODO: Uncomment / figure out what to do about directories
-        // selectedFiles = dc.showDialog(GuiContext.getInstance().getMainWindow());
+        selectedDirectory = dc.showDialog(GuiContext.getInstance().getMainWindow());
       }
       case ESS -> {
         FileChooser fc = new FileChooser();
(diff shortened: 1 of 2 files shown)


src/main/resources/network/brightspots/rcv/hints_cvr_files.txt
```diff
@@ -7,8 +7,12 @@ For each of your CVR files, provide the necessary information and then use the A
 Provider (required): The vendor/machine that generated (produced) the file. After you select the field, the tabulator will fill in as many of the other fields as it can based on what it knows about that provider. You can adjust those values as necessary.
 
 Path (required): Location of the CVR file.
-  * Example: /Users/test data/2015-portland-mayor-cvr.xlsx
-  * Example: C:\Users\test data\precinct14.json
+  * Example: /test data/2015-portland-mayor-cvr.xlsx
+  * Example: C:\test data\2015-portland-mayor-cvr.xlsx
+NOTE: You can select multiple files by holding down Shift in the dialog. You can also manually add multiple if you enter the values in the Path input box, separated with semicolons:
```
tarheel, line 12: This guidance is Windows-specific. On macOS, you can select a set of consecutive files by Shift-clicking, or you can select two or more non-adjacent files by Command-clicking.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -518,71 +521,95 @@ public void buttonClearDatePickerContestDateClicked() {
...
+              treatBlankAsUndeclaredWriteIn
+          );
+      if (ContestConfig.passesBasicCvrSourceValidation(cvrSource)) {
+        tableViewCvrFiles.getItems().add(cvrSource);
+      } else {
+        failedFilePaths.add(filePath);
+      }
+    });
+    // If any entries failed validation, preserve them in the text box so the user can try again
+    textFieldCvrFilePath.setText(String.join(CVR_FILE_PATH_DELIMITER, failedFilePaths));
```
tarheel, line 612: Is this standard behavior elsewhere, or just something you thought of on your own? I can see how it might be useful, but it may be confusing if we don't tell the user explicitly what just happened (x files added successfully; y invalid files remain).

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -96,6 +98,7 @@ public class GuiConfigController implements Initializable {
   private static final String HINTS_VOTER_ERROR_RULES_FILENAME =
       "network/brightspots/rcv/hints_voter_error_rules.txt";
   private static final String HINTS_OUTPUT_FILENAME = "network/brightspots/rcv/hints_output.txt";
+  private static final String CVR_FILE_PATH_DELIMITER = ";";
```
tarheel, line 101: On macOS/Linux, it's possible to have a semicolon in a filename. :/

tarheel (commented):
Cool! Just some minor things...

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Yes, it's standard to leave the values they previously entered in place if validation failed (I believe this was specifically request a while back).

In practice, we should never have a situation where something is invalid for only one of the items in the list (since they're all going to have the same properties; the one exception is if you do something like `;;;;;;` between other valid entries), but even if that were the case, the logging console should make it clear below what happened.

Essentially what this ends up doing is preserving everything in the Path input if validation fails for one or more of the properties (like a missing first vote column value), which I think is a good and intuitive thing.

src/main/resources/network/brightspots/rcv/hints_cvr_files.txt
HEdingfield: Updated this, thanks!

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Same in Windows, actually, I just chose something that I thought was unlikely. It seems the only illegal character for *nix is /, which is obviously not viable to use as a delimiter. I changed to to `;;` to make it less likely to interfere with a legit file path.

src/main/java/network/brightspots/rcv/GuiConfigController.java
tarheel: Yeah, I just think it might be more intuitive to make the code explicitly all or none. If somehow we successfully added some but not all, that would be kind of weird.

tarheel (approved):

PR #636, opened 2023-03-14
UI and config updates for Stop Tabulation Early

artoonie:
No logic changes to tabulation yet, just UI + configuration changes.
image

Being my first Pull Request with a substantial feature, I'm asking for a more detailed review than I'd likely need in the future. I've called out specific decisions throughout the code that I'd like feedback on.

If all looks good, I believe this is ready to merge.


src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -27,7 +27,7 @@
 public class Main extends GuiApplication {
 
   public static final String APP_NAME = "RCTab";
-  public static final String APP_VERSION = "1.3.0";
+  public static final String APP_VERSION = "1.3.1";
```
tarheel, line 30: We're doing 1.4.0 for this release.

tarheel (commented):
I can look more closely at your code later, but some initial responses:

> Style. Is there an automatic linter or styleguide?

Yes, please use Checkstyle. (The config is under `config/checkstyle`. Not sure about other IDEs, but it integrates well with IntelliJ.)

> Config. I followed the paradigm of a specific keyword to specify "max" in the config, which translates to checkboxes in the UI. Is that reasonable?

I would prefer to match the UI we use for other optional fields: we just have a single text box, and the user leaves it blank (which corresponds to null internally) if they don't want to specify it.

> Bumping version to update config -- any unforeseen issues there?

I don't believe so...? Hylton might remember something.

Btw, unrelated to your PR, but while looking at your screenshot I noticed that we have an asterisk next to "Random seed." This is inaccurate: that field is only required if the config selects one of the random tie-break modes. You might as well fix it while you're here.

artoonie:
Thank you -- that all makes sense. Updated the main post with the new UI.

What do you use to edit the .fxml files? I'm using SceneBuilder but it changes the formatting and creates a too-large diff. For these commits I'm manually editing the files.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -70,6 +70,8 @@ class ContestConfig {
   private static final int MAX_ROW_INDEX = 100000;
   private static final int MIN_MAX_RANKINGS_ALLOWED = 1;
   private static final int MIN_MAX_SKIPPED_RANKS_ALLOWED = 0;
+  private static final int MIN_NUMBER_OF_ROUNDS = 0;
+  private static final int MAX_NUMBER_OF_ROUNDS = Integer.MAX_VALUE;
```
artoonie, line 74: Should MAX_NUMBER_OF_ROUNDS be here or on line 710 in the call to `fieldOutOfRangeOrNotInteger`?

src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
```diff
@@ -22,6 +22,8 @@ Use Batch Elimination: Batch elimination, or simultaneous elimination of all can
 
 Continue until Two Candidates Remain: Single-winner ranked-choice voting elections can stop as soon as a candidate receives a majority of votes, even though 3 or more candidates may still be in the race. Selecting this option will run the round-by-round count until only two candidates remain, regardless of when a candidate wins a majority of votes. Available only when Winner Election Mode is "Single-winner majority determines winner" or "Multi-pass IRV."
 
+Stop Tabulation Early: If a winner is not found by the given round, tabulation stops early.
```
artoonie, line 25: Does anybody else need to look over this to make sure it's sensible? Does this sync with any external documentation?

artoonie: Approved by Rene

src/test/resources/network/brightspots/rcv/test_data/stop_tabulation_early_test/stop_tabulation_early_test_config.json
```diff
@@ -0,0 +1,55 @@
+{
```
artoonie, line 1: To make this test, I copied and pasted another directory with slight modifications. Is that the recommended way to go about this? If I were being a little more strict, I'd test, e.g.:
1. Invalid inputs (e.g. "-1" in the config)
2. Config set to zero (stops tabulating successfully before any tabulation, i.e. just the config is spit out)
3. Config set to > than the number of rounds provided

But with a new directory and large files needed for each test, this feels like overkill. I believe I've followed the existing best-practices laid out in the repo, but let me know if anybody has other preferences here.

src/test/resources/network/brightspots/rcv/test_data/stop_tabulation_early_test/stop_tabulation_early_test_expected_summary.json
```diff
@@ -0,0 +1,33 @@
...
+    } ]
+  }, {
+    "round" : 2,
+    "tally" : {
+      "George Gervin" : "3",
+      "Mookie Blaylock" : "3"
+    },
+    "tallyResults" : [ {
+      "eliminated" : "George Gervin",
+      "transfers" : { }
```
artoonie, line 30: Reading the legislation, having no transfers on this round seems correct. Let me know if anybody disagrees. (It doesn't make sense to add in last-round transfers to me, but I did want to flag just in case.)

src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -395,6 +395,12 @@ void testSkipToNext() {
     runTabulationTest("skip_to_next_test");
   }
 
+  @Test
+  @DisplayName("test stopping tabulation after round 2")
```
HEdingfield, line 399: Nit: "test stopping tabulation early", no real value in hard-coding a number in the name at this point.

src/main/java/network/brightspots/rcv/ContestConfigMigration.java
```diff
@@ -188,6 +188,11 @@ static void migrateConfigVersion(ContestConfig config)
         }
       }
 
+      // New rule to stop tabulation early
```
HEdingfield, line 191: Nit: unnecessary comment.

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
```diff
@@ -430,6 +430,19 @@
...
                     </HBox>
+                    <HBox spacing="4.0">
+                       <children>
+                          <Label prefWidth="220.0" text="Stop Tabulation Early on Round">
+                             <HBox.margin>
+                                <Insets />
+                             </HBox.margin>
+                          </Label>
+                          <TextField fx:id="textFieldStopTabulationEarly" maxWidth="80.0" />
+                       </children>
```
HEdingfield, line 441: Appreciate the screenshot! Curious why you're adding this extra `children`, `insets`, etc. stuff though? I figured you could just copy what was done for Minimum Vote Threshold above and tweak it slightly.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -795,6 +797,7 @@ private void clearAndDisableWinningRuleFields() {
     checkBoxMaxRankingsAllowedMax.setDisable(true);
     textFieldMinimumVoteThreshold.clear();
     textFieldMinimumVoteThreshold.setDisable(true);
+    textFieldStopTabulationEarly.clear();
```
HEdingfield, line 800: Please follow the other examples and `setDisable` too. Then around line 1115, undo it, as is done with the others.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -240,6 +240,8 @@ public class GuiConfigController implements Initializable {
   @FXML
   private CheckBox checkBoxContinueUntilTwoCandidatesRemain;
   @FXML
+  private TextField textFieldStopTabulationEarly;
```
HEdingfield, line 243: Nit: might as well name it the full `textFieldStopTabulationEarlyOnRound`.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -789,6 +800,10 @@ private String getNumberOfWinnersRaw() {
     return rawConfig.rules.numberOfWinners;
   }
 
+  private String getStopEarlyOnRoundRaw() {
```
HEdingfield, line 803: Nit: `getStopTabulationEarlyOnRoundRaw`

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -947,6 +962,12 @@ boolean isContinueUntilTwoCandidatesRemainEnabled() {
     return rawConfig.rules.continueUntilTwoCandidatesRemain;
   }
 
+  Integer getStopTabulationAtRound() {
```
HEdingfield, line 965: Nit: `getStopTabulationEarlyOnRound`

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1125,6 +1146,7 @@ enum ValidationError {
     RULES_MIN_DECIMAL_PLACES_FOR_VOTE_ARITHMETIC_INVALID,
     RULES_MIN_VOTE_THRESHOLD_INVALID,
     RULES_MULTI_SEAT_BOTTOMS_UP_PERCENTAGE_THRESHOLD_INVALID,
+    RULES_STOP_EARLY_ON_ROUND_INVALID,
```
HEdingfield, line 1149: Nit: `RULES_STOP_TABULATION_EARLY_ON_ROUND_INVALID`

HEdingfield (changes requested):
Great first PR!

HEdingfield:
> Btw, unrelated to your PR, but while looking at your screenshot I noticed that we have an asterisk next to "Random seed." This is inaccurate: that field is only required if the config selects one of the random tie-break modes. You might as well fix it while you're here.

@tarheel I believe this is correct actually, and would argue we should leave it as-is. We don't dynamically generate these asterisks, so it looks like they're used whenever something is required in certain cases as well.

> What do you use to edit the .fxml files? I'm using SceneBuilder but it changes the formatting and creates a too-large diff. For these commits I'm manually editing the files.

Yeah, this sucks. I sometimes use SceneBuilder in IntelliJ to get a basic wireframe in place (or if I'm confused about how to add a particular element), but then I'll go back into the code, copy the main chunk it added, revert the file, then paste it in and manually tweak it until it looks good.

> PS I prefer force-pushes + clean commit history, but force pushing dirties Pull Requests. LMK if you favor additional commits over force-pushes, or if you prefer me to push less frequently.

We do squash commits on our PRs, so I think that means either way would be fine?

src/test/resources/network/brightspots/rcv/test_data/stop_tabulation_early_test/stop_tabulation_early_test_config.json
artoonie: The other tests won't flip out, but it will force you to re-save the config when loading the test, before tabulating. I didn't actually realize that's what was sometimes forcing me to re-save until I saw this comment. It is annoying, but not terrible? I'm happy to manually update all configs if needed, though perhaps we should wait til the end of 1.4 work so we only need to update each config once.

artoonie:
Thanks for the review! All changes made.

src/main/java/network/brightspots/rcv/ContestConfigMigration.java
artoonie: I updated the comment to clarify that this is for a 1.3.0 to 1.4.0 migration, since I expect there will be additional migrations and it would be nice to keep them organized in case we eventually create a more full-featured migration system.

artoonie:
> > Btw, unrelated to your PR, but while looking at your screenshot I noticed that we have an asterisk next to "Random seed." This is inaccurate: that field is only required if the config selects one of the random tie-break modes. You might as well fix it while you're here.
> 
> @tarheel I believe this is correct actually, and would argue we should leave it as-is. We don't dynamically generate these asterisks, so it looks like they're used whenever something is required in certain cases as well.

FYI this was already merged: https://github.com/BrightSpots/rcv/pull/637
LMK if you'd like me to revert that

src/test/resources/network/brightspots/rcv/test_data/stop_tabulation_early_test/stop_tabulation_early_test_config.json
HEdingfield: Ok, that sounds reasonable to me, filed #642 to remember this.

src/main/java/network/brightspots/rcv/ContestConfigMigration.java
HEdingfield: Cool, that seems more useful to me.

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: This still says "at" instead of "on", so recommend changing it to be consistent with everything else.

HEdingfield (approved):
Great work on this! Just one more nit I suggest fixing and then I think you're good to merge.

HEdingfield:
> > > Btw, unrelated to your PR, but while looking at your screenshot I noticed that we have an asterisk next to "Random seed." This is inaccurate: that field is only required if the config selects one of the random tie-break modes. You might as well fix it while you're here.
> > 
> > 
> > @tarheel I believe this is correct actually, and would argue we should leave it as-is. We don't dynamically generate these asterisks, so it looks like they're used whenever something is required in certain cases as well.
> 
> FYI this was already merged: #637 LMK if you'd like me to revert that

Yes, please revert, per https://github.com/BrightSpots/rcv/pull/637#issuecomment-1474973967.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -701,6 +702,15 @@ && getMaxRankingsAllowed() < MIN_MAX_RANKINGS_ALLOWED)) {
           ValidationError.RULES_MULTI_SEAT_BOTTOMS_UP_PERCENTAGE_THRESHOLD_INVALID);
     }
 
+    if (fieldOutOfRangeOrNotInteger(
+        getStopTabulationEarlyOnRoundRaw(),
+        "stopEarlyOnRound",
+        MIN_NUMBER_OF_ROUNDS,
```
tarheel, line 708: Shouldn't this be 1? If they say to stop after 0 rounds, that's the same as just not running the tabulation at all, isn't it?

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -412,7 +412,9 @@ private boolean shouldContinueTabulating() {
     int numEliminatedCandidates = candidateToRoundEliminated.size();
     int numWinnersDeclared = winnerToRound.size();
     // apply config setting if specified
```
tarheel, line 414: let's just delete this comment; it's ambiguous and doesn't add anything anyway

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -789,6 +799,10 @@ private String getNumberOfWinnersRaw() {
     return rawConfig.rules.numberOfWinners;
   }
 
+  private String getStopTabulationEarlyOnRoundRaw() {
```
tarheel, line 802: If I'm not mistaken, this setting means that we'd stop _after_ the specified number of rounds, right? If so, let's call it that instead to make it more clear. Stop _on_ a round makes it sound like maybe we're stopping at the beginning of that round.

src/test/resources/network/brightspots/rcv/test_data/stop_tabulation_early_test/stop_tabulation_early_test_expected_summary.json
tarheel: I'm inclined to agree, since transfers would only be relevant if you were going to then proceed and show the tallies for the next round. But @chughes297 and @rrojas350 may have an opinion.

tarheel (commented):
I added a few inline thoughts.

One bigger question (and apologies for not thinking of this earlier): does this actually satisfy the requirement from the legislation?

> "The voting system must allow the user to decide whether to pause the tabulation session after each round or to continue until a winner is determined or a manual tie break for elimination is required."

To me, that sounds like we actually want a boolean setting for this -- and when it's enabled, the tabulator stops after every round and lets the user decide whether to proceed.

cc @chughes297 and @rrojas350 for this question also.

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: Yes, I figured that might be useful in some cases -- you'd still get the output files with zero rounds (similar to running a Zero Report?). I can change it to `1` though if you prefer.

src/main/java/network/brightspots/rcv/Tabulator.java
artoonie: Can do. (Note: It wasn't part of this PR.)

artoonie:
https://github.com/BrightSpots/rcv/pull/648 addresses all the comments here

chughes297:
Sorry to have missed earlier conversations here, but I don't see a use case for telling tabulation to stop at a specific round of counting. Like Louis, I read the law as requiring software that can be told to stop after each round of counting, not to stop at an arbitrary round of counting. When I read the requirement I think of something like [what Alaska did for their Nov 2022 results livestream](https://youtu.be/egMMy1QGqRQ?t=186). That seems to require far more than just what is being considered in this issue though.

tarheel:
I think the other way would be preferable for sure, but if folks think that the implementation here is technically sufficient, maybe we make the revised approach a P1 or P2 for this release.

rrojas350:
I reached out to the CO SOS's office to get their opinion on this item but for now we can hold off on making any further changes. We may revisit this item in v1.5, depending on their feedback and requirements but as far as 1.4, we're good.

PR #637, opened 2023-03-14
Random Seed is optional; mark it as such in the UI

artoonie:
Prefer single-issue PRs just in case we close the other one and forget to re-integrate this minor change

### DIFF
--- a/src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
+++ b/src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
@@ -461,7 +461,7 @@
                       </padding>
                     </HBox>
                     <HBox alignment="CENTER_LEFT" spacing="4.0">
-                      <Label prefWidth="220.0" text="Random Seed *"/>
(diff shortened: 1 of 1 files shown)


tarheel (approved):

HEdingfield:
Per https://github.com/BrightSpots/rcv/pull/636#issuecomment-1474931006, this should be reverted. It automatically disables the text box when it's not required, and it enables the textbox whenever it is required (in which case the * is accurate because it's a required field).

tarheel:
How hard would it be to make the * dynamic also? I don't like that the label indicates that this field is required when we actually have (I believe) zero real users who need to use it. Not a big deal, but right now we're sort of optimizing for an edge case at the expense of the normal UX.

(I haven't taken the time to look, but I believe there are other fields that would benefit from similar treatment.)

HEdingfield:
I think it would be a pain with very little gain, TBH, but not impossible.

This field _is_ required though whenever they choose an option that enables it. It won't pass validation if they choose a random method but don't provide a seed.

tarheel:
Filed #647 for this as a nice-to-have.

PR #638, opened 2023-03-15
Add checks and nicer errors to DF CVR reader

artoonie:


src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -132,6 +132,25 @@ void parseXml(List<CastVoteRecord> castVoteRecords)
...
+        throw new CvrParseException();
+      }
+
+      // These fields are also required, but we are not aware of any standard format where
+      // they would be missing, so we group all these checks together.
+      if (cvrReport.CVR == null
+              || cvrReport.Election == null
+              || cvrReport.ReportGeneratingDeviceIds == null
+              || cvrReport.ReportingDevice == null
+              || cvrReport.Party == null) {
```
HEdingfield, line 149: No need to worry about the other fields in `CastVoteRecordReport`? E.g.:
* `GeneratedDate`
* `Notes`
* `ReportType`
* `OtherReportType`
* `Version`

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -132,6 +132,25 @@ void parseXml(List<CastVoteRecord> castVoteRecords)
       CastVoteRecordReport cvrReport = xmlMapper.readValue(inputStream, CastVoteRecordReport.class);
       inputStream.close();
 
+      // Some checks to provide nicer error messages.
+      // This is common with Unisyn's CDF CVR, which is not compatible with RCTab.
+      if (cvrReport.GpUnit == null) {
+        Logger.severe("Field \"GPUnit\" missing from CVR. "
```
HEdingfield, line 138: Nit: I think we generally put an exclamation point for `severe` log messages. E.g. here would be "... missing from CVR!". Below would be "... missing from CDF CVR file!".

Also recommend consistency between these messages, i.e. first would say "...missing from CDF CVR file!".

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
artoonie: I have only checked the arrays, which would lead to NPEs when somebody tries to iterate over them later. I assume the others could be optional?

If we wanted to ensure nothing is null, I could look into using `@JsonInclude(JsonInclude.Include.NON_NULL)` to disallow any null fields on the XML (though I'd need to be careful to ensure no unforeseen consequences there).

HEdingfield: You might be right; this is genuinely out of my area of expertise, so I'm fine leaving as-is. Might want to run this by @tarheel to see what he thinks before merging, but LGTM overall!

HEdingfield (approved):

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
artoonie: @tarheel can you confirm this looks good to you?

Issue #591, opened 2022-05-09
Alias for Candidates


rrojas350:
Ability to add an Alias for Candidates. Some CVRs may contain variations of a Candidate's name, as not all jurisdictions are given the freedom to adjudicate write-ins to match an existing candidate. Having aliases for candidates will help produce accurate results for these jurisdictions. 

Related to #60

rrojas350:
Adding to this. When testing multi-vendor tabulation I ran into an issue where a candidate's name had a variation in one CVR and not the rest. Ideally, everything should be consistent when dealing with a multi-jurisdiction race but this feature could be used to resolve any potential candidate name variations.

tarheel:
Makes sense. It might be a little tricky to figure out how to support this on the UI side, but the internal implementation shouldn't be too bad.

chughes297:
Could we resolve this with #60 ?

tarheel:
The original intention in #60 (gathering candidate names from CVR sources) would not be sufficient on its own. But we actually ended up discussing the alias idea in that issue also.

My preference would be to move forward just with alias support for now, since that's relatively straightforward and should provide a reasonable workaround to the problematic cases people have raised.

ggilber1:
I am certainly in favor of separating the issues #60 and #591.  The original intent of #60 got seriously sidetracked by the normalization of write-ins discussion.  While the added complexity Rene has raised here, of name variations between CVRs supplied by different voting systems, certainly comingles the issues, I feel auto-populating candidate names is a priority that needs to finally be incorporated into RCTab for all vendors.

artoonie:
Thoughts on just a big textbox asking for one name per line? I toyed with a couple options (e.g. "Add Alias" similar to the "Add Candidate") but that had a confusing UX.

Pros: easy, clear UI
Cons: more potential for error, e.g. user may comma-separate, tab-separate, or use both commas and newlines.

![Image](https://user-images.githubusercontent.com/537316/225206181-7c9a7fb1-5c58-41fe-bcc3-f57181b79f2d.png)

artoonie:
Thinking on this more -- it seems that the notion of "Alias" supersede the notion of a "Code".

i.e. if instead of a Code, users entered a single Alias, wouldn't that be the same?

I'm still familiarizing myself with the source code, but this seems true on the surface. If it's true, I can handle them the same way internally, and we can have a discussion about whether the UI should keep them separated or not.

rrojas350:
[...]

Issue #590, opened 2022-05-09
Ability to Edit Info in Tables


rrojas350:
Users would like the ability to edit tables in CVR files and Candidates, making it easier to edit the info if necessary. This could be achieved by adding a clickable icon to the table/row, such as a pencil, to edit the contents in the table and then save their changes by clicking the icon again.

rrojas350:
Related to #502

tarheel:
@HEdingfield: what do you think? I know we disabled editing in the past because of the confusing UX... but what about a pencil icon (as @rrojas350 suggested) that triggers an editing state for the row? And then, if you modify something in the row, we show Update/Cancel buttons. If there's no good way to fit this in the existing window, opening a separate modal would also be an option.

HEdingfield:
Lowest effort would be to bring back what we had before, because I could probably piece it together by reversing the PRs that removed it. A modal would probably take 2 or 3 times as much effort, and would be tricky (as would the pencil thing), because we're working with a particular widget here that likely has limited APIs.

tarheel:
The pencil approach seems worth investigating, since there's at least the possibility that it's not too bad. I'm inclined to agree that a modal probably isn't worth the trouble, so if the pencil is too difficult or impossible, restoring the old behavior seems OK.

ggilber1:
Glad to see you guys considering restoration of the edit function.  As a former election official who created many election databases, the ability to edit an erroneous entry is always the first option looked for.

HEdingfield:
Important discussion in [this thread](https://github.com/BrightSpots/rcv/pull/639/files#r1141150539) about how this design should work for the aliases field.

chughes297:
Just commenting here to flag that Gretchen Reinmeyer in Arlington requested this feature as well. Seems high priority for the next release. 

May need a bit of context for this comment:

> Important discussion in [this thread](https://github.com/BrightSpots/rcv/pull/639/files#r1141150539) about how this design should work for the aliases field.

I read through and not seeing how this relates? May have missed some discussion. 


> @HEdingfield: what do you think? I know we disabled editing in the past because of the confusing UX... but what about a pencil icon (as @rrojas350 suggested) that triggers an editing state for the row? And then, if you modify something in the row, we show Update/Cancel buttons. If there's no good way to fit this in the existing window, opening a separate modal would also be an option.

I didn't realize this was removed for UX reasons. I don't remember the UX being all that confusing?

HEdingfield:
[...]

PR #639, opened 2023-03-15
Aliases as a replacement for candidate codes

artoonie:
# Implementation Details
An alias is just a general case of the old concept of a "Code". Candidates now have three fields: `name` (aka display name, or canonical name), `code` (only used in the output CSVs - see Question below), and `aliases` (generalized `code`s).

In the tabulation logic, all names encountered are translated from whatever form they're in to the canonical name. All logic is now done on the canonical name, not on the code or any alias.

I have designed this so that the concept of `code` can be removed entirely in time, but have not done so yet.

# Question: removing the concept of "Code" entirely
The current logic on `develop` is for output CSVs to print the Candidate `code` rather than the `name`. In all other cases, `code` is treated exactly like an `alias`.

To ensure 1-to-1 consistency with `develop`, I have opted to leave both "Code" and "Aliases" internally, even though the UI now only shows "Aliases." Removing the concept of "Code" will be trivial, and would clean up the code, but would result in test files differing. I'm not sure if there are other consequences.

*Question*: Do you agree with the current implementation (keeping "Code" internally so there are no changes to any output files), or do you prefer a cleaner implementation (done by removing "Code" and updating all tests that used "Code").
*Answer*: In a separate pull request, we'll get rid of code

# Screenshot: an old Dominion config.
Notice the "Code" is successfully shown as an "Alias"
image

# Screenshot: multiple aliases
Semicolon-separated, which will be easy to translate back and forth into a list when the Table becomes editable.
image

# Test details (on both JSON and XLSX tests)
1. Ensures the non-alias name is the name used in the JSON summary output
2. Ensures that both aliases and names are treated identically otherwise
3. Ensures unused aliases don't cause issues
4. Ensures unused names don't cause issues (i.e. the "Code" use case, where only the alias is used in the CVR and not the name)


src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -88,8 +89,12 @@ class ContestConfig {
...
   // Used to track a sequential multi-seat race
   private final List<String> sequentialWinners = new LinkedList<>();
-  // Mapping from candidate code to full name
-  private Map<String, String> candidateCodeToNameMap;
+  // Canonical candidate names (no aliases or codes)
+  private Set<String> candidateNames;
+  // Mapping from any candidate alias to the candidate's display name
+  private Map<String, String> candidateAliasesToNameMap;
+  // Mapping from any candidate alias to the candidate's code
+  private Map<String, String> candidateAliasesToCodeMap;
```
artoonie, line 97: Note: this is only used when outputting the result CSVs

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1111,7 +1115,6 @@ enum ValidationError {
     CANDIDATE_NAME_MISSING,
     CANDIDATE_NAME_INVALID,
     CANDIDATE_CODE_INVALID,
-    CANDIDATE_DUPLICATE_CODES,
```
artoonie, line 1114: Note: there is no longer a distinction between duplicate names and duplicate codes. Names, codes, and aliases must all be unique.

src/main/resources/network/brightspots/rcv/hints_candidates.txt
```diff
@@ -4,6 +4,9 @@ Fill in the fields and click the Add button:
 
 Name (required): E.g.: Dave Harris.
 
-Code (optional): Some CVR files use codes in lieu of full candidate name. E.g.: "JCD" or "14".
+Codes or Aliases (optional): Enter one code or alias per line.
+Codes are used by some CVR files in lieu of full candidate name. E.g.: "JCD" or "14".
+Aliases are helpful when candidate names are not normalized across jurisdictions, e.g. "JAMES" may be an alias for "James".
```
HEdingfield, line 9: Do we really want these to be case-sensitive? Seems to me like it'd probably be best to run them all through `String.upper()` before processing them in the backend.

I think a better example might be "Jim" might be an alias for "James".

src/main/resources/network/brightspots/rcv/hints_candidates.txt
```diff
@@ -4,6 +4,9 @@ Fill in the fields and click the Add button:
 
 Name (required): E.g.: Dave Harris.
 
-Code (optional): Some CVR files use codes in lieu of full candidate name. E.g.: "JCD" or "14".
+Codes or Aliases (optional): Enter one code or alias per line.
```
HEdingfield, line 7: Please exactly match how it's displayed in the GUI here.

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
```diff
@@ -336,13 +336,12 @@
                     <Insets bottom="4.0" left="4.0" right="4.0" top="4.0"/>
                   </padding>
                 </HBox>
-                <HBox alignment="CENTER_LEFT" spacing="4.0">
-                  <Label text="Code" prefWidth="50.0"/>
-                  <TextField prefHeight="25.0" prefWidth="100.0" fx:id="textFieldCandidateCode"/>
+                <Label text="Aliases, codes, and other spellings (one per line):">
```
HEdingfield, line 339: Nit: remove colon, since we don't use it for other fields.

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
```diff
@@ -374,8 +373,8 @@
                   <columns>
                     <TableColumn fx:id="tableColumnCandidateName" minWidth="20.0"
                       prefWidth="230.0" text="Name"/>
-                    <TableColumn fx:id="tableColumnCandidateCode" minWidth="20.0"
-                      prefWidth="100.0" text="Code"/>
+                    <TableColumn fx:id="tableColumnCandidateAliases" minWidth="20.0"
+                      prefWidth="200.0" text="Aliases / Code" />
```
HEdingfield, line 377: Nit: "Codes" since "Aliases" is plural.

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -286,7 +286,9 @@ private List<CastVoteRecord> parseCastVoteRecords(ContestConfig config, Set<Stri
           reader.readCastVoteRecords(castVoteRecords, source.getContestId());
           // Before we tabulate, we output a converted generic CSV for the CVRs.
           try {
-            ResultsWriter writer = new ResultsWriter().setTimestampString(timestampString);
+            ResultsWriter writer = new ResultsWriter()
+                    .setContestConfig(config)
```
HEdingfield, line 290: I'm baffled by how this was working before without setting the config?!

src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -245,6 +266,56 @@ public boolean isExcluded() {
...
+
+      return Stream.concat(this.aliases.stream(), otherNames.stream());
+    }
+
+    /**
+     * For display purposes, get a semicolon-separated list of aliases, including the code.
+     *
+     * @return a potentially-empty string
+     */
+    public String getSemicolonSeparatedAliases() {
```
HEdingfield, line 293: I'm a little concerned about this. E.g. if a user includes a semicolon in an alias (on a single line), it's a confusing display in the table column. This will particularly become an issue when we allow editing of table cells again (which is P1 issue #590).

Hard to think of a way around this at the moment, other than banning the use of semicolons in aliases, or forcing it to use newlines in the column...

HEdingfield:
> _Question_: Do you agree with the current implementation (keeping "Code" internally so there are no changes to any output files), or do you prefer a cleaner implementation (done by removing "Code" and updating all tests that used "Code").

I personally don't see a big benefit to keeping the "Code" field around at all. We're guaranteed that names will be unique, so why not just display that in the output? Let's confirm with @rrojas350 though.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -196,11 +196,11 @@ void parseXml(List<CastVoteRecord> castVoteRecords)
               Logger.severe("ContestSelection \"%s\" from CVR not found!", contestSelectionId);
               throw new CvrParseException();
             }
-            String candidateId;
+            String candidateName;
```
HEdingfield, line 199: (Not related to this code) Can we get at least one test in here that makes use of the aliases?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1011,12 +1004,16 @@ boolean isExhaustOnDuplicateCandidateEnabled() {
...
 
-  Set<String> getCandidateCodeList() {
-    return candidateCodeToNameMap.keySet();
+  Set<String> getCandidateNames() {
+    return candidateNames;
   }
 
-  String getNameForCandidateCode(String code) {
-    return candidateCodeToNameMap.get(code);
+  String getNameFor(String nameOrAlias) {
```
HEdingfield, line 1011: Nit: Let's call this `getNameForCandidate` or `getCandidateName` instead.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1011,12 +1004,16 @@ boolean isExhaustOnDuplicateCandidateEnabled() {
...
+    return candidateAliasesToNameMap.get(nameOrAlias);
+  }
+
+  String getCodeFor(String nameOrAlias) {
```
HEdingfield, line 1015: Nit: `getCodeForCandidate` or `getCandidateCode`

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -88,8 +89,12 @@ class ContestConfig {
   private final String sourceDirectory;
   // Used to track a sequential multi-seat race
   private final List<String> sequentialWinners = new LinkedList<>();
-  // Mapping from candidate code to full name
-  private Map<String, String> candidateCodeToNameMap;
+  // Canonical candidate names (no aliases or codes)
+  private Set<String> candidateNames;
+  // Mapping from any candidate alias to the candidate's display name
```
HEdingfield, line 94: Nit: display = canonical?

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: I think we should delete this and all related code after confirming with @rrojas350, per my other comment.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -948,7 +941,7 @@ boolean isContinueUntilTwoCandidatesRemainEnabled() {
   }
 
   int getNumDeclaredCandidates() {
-    int size = getCandidateCodeList().size();
+    int size = candidateNames.size();
```
HEdingfield, line 944: How do we know this will be defined yet? Shouldn't we do `getCandidateNames()` here instead?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -957,7 +950,7 @@ int getNumDeclaredCandidates() {
   }
 
   int getNumCandidates() {
-    return getCandidateCodeList().size() - excludedCandidates.size();
+    return candidateNames.size() - excludedCandidates.size();
```
HEdingfield, line 953: Same comment as above. Shouldn't we do `getCandidateNames()` here instead?

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: So what error does it throw if it hits a duplicate name then? CANDIDATE_NAME_INVALID? Might there still be a use for a CANDIDATE_DUPLICATE_NAMES error at least?

src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -315,23 +315,25 @@ private int parseCvrFile(
...
             Integer candidateId = (Integer) rankingMap.get("CandidateId");
             String candidateCode = candidateId.toString();
-            Set<String> candidates = contestIdToCandidateCodes.get(contestId);
-            if (!candidates.contains(candidateCode)) {
+            String candidateName = config.getNameFor(candidateCode);
+            Set<String> candidates = contestIdToCandidateNames.get(contestId);
+            if (!candidates.contains(candidateName)) {
               Logger.severe(
-                  "Candidate code '%s' is not valid for contest '%d'!", candidateCode,
+                  "Candidate code '%s' is not valid for contest '%d'!", candidateName,
```
HEdingfield, line 322: code => name

src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -315,23 +315,25 @@ private int parseCvrFile(
             }
             Integer candidateId = (Integer) rankingMap.get("CandidateId");
             String candidateCode = candidateId.toString();
-            Set<String> candidates = contestIdToCandidateCodes.get(contestId);
-            if (!candidates.contains(candidateCode)) {
+            String candidateName = config.getNameFor(candidateCode);
```
HEdingfield, line 318: Are you still calling this `candidateCode` on purpose?

HEdingfield (changes requested):
Haven't reviewed everything yet, but here are some comments to get your started!

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: Sounds good - shall we do that in a separate PR to show that this code has no side-effects? Or do you prefer it all together?

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: This may be a java convention I'm unfamiliar with -- `getCandidateNames` doesn't do a null check, why would it help?

I can make this safer by:
1. `candidateNames == null ? 0 : candidateNames.size()`
2. Initializing it when declaring it
3. Initialize `candidateNames` in `getCandidateNames()` if needed.

Though, I'll note that the non-null assumption existed in the initial code (unless there's some magic Java does with getters that I'm unfamiliar with).

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: `CANDIDATE_NAME_INVALID` was only ever doing a duplicate check, so I agree we should rename it as such.

src/main/java/network/brightspots/rcv/DominionCvrReader.java
artoonie: Yes, Dominion uses a code which we translate here to the display name. In other files I use `candidateName` not `candidateDisplayName`, but I'm happy to rename the variable here to clarify (or add a comment).

src/main/java/network/brightspots/rcv/RawContestConfig.java
artoonie: I could use a less-common character (\cdot, `·`?), though semicolon has the benefit of making it easy to edit the table inline rather than hitting an edit button and having it populate the original textboxes. I'm happy to add a no-semicolon check when checking if names are valid.

I haven't fully thought through the design in #590: do you envision an inline edit, a modal, or populating the original create fields? It seems like the decision there would affect the design here.

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
artoonie: This is the first textarea I believe. The other fields are labels on a field on the same line, so it's more clear without a semicolon.

I don't have extremely strong opinions on this, but I do want to note that this was intentional: I tried it without a semicolon at first, and this looked better to my eye. Happy to remove it, though -- let me know!

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
artoonie: I don't have super strong opinions here, but do want to flag this was also intentional: although there is no difference to us about codes and aliases, in the CVRs that use a `code`, there can be only one `code`.

I did make the change in this case, but let me know if you prefer it to remain singular after hearing this explanation.

src/main/java/network/brightspots/rcv/TabulatorSession.java
artoonie: The function just never used `config` before, but now we use it to map display name to code. Per your other comment, we'll probably be reverting this if we get rid of `code`.

HEdingfield:
> [artoonie](https://github.com/artoonie) [force-pushed](https://github.com/BrightSpots/rcv/compare/cb1080d3c24ae6bc0b19d7fa188cc9d9d76e38fc..f0cec41612a688da23388a2dfc71e4256b43c213) the feature/issue-591_aliases branch from [cb1080d ](https://github.com/BrightSpots/rcv/commit/cb1080d3c24ae6bc0b19d7fa188cc9d9d76e38fc)to [f0cec41 ](https://github.com/BrightSpots/rcv/commit/f0cec41612a688da23388a2dfc71e4256b43c213)
[3 days ago](https://github.com/BrightSpots/rcv/pull/639#event-8762381392)
@artoonie
[Make CastVoteRecord use 50% less memory](https://github.com/BrightSpots/rcv/pull/639/commits/8035f6410cc48b470a7e438502e67a5bb8788da2)

Not sure why this was done, as that code hasn't been reviewed yet. Could you please keep it out of this branch?

artoonie:
Ugh sorry, let me do some git cleanup. I'm using the github desktop client for the first time and it did this behind the scenes...not sure why or how. Will fix.

artoonie:
> Can someone explain to me in English (and without acronyms) what is being done here? Thanks. I'll assign this to @rrojas350 since you are probably at at least an intermediate level of technical comprehension.

Check out the Issue linked to this pull request here: https://github.com/BrightSpots/rcv/issues/591
In sum: there are often cases where a candidate has multiple names across different CVRs -- for example, write-ins, or multi-vendor tabulations. This provides support for those cases: a candidate can now have aliases, and we will look for both their name and their aliases in CVRs.

ggilber1:
Got the overall idea.  I am unclear about some of the details, eg., how
this is being handled in regard to vendor systems which use candidate codes
rather than candidate names in their CVRs.  I very much favor pulling the
candidate names into the config, which it appears you are doing.  Every
election administrator wants to be able to look one place to ck names on a
ballot vs. candidate lists.

Also need to understand how this issue relates to the issue of
auto-populating candidate names in the config tile.

Thanks, Armin.

George Gilbert
***@***.***
336-906-0047

artoonie:
@ggilber1 It now treats Candidate Codes and Aliases in the same manner. An election administrator using ballots with candidate codes would enter the Code in the new textbox. An election administrator using aliases would enter the aliases in that same textbox. They can also mix and match codes and aliases if they have a multi-vendor tabulation that needs both codes and aliases.

This probably does not affect the auto-populating of candidate names in any way, though I haven't given that enough thought to be confident about it.

rrojas350:
@HEdingfield If I'm understanding correctly, I'm ok getting rid of candidate code in the output.

Judging from @artoonie's reply to @ggilber1, it sounds like candidate codes on the front end shouldn't cause any issues when trying to tabulate Hart and Dominion CVRs, which is the only concern I have regarding candidate codes.

andyanderson:
> Also need to understand how this issue relates to the issue of auto-populating candidate names in the config tile.
> […](#)

Currently the software ignores CVRs for candidates who aren’t pre-populated in the software, i.e. write-ins. Auto-populating them eliminates the need for an administrator to enter them by hand. Then all they need to do is correlate them in the alias table if there are multiple variants (likely).

src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -194,6 +194,12 @@ static void setup() {
     Logger.setup();
   }
 
+  @Test
+  @DisplayName("aliases")
+  void aliases() {
+    runTabulationTest("aliases");
```
HEdingfield, line 200: Amazing test! But dear god the CDF format is hard to parse... is there a reason you chose to have the tests use that, rather than a more standard format?

But now that you've already done the work of creating it, maybe it's worth having both types of tests with the aliases (e.g. CDF and some xlsx or CSV input format)? This could be overkill, it's just really hard for me to vet that this test is actually WAI.

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: A follow-up PR sounds good, since this one is already a monster. Would you mind filing an issue for this or noting it in an existing related issue that we make sure to handle before closing? You could link to this comment thread for context.

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: I think it's just generally good practice to use the getter instead of accessing the variable directly, in case we ever want to change the behavior in the future.

I like the idea of making the getter check if it's not defined yet, and returning an empty set if it's not.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -863,7 +865,6 @@ private void clearConfig() {
     tableViewCvrFiles.getItems().clear();
 
     textFieldCandidateName.clear();
-    textFieldCandidateCode.clear();
```
HEdingfield, line 866: I'd think you'd want to replace this with:
`textAreaCandidateAliases.clear();`

src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -315,23 +315,25 @@ private int parseCvrFile(
...
               throw new CvrParseException();
             }
             // We also need to throw an error if this candidate doesn't appear in the tabulator's
             // config file for this contest.
             if (candidateCode.equals(undeclaredWriteInLabel)) {
-              candidateCode = Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL;
-            } else if (!config.getCandidateCodeList().contains(candidateCode)) {
+              candidateName = Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL;
+            } else if (candidateName == null) {
               unrecognizedCandidateCounts.merge(candidateCode, 1, Integer::sum);
```
HEdingfield, line 331: Really not sure at this point, but we still want to merge `candidateCode` here instead of `candidateName`? Only asking because I see `candidateName` being merged in `HartCvrReader`.

src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -230,6 +243,14 @@ public void setName(String name) {
       this.name = name;
     }
 
+    public List<String> getAliases() {
+      return List.copyOf(aliases);
```
HEdingfield, line 247: Just curious: why a re we returning a `copyOf` here instead of the object itself?

src/main/java/network/brightspots/rcv/RawContestConfig.java
HEdingfield: The way I kind of had it working before was double-clicking the cell would allow you to edit it and type right in there, then you had to press enter to commit it. Definitely open to other designs though. It'd probably actually be really cool if we had a modal pop up with the full text area to edit, but that would take some work.

Maybe we can go with a happy medium, like a pipe | since that's at least typeable on a standard keyboard and my gut tells me it's less likely to appear in an alias than a semicolon??

Alternatively, a no-semicolon check would work too. I'm happy either way. @chughes297 might have useful thoughts on this.

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
HEdingfield: I'm ok with leaving it given that justification. :)

src/main/java/network/brightspots/rcv/TabulatorSession.java
HEdingfield: Sounds good -- let's note / link to this comment in the issue / comment you make to remove code.

src/test/java/network/brightspots/rcv/TabulatorTests.java
artoonie: I just grabbed a random test and copied/pasted. Happy to update with two additional formats testing.

src/main/java/network/brightspots/rcv/TabulatorSession.java
artoonie: https://github.com/BrightSpots/rcv/issues/663

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: https://github.com/BrightSpots/rcv/issues/663

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: Happy to make these changes, but it feels very strange to me to have `getCandidateNames()` alongside `excludedCandidates`. I could add a `getExcludedCandidates` function, but...we'd be creating a lot of bloat.

I'm used to using getters and setters from outside the class, and accessing members directly within the class. Happy to follow your style here, but do want to flag that there are dozens of direct member accesses throughout this file that aren't getting the same treatment.

src/main/java/network/brightspots/rcv/DominionCvrReader.java
artoonie: `candidateName` is null here, but code is probably not

src/main/java/network/brightspots/rcv/RawContestConfig.java
artoonie: This is public, so I'm afraid of a caller assuming they own the List and mutating it. It's not currently used, just added for completeness, so I'd rather be safe than let a usage bug sneak in. I can also just delete the function entirely.

src/main/java/network/brightspots/rcv/RawContestConfig.java
artoonie: I figured semicolon was unlikely to appear in a name, but happy to swap with a pipe, or leave as-is and use newlines with a full text area modal popup (or even inline textarea). Neither will be a huge lift, so happy to change this to whichever you prefer.

src/test/java/network/brightspots/rcv/TabulatorTests.java
artoonie: The only CSV input format I see is ClearBallot, where this doesn't apply: tabulation occurs by putting values in a `1` in a column corresponding to a unique candidate, so it seems that aliases are already consolidated. I could create a separate column with an alias, but that doesn't feel super useful.

I have gone ahead and added an XLSX test

HEdingfield (commented):
Very close now! I think we'll all breathe a sigh of relief once this one is submitted.

src/main/java/network/brightspots/rcv/RawContestConfig.java
artoonie: The backend can save a list, so it won't have newline or semicolon separation -- just confirming that's okay?

artoonie:
All comments resolved. Thanks for catching the accidentally-updated test -- perhaps we should stop the UI from forcing you to re-save a config if the version is "TEST"? It's a pain to constantly revert those. I can make a ticket for that if you think it'd help y'all as well.

src/main/java/network/brightspots/rcv/DominionCvrReader.java
HEdingfield: Not sure if you missed this comment, but I think it'd be nice to take care of this still.

src/test/resources/network/brightspots/rcv/test_data/aliases_cdf_json/aliases_cdf_json_config.json
```diff
@@ -0,0 +1,68 @@
+{
+  "tabulatorVersion": "TEST",
```
HEdingfield, line 2: > Thanks for catching the accidentally-updated test -- perhaps we should stop the UI from forcing you to re-save a config if the version is "TEST"? It's a pain to constantly revert those. I can make a ticket for that if you think it'd help y'all as well.

Yeah, I think that's a good idea, please do file an issue for that.

src/main/java/network/brightspots/rcv/RawContestConfig.java
HEdingfield: Yeah, that sounds perfect. Does that require code changes to this PR, or is it purely something you'd want to handle in future work around this?

HEdingfield:
Happy to add a commit that takes care of this last bit if you'd like (also happy to wait if you'd rather do it yourself), just let me know.

src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -245,6 +266,56 @@ public boolean isExcluded() {
     public void setExcluded(boolean excluded) {
       this.excluded = excluded;
     }
+
+    /**
+     * A stream of all aliases (which is guaranteed to be unique) and the candidate name
+     * (which is not guaranteed to be unique, i.e. it may exist in the list twice)
```
tarheel, line 272: why might the name be present twice?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1011,12 +1004,19 @@ boolean isExhaustOnDuplicateCandidateEnabled() {
     return rawConfig.rules.exhaustOnDuplicateCandidate;
   }
 
-  Set<String> getCandidateCodeList() {
-    return candidateCodeToNameMap.keySet();
+  Set<String> getCandidateNames() {
+    if (candidateNames == null) {
+      candidateNames = new HashSet<>();
+    }
```
tarheel, line 1010: nit: why not just initialize this when you declare it?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1032,32 +1032,39 @@ void setCandidateExclusionStatus(String candidateCode, boolean excluded) {
...
   // perform pre-processing on candidates:
-  // 1) build map of candidate ID to candidate name
+  // 1) build map of candidate aliases to candidate name
   // 2) generate tie-break ordering if needed
   // 3) add uwi candidate if needed
   private void processCandidateData() {
-    candidateCodeToNameMap = new HashMap<>();
+    candidateAliasesToNameMap = new HashMap<>();
+    candidateAliasesToCodeMap = new HashMap<>();
+    candidateNames = new HashSet<>();
```
tarheel, line 1041: similarly, why not just initialize these when you declare them?

src/main/java/network/brightspots/rcv/RawContestConfig.java
tarheel: You're saying that it'll just use standard JSON list format, right?

tarheel (commented):
Sorry for taking forever to comment on this. It makes sense to me! And I definitely agree that we should dump candidate codes.

src/main/java/network/brightspots/rcv/RawContestConfig.java
artoonie: It's valid for a candidate name to be equal to a candidate code, and therefore for a candidate name to be equal to an alias.

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: 1. Nothing else is initialized when declared
2. This was a request from @HEdingfield to add this: https://github.com/BrightSpots/rcv/pull/639#discussion_r1181277098
https://github.com/BrightSpots/rcv/pull/639#discussion_r1179981350

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: Just following the existing model, a la `validateCandidates`. I don't mind changing these to init when declared, but perhaps we should have a larger movement to do that consistently across this and other files?

src/main/java/network/brightspots/rcv/RawContestConfig.java
artoonie: Yes, that's correct, and that is how it's saved today.

src/test/resources/network/brightspots/rcv/test_data/aliases_cdf_json/aliases_cdf_json_config.json
artoonie: https://github.com/BrightSpots/rcv/issues/664

src/main/java/network/brightspots/rcv/DominionCvrReader.java
artoonie: Looks like the comment didn't submit before :) both your comments show as appearing 5 days ago. Yes, I can rename this variable `dominionCandidateId`

HEdingfield: Oh weird, sorry about that. Looks like we got our wires crossed a bit on this, so I just added a commit to clear it up. Please give it a look over to verify that it makes sense.

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: I think either way would probably work fine. Also agree with @artoonie that we should make a pass on all this at some point to make it consistent.

HEdingfield (approved):
Ok, I think this is all set! I added a small commit that fixes a couple small bugs and cleans stuff up more, so will let you go ahead and look that over before you merge it. Great work on this!

PR #640, opened 2023-03-16
Make CastVoteRecord use 50% less memory

artoonie:
The biggest memory footprint came from using the HashMap `SortedMap<Integer, Set<String>> rankToCandidateNames`.

That structure was helpful when storing the edge case of skipped and duplicated rankings. I changed this to an array of arrays instead with some custom classes. The new structure is optimized for the common case: a valid ballot. It works fine for invalid ballots.

This pull request has two changes:
1. Change `rankToCandidateNames` to `candidateRankings`. This reduces the CVR structure from 1860 bytes to 877 bytes on one test.
2. Remove `fullCvrData`, reducing 877 bytes to 580 bytes.

Using `String.intern()` had no impact. I'm not sure if strings are interned elsewhere or if there's something else holding on to the non-interned string.

This cuts the memory footprint of CVRs to ~30% its original size. I'm still getting my feet wet learning Java memory profiling, so I'm not fully confident on how much this reduces the peak memory footprint.


HEdingfield:
I'm confused. It looks like this PR is no longer targeted to the right branch (develop), and that was force pushed into the other PR related to aliases. Could you please undo this and give @tarheel a chance to review this one first?

artoonie:
Fixed -- though this never targeted develop, since it's based off of the other branch, and I want to keep the diff here small.
When `issue-552` is merged, GitHub will automatically redirect this PR to point at develop.

HEdingfield:
Sounds good, thanks! You might want to resolve the conflicting files before @tarheel reviews.

src/main/java/network/brightspots/rcv/CastVoteRecord.java
```diff
@@ -230,13 +215,8 @@ Map<String, BigDecimal> getWinnerToFractionalValue() {
   }
 
   // create a sorted map of ranking to candidates selected at that rank
```
tarheel, line 217: this comment is no longer quite accurate

src/main/java/network/brightspots/rcv/StreamingCvrReader.java
```diff
@@ -199,13 +199,16 @@ private void endCvr() {
       encounteredDataErrors = true;
     }
 
+    // Log the raw data for auditing
+    Logger.fine(" [Raw Data] ");
+    Logger.fine(currentCvrData.toString());
```
tarheel, line 204: So, as suggested in the filed issue, you've moved the raw data logging so that it all happens when we're first ingesting the CVR rather than being included as part of the round 1 output for each CVR. We should run this by the RCVRC folks to confirm that they're OK with the slight change to the audit log format.

Also worth noting that (I believe) the audit log doesn't contain the raw ingested data at all when the CVR source isn't XLS...

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -812,7 +813,8 @@ private OvervoteDecision getOvervoteDecision(Set<String> candidateSet) {
       decision = OvervoteDecision.NONE;
       // keep track if we encounter a continuing candidate
       String continuingCandidate = null;
-      for (String candidate : candidateSet) {
+      for (int candidateIndex = 0; candidateIndex < candidates.count(); ++candidateIndex) {
+        String candidate = candidates.get(candidateIndex);
```
tarheel, line 817: can this just be `for (var candidate : candidates)`?

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -927,23 +929,23 @@ && isCandidateContinuing(cvr.getCurrentRecipientOfVote())) {
       String selectedCandidate = null;
 
       // iterate over all ranks in this cvr from most preferred to least
-      for (int rank : cvr.rankToCandidateNames.keySet()) {
+      for (Pair<Integer, CandidatesAtRanking> rankCandidatesPair : cvr.candidateRankings) {
+        CandidatesAtRanking candidates = rankCandidatesPair.getValue();
+
         // check for undervote exhaustion
+        Integer rank = rankCandidatesPair.getKey();
```
tarheel, line 936: nit: move this up a couple lines

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -927,23 +929,23 @@ && isCandidateContinuing(cvr.getCurrentRecipientOfVote())) {
...
-        // candidateSet contains all candidates selected at the current rank
-        // some ballots support multiple candidates selected at a single rank
-        Set<String> candidateSet = cvr.rankToCandidateNames.get(rank);
-
         // check for a duplicate candidate if enabled
         if (config.isExhaustOnDuplicateCandidateEnabled()) {
           String duplicateCandidate = null;
-          for (String candidate : candidateSet) {
+          for (int candidateIndex = 0; candidateIndex < candidates.count(); ++candidateIndex) {
+            String candidate = candidates.get(candidateIndex);
```
tarheel, line 948: same as above -- I think this can be simpler

tarheel (commented):
Looks like a big improvement!

src/main/java/network/brightspots/rcv/CandidateRankingsList.java
```diff
@@ -0,0 +1,121 @@
...
+      return;
+    }
+
+    rawRankings.sort(Comparator.comparingInt(Pair::getKey));
+
+    // Initialize up to maxRankings, leaving empty arrays for any undervotes
+    int minRanking = rawRankings.get(0).getKey();
+    if (minRanking <= 0) {
+      Logger.severe("Rankings must start at 1, but you have a ranking at %d", minRanking);
+      // TODO should I throw from here or not?
```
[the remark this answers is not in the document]
artoonie: Since this should never happen, I don't want to litter the code with `throws` and `try/catch`. Is it okay to throw an unchecked exception here (`RuntimeException`?) Not sure what the best practices are in this regard.

src/main/java/network/brightspots/rcv/CandidateRankingsList.java
```diff
@@ -0,0 +1,121 @@
...
+      }
+    }
+  }
+
+  CandidatesAtRanking get(int i) {
+    return rankings[i - 1];
+  }
+
+  boolean hasRankingAt(int num) {
+    return num >= 1 && num <= rankings.length && rankings[num - 1].count() != 0;
```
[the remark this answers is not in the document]
artoonie: Not sure I follow -- you mean, just return `rankings[num - 1].count() != 0` and don't check if `num` is in bounds?

I think the caller should at least be able to have `num > rankings.length`, since a natural call may be `while(++i) hasRankingAt(i)` -- i.e. don't check against `maxNumRankings`. I'm fine asserting that `num >= 1` though, since a caller should never call with a zero or negative index.

src/main/java/network/brightspots/rcv/CandidateRankingsList.java
```diff
@@ -0,0 +1,121 @@
...
+
+  private CandidatesAtRanking[] rankings;
+  private int numRankings;
+  private int iteratorIndex = 0;
+
+  CandidateRankingsList(List<Pair<Integer, String>> rawRankings) {
+    if (rawRankings.isEmpty()) {
+      this.rankings = new CandidatesAtRanking[0];
+      numRankings = 0;
+      return;
```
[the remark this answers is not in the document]
artoonie: That includes returning early from a function? 👍

artoonie:
All PR Comments are addressed @tarheel

src/main/java/network/brightspots/rcv/StreamingCvrReader.java
tarheel: I would just email Rene/Chris/George to explain what's changing and see if anyone objects. It's not conforming to any official standard; it's just something we agreed on as a group years ago.

src/main/java/network/brightspots/rcv/CandidateRankingsList.java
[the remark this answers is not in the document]
tarheel: Unfortunately, yes. Having short-circuit conditions in functions can make them a lot more readable, so this is a pretty annoying requirement. But we can't change it.

src/main/java/network/brightspots/rcv/CandidatesAtRanking.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  private String[] candidateNames;
+
+  int count() {
+    int numCandidates = 0;
+    if (candidateNames != null) {
+      numCandidates = candidateNames.length;
+    }
+    return numCandidates;
```
tarheel, line 60: to be concise:
```
return candidateNames == null ? 0 : candidateNames.length;
```

src/main/java/network/brightspots/rcv/CastVoteRecord.java
```diff
@@ -229,14 +214,9 @@ Map<String, BigDecimal> getWinnerToFractionalValue() {
...
-  // create a sorted map of ranking to candidates selected at that rank
-  private void sortRankings(List<Pair<Integer, String>> rankings) {
-    rankToCandidateNames = new TreeMap<>();
-    for (Pair<Integer, String> ranking : rankings) {
-      Set<String> candidatesAtRank =
-          rankToCandidateNames.computeIfAbsent(ranking.getKey(), k -> new HashSet<>());
-      candidatesAtRank.add(ranking.getValue());
-    }
+  // initialize the CandidateRankingsList
+  private void loadRankings(List<Pair<Integer, String>> rankings) {
```
tarheel, line 218: nit: no need for a separate function for this

src/main/java/network/brightspots/rcv/StreamingCvrReader.java
tarheel: Also, just want to confirm: this still goes to the correct file, yes?

src/main/java/network/brightspots/rcv/CandidatesAtRanking.java
artoonie: I wasn't sure if trinary operaters counted as multiple returns -- I did some investigation and the bytecode looked [sufficiently close](https://stackoverflow.com/a/63045843/1057105) to `if/else`. Further, the exact implementation isn't specified, so it can vary from implementation to implementation.

All this to ask: what are the exact rules about multiple returns, and are we sure trinary operators follow these rules?

src/main/java/network/brightspots/rcv/CandidatesAtRanking.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public String next() {
+      if (iteratorIndex == candidates.candidateNames.length) {
+        throw new NoSuchElementException();
+      }
+
+      iteratorIndex++;
+
+      return candidates.get(iteratorIndex - 1);
```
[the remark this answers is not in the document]
artoonie: As in, you prefer `return candidates.get(iteratorIndex++)`?

src/main/java/network/brightspots/rcv/StreamingCvrReader.java
artoonie: Yes -- I'll post a before/after audit log to ensure everyone is on board.

src/main/java/network/brightspots/rcv/CandidatesAtRanking.java
tarheel: I really don't think the certification folks are looking at the bytecode. :) I think they just want to see that every function has at most one `return` statement and, if present, it's the final statement. So a ternary should be fine, as far as I know.

src/main/java/network/brightspots/rcv/CandidatesAtRanking.java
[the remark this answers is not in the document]
tarheel: Yeah, I guess so. I really don't feel strongly about it, though.

src/main/java/network/brightspots/rcv/CandidateRankingsList.java
```diff
@@ -0,0 +1,117 @@
...
+      // Initialize up to maxRankings, leaving empty arrays for any undervotes
+      int minRanking = rawRankings.get(0).getKey();
+      if (minRanking <= 0) {
+        throw new RuntimeException("Rankings must start at 1, but you have a ranking at %d"
+            .formatted(minRanking));
+      }
+
+      int maxRanking = rawRankings.get(rawRankings.size() - 1).getKey();
+      this.rankings = new CandidatesAtRanking[maxRanking];
+      for (int i = 0; i < maxRanking; ++i) {
```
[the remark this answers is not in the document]
artoonie: I wonder if we could catch this with the checkstyle file -- it looks possible, but I'm not familiar with checkstyle and not sure we want to stray from the default Google file.

src/main/java/network/brightspots/rcv/CandidateRankingsList.java
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * Purpose: Low-Memory container for a single cast vote record
+ * Design: Assumes valid ballots but supports overvotes and skipped rankings.
+ * Conditions: Always.
```
tarheel, line 13: what does this mean in headers?

tarheel (approved):

artoonie:
There were a few dozen merge conflicts, not sure why -- perhaps caused by squash commits? Anyway, would love another sanity check before I merge this @tarheel or @HEdingfield

HEdingfield:
> There were a few dozen merge conflicts, not sure why -- perhaps caused by squash commits? Anyway, would love another sanity check before I merge this @tarheel or @HEdingfield

Sorry, for some reason I didn't get an email about this... it looks like there was a bad merge in `DominionCvrReader`. I fixed the part I noticed in PR #665.

PR #644, opened 2023-03-22
Basic CDF validation

artoonie:
Some basic validation for CDF, plus a generic try/catch to handle uncaught errors which outputs a generic error message.

## 1. Invalid Contest Selection
*Problem*: Using a nonexistent ID for Contest Selection is already handled with an error like  `ContestSelection "ics-b" from CVR not found!`

## 2. Invalid `CandidateIds`
*Problem*: Entering an invalid candidate id in `CandidateIds` in `Election>Contest>ContestSelection`.
*Previously*: An error message: `Cannot invoke "java.util.HashMap.get(Object)" because "candidate" is null`
*Now*: `Candidate ID "candidate-f" in Contest Selection "cs-c" is not in the candidate list.`

## 3. Invalid Candidate `@ID`
*Problem*: Duplicate Candidate `@id` in `Election>Candidate` is not handled.
*Previously*:  Uses last-entered Candidate ID and overwrites any previous usage.
*Now*: Message like `Candidate "candidate-d" appears multiple times`

## 4. Duplicate Candidate `Name`
*Problem*: Having multiple candidates in the candidate list with the same name
This is caught in validation, so not making any changes.

## 5. Multiple `CandidateIds` in `Election>Candidate`
This will be supported soon thanks to #639. The new issues #645 will handle this.


tarheel:
Is this one ready for review? (Not sure what your workflow is. Sometimes people create a PR but don't add reviewers until it's ready for review.)

artoonie:
Yes -- unless I add a "WIP" label, the PR is ready for review. I'll also drag the corresponding task here to "In Review" when it's ready for review: https://github.com/orgs/BrightSpots/projects/1/views/4

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -314,7 +361,7 @@ void parseJson(List<CastVoteRecord> castVoteRecords)
     for (Object gpUnitObject : gpUnitArray) {
       HashMap gpUnit = (HashMap) gpUnitObject;
       String gpUnitId = (String) gpUnit.get("@id");
-      gpuUnits.put(gpUnitId, gpUnit);
+      putIfUnique(gpuUnits, gpUnitId, gpUnit, "GPUUnits");
```
HEdingfield, line 364: I think `gpuUnits` is a typo and it should be `gpUnits`; would you mind fixing?

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -85,11 +85,38 @@ CVRContest getCvrContestXml(CVR cvr, Contest contestToTabulate) {
...
+      throw new CvrParseException();
+    }
+  }
+
+  /**
+   * Sets map[key] = value if map does not contain key already.
+   * Throws CvrParseException if it does.
+   *
+   * @param map The map to modify
+   * @param humanReadableName Used as part of the error message if the key is already in the map
```
HEdingfield, line 109: Nit: Let's just call `humanReadableName` => `keyName` instead, and the description can say it's a "human-readable name of the key, used to make the error message more useful".

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -132,6 +159,25 @@ void parseXml(List<CastVoteRecord> castVoteRecords)
...
+      // Some checks to provide nicer error messages.
+      // This is common with Unisyn's CDF CVR, which is not compatible with RCTab.
+      if (cvrReport.GpUnit == null) {
+        Logger.severe("Field \"GPUnit\" missing from CDF CVR file! "
+                + "This is common with older, unsupported formats.");
+        throw new CvrParseException();
+      }
+
+      // These fields are also required, but we are not aware of any standard format where
+      // they would be missing, so we group all these checks together.
```
HEdingfield, line 171: From here down, it seems to me like it might be slightly more helpful to throw a specific error for each field, even if we don't know a standard format where they might be missing.

The way I'd probably design it is to initialize a `fieldsAreMissing = false` variable, then go through each check and set it to true if any trigger. Then at the very end I'd throw the `CvrParseException` if it's `true`. That way we can clearly see every field that's missing laid out in the logs.

If you feel strongly otherwise though, feel free to leave as-is.

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -155,20 +201,21 @@ void parseXml(List<CastVoteRecord> castVoteRecords)
       HashMap<String, Candidate> candidateById = new HashMap<>();
       for (Election election : cvrReport.Election) {
         for (Candidate candidate : election.Candidate) {
-          candidateById.put(candidate.ObjectId, candidate);
+          putIfUnique(candidateById, candidate.ObjectId, candidate, "Candidate");
```
HEdingfield, line 204: Similar to my above comment, the way I like to design validation flows typically is to have a boolean that gets triggered if any conditions fail, but only check that boolean at the very end all the validation and throw the exception at that point. This allows us to avoid situations where there are maybe 30 things wrong with an input, but we don't need to run it 30 times (each after fixing the one error it gives us) to figure every one of these things out. As is, this helper throws the exception, so it doesn't behave that way.

Let me know if that doesn't make sense, or if you disagree.

HEdingfield (commented):
Great start on this!

artoonie:
Makes sense -- all comments addressed! A bit of repeated code, worth considering separating out the logic of "find all errors of Type X and check at the end" logic. It's used here twice, when finding undeclared candidates, and perhaps more places.

Something like:
```
class ErrorAccumulator() {
   private Dictionary<string,int> errorCounts;
   void addError(string);
   void verify() throws CvrParseException();
}
```

src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
```diff
@@ -85,13 +86,44 @@ CVRContest getCvrContestXml(CVR cvr, Contest contestToTabulate) {
...
+   * @param key Key to check for uniqueness
+   * @param value Value to place if unique
+   * @param keyName Human-readable name of the key, used to make the error message more useful
+   * @param errors If not unique, adds keyName to this list
+   */
+  private static <K, V> void putIfUnique(
+          HashMap<K, V> map, K key, V value, String keyName, HashSet<String> errors) {
+    if (map.containsKey(key)) {
+      Logger.severe("%s \"%s\" appears multiple times", keyName, key);
+      errors.add(keyName);
```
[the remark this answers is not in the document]
artoonie: I've double-checked, the mutability works fine here. This makes method makes it easier to read the output if there are many duplicate fields by giving a count of the number of unique errors; happy to make the change though. Code has been updated to reflect this.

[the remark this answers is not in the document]
HEdingfield: Ah ok, fair enough. I think this is better anyway, as I dislike the "magical" implicit nature of using mutability like that (unless it's saving a significant number of lines).

HEdingfield (approved):
Nice work!

PR #646, opened 2023-03-31
Allow single-winner threshold to be based on first-round votes only

artoonie:
## Single-Winner races (enabled):
image

## Multi-Winner races (disabled):
image

Included a simple test to verify.


tarheel:
Is this one ready for review?

artoonie:
@tarheel yep

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -158,7 +158,18 @@ Set<String> tabulate() throws TabulationAbortedException {
       // votes in the first round.
       // In a single-seat contest or in the special multi-seat bottoms-up threshold mode, it's based
       // on the number of active votes in the current round.
-      if (currentRound == 1 || config.getNumberOfWinners() <= 1) {
+      boolean doRecomputeThreshold;
```
tarheel, line 161: super nit: I think `shouldRecomputeThreshold` would be more clear

tarheel (commented):
In addition to (or instead of) the test case you created, could you include one in which the winner fails to reach the threshold? I just want to confirm that that case works, since (I believe) it violates an assumption that we previously held about how thresholds work.

tarheel:
@chughes297 can probably confirm, but I believe the desired behavior would be to stop when there are two candidates remaining and elect the candidate with more votes.

On a related note: it just occurred to me that I'm not 100% sure we handle it properly if you end up with two candidates remaining and they're tied. Ordinarily we go to the tie-breaker logic, but I'm not sure that accounts for this scenario...

artoonie:
The tiebreaker logic occurs to figure out which of the two candidates to eliminate, but not which of the two to elect. The tabulation still runs down to just one candidate.

I've added another commit with a test for this so you can see how it works.

Would love to hear what Chris thinks. Chris, any word on whether it's okay to run down to one candidate when using this rule?

andyanderson:
> Modified test so threshold isn't reached, but winner is still selected based on being the sole remaining candidate.
> 
> This may not be the desired behavior: the rounds continue until just one candidate remains, rather than picking whoever has the most votes when two candidates remain. I'm not sure if this is the desired behavior? Should nobody be elected? Should somebody be elected when two candidates remain?

Eliminating the second-place candidate might be necessary to push the first-place candidate over the threshold, so this seems to me to be desirable.

But is there a description somewhere of how this would be an impactful change? There must be some other consequence planned if they haven’t reached the threshold, e.g. nobody elected, rerun the election, etc., because once a candidate reaches a majority of the vote they will always have a majority of the vote.

tarheel:
> But is there a description somewhere of how this would be an impactful change? There must be some other consequence planned if they haven’t reached the threshold, e.g. nobody elected, rerun the election, etc., because once a candidate reaches a majority of the vote they will always have a majority of the vote.

Ha, you would think. But apparently it's just something that's now required in Minnesota even though it has no impact on the outcome.

endolith:
> because once a candidate reaches a majority of the vote they will always have a majority of the vote.

Ah, but if you eliminate the 2nd candidate, the winner is now guaranteed to always have _unanimous_ support! 🤣

andyanderson:
Thinking more about this, I can see that this requirement of surpassing the original threshold is going to result in strategic voting in the form of candidates encouraging their supporters to “rank me first and don’t rank anyone else, because the more exhausted ballots the less likely it is that another candidate might win”. A bad, bad idea, Minnesota.

cjerdonek:
> > because once a candidate reaches a majority of the vote they will always have a majority of the vote.
> 
> Ah, but if you eliminate the 2nd candidate, the winner is now guaranteed to always have _unanimous_ support! 🤣

I was randomly looking at issues and saw this. Note that even if you eliminate the 2nd-place out of two candidates left, it's still possible that the winner won't reach the threshold established in the first round (for example, if none of the ballots for the losing candidates rank the other candidate further down).

artoonie:
@cjerdonek I agree, but I think this has some bizarre consequences.

I've added a commit which feels very dirty to me, but is probably the correct way of doing things...
1. If nobody reaches the threshold, it doesn't mark anybody as a winner
2. But, as a natural consequence, the lowest vote-getter should be eliminate
3. Therefore, there's a final round with no candidates remaining

Here's a [visualization](https://rcvis.com/v/portland-2015-mayoral-race-3), with a table below.

Is this what we want? This feels better than leaving the last-vote-getter uneliminated if they didn't reach the threshold, since that looks too similar to the last-vote-getter being elected.

image

@chughes297 can you look at these visualizations and let me know if you think that's how the code should react when nobody reaches the threshold?

chughes297:
> @chughes297 can probably confirm, but I believe the desired behavior would be to stop when there are two candidates remaining and elect the candidate with more votes.
> 
> On a related note: it just occurred to me that I'm not 100% sure we handle it properly if you end up with two candidates remaining and they're tied. Ordinarily we go to the tie-breaker logic, but I'm not sure that accounts for this scenario...

This, from upthread, is the desired behavior. It's functionally only a change in how the results information is reported - all the Minnesota RCV jurisdictions require the results to report winners based on the threshold set by valid first-round votes. It's essentially IRV reported using STV rules. Which means that even if no one crosses threshold the person with the most votes at the end wins. The final round is always a head-to-head matchup. Minneapolis Mayor 2021 Results example: https://vote.minneapolismn.gov/media/-www-content-assets/documents/2021-Mayor-Round-Results-Summary.pdf. 

Sorry for not including an example originally! Would have saved a lot of back and forth.

chughes297:
> > Oh sorry just noticed the tiebreaker question - I would treat a top-two-candidates tie as a tie between candidates to be elected. Does that answer that question?
> 
> Sorry, I'm not exactly sure what you mean. I _think_ the desired behavior should be to run the tie-breaker and then stop and declare the winner of the tie-breaker to be the election winner. (In any earlier round, we would run the tie-breaker and then eliminate the loser, but that doesn't make sense in this context because it would lead to an additional round in which only the winner remains.) And I haven't confirmed, but I don't believe that's what will happen with the code as it currently stands.

Yes, the desired behavior is to run the tie-breaker and then stop and declare the winner of the tie-breaker to be the election winner. I agree it will look weird in the results as people discussed elsewhere, but that's something that's better left to a textual explanation in the results, IMO.

artoonie:
I reverted the last commit to return to the original functionality: when there are two remaining candidates, the one with the most votes wins.

We can check tiebreaker consistency as part of #666. I believe all comments here are now resolved?

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -1135,6 +1142,7 @@ public LocalDate fromString(String string) {
           textFieldNumberOfWinners.setDisable(false);
           checkBoxBatchElimination.setDisable(false);
           checkBoxContinueUntilTwoCandidatesRemain.setDisable(false);
+          checkBoxFirstRoundDeterminesThreshold.setDisable(false);
```
HEdingfield, line 1145: The note in the txt file seems to imply that it only works for single-winner. Doesn't that conflict with this?

HEdingfield (commented):
I'll rely on @tarheel to give the final approval, since he was much closer to this than I was. Just a couple questions from my end.

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: It theoretically works with sequential winner-takes-all, but doesn't make any political sense and isn't needed by anybody.

I'll remove this to make it match the notes.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -846,6 +847,10 @@ boolean isSingleWinnerEnabled() {
     return getWinnerElectionMode() == WinnerElectionMode.STANDARD_SINGLE_WINNER;
   }
 
+  boolean isFirstRoundDeterminesThresholdEnabled() {
```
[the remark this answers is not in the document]
artoonie: Added validation.

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -158,7 +158,18 @@ Set<String> tabulate() throws TabulationAbortedException {
...
+      boolean shouldRecomputeThreshold;
+      if (config.getNumberOfWinners() > 1) {
+        // Multi-winner only computes threshold on round 1
+        shouldRecomputeThreshold = currentRound == 1;
+      } else {
+        // Single-Winner recomputes threshold on round 1 always,
+        // and on other rounds if First Round Determines Threshold is not set
+        shouldRecomputeThreshold =
+                !config.isFirstRoundDeterminesThresholdEnabled()
+                || currentRound == 1;
```
tarheel, line 170: this quadruple indentation seems pretty off. the auto-styling leads to this?

src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
```diff
@@ -22,6 +22,8 @@ Use Batch Elimination: Batch elimination, or simultaneous elimination of all can
 
 Continue until Two Candidates Remain: Single-winner ranked-choice voting elections can stop as soon as a candidate receives a majority of votes, even though 3 or more candidates may still be in the race. Selecting this option will run the round-by-round count until only two candidates remain, regardless of when a candidate wins a majority of votes. Available only when Winner Election Mode is "Single-winner majority determines winner" or "Multi-pass IRV."
 
+First Round Determines Threshold: Single-winner ranked-choice voting elections can use the first round to determine the threshold, rather than the last round. This will not affect the eventual winner. This may elect a winner that does not reach the threshold.
```
tarheel, line 25: maybe clarify that this is the winning threshold (just to be explicit)

tarheel (commented):
I think the only potential issue here is that when we end up with two candidates remaining and neither one meets the threshold, we're not actually supposed to generate a final round in which only the winning candidate remains; instead, the tabulation should behave the same way it does when the winner meets the threshold, i.e. they get elected and the tabulation stops without eliminating the runner-up. This has no impact on who wins, of course, but IIRC it's a nitpicky detail that's been flagged in the past by someone or other. And ditto if we need a tie-breaker to determine the winner at this point.

You can confirm with @chughes297 that my understanding here is accurate...

artoonie:
@tarheel to get us all on the same page --

# Current Behavior
image

# Your Suggestion
image

@chughes297 etc, can you chime in? Which of these should RCTab produce?

tarheel:
Except in the second image I would want Vail to have the green highlight in the last round to indicate that he won.

chughes297:
agree!

sent from my phone

chughes297:
> Except in the second image I would want Vail to have the green highlight in the last round to indicate that he won.

agree!

artoonie:
Done.

## Without tie-breaker
image


## With tie-breaker
image

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -482,19 +493,32 @@ private List<String> identifyWinners(
...
         }
       }
 
       // Edge case: if we've identified multiple winners in this round, but we're only supposed to
-      // elect one winner per round, pick the top vote-getter and defer the others to subsequent
-      // rounds.
-      if (config.isMultiSeatAllowOnlyOneWinnerPerRoundEnabled() && selectedWinners.size() > 1) {
+      // elect one winner per round, pick the top vote-getter. If this is a multi-winner election,
+      // defer the others to subsequent rounds. If this is a single-winner election where its
+      // possible  for no candidate to reach the threshold (i.e. "first round determines threshold"
```
tarheel, line 517: super nits: `s/where its/in which it's/` and `s/  / /`

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -482,19 +493,32 @@ private List<String> identifyWinners(
...
+          // Otherwise, select all winners above the threshold
           selectWinners(currentRoundTallyToCandidates, selectedWinners);
         }
       }
 
       // Edge case: if we've identified multiple winners in this round, but we're only supposed to
-      // elect one winner per round, pick the top vote-getter and defer the others to subsequent
-      // rounds.
-      if (config.isMultiSeatAllowOnlyOneWinnerPerRoundEnabled() && selectedWinners.size() > 1) {
+      // elect one winner per round, pick the top vote-getter. If this is a multi-winner election,
```
tarheel, line 515: Just for clarity, make this a bullet list:
```
Edge cases:
- (Case 1)
- (Case 2)
```

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -482,19 +493,32 @@ private List<String> identifyWinners(
...
+      // defer the others to subsequent rounds. If this is a single-winner election where its
+      // possible  for no candidate to reach the threshold (i.e. "first round determines threshold"
+      // is set), the tiebreaker will choose the only winner.
+      boolean useTiebreakerIfNeeded = config.isMultiSeatAllowOnlyOneWinnerPerRoundEnabled()
+          || config.isFirstRoundDeterminesThresholdEnabled();
```
tarheel, line 520: What about the case in which it's single-winner and `isFirstRoundDeterminesThresholdEnabled` is false but we end up with two candidates who both have exactly half of the votes?

If that's not already handled, we should add a test for it, and I would recommend tweaking this to something like:
```
boolean tiebreakNeeded = selectedWinners.size() > 1
  && (config.getNumberOfWinners() == 1 || config.isMultiSeatAllowOnlyOneWinnerPerRoundEnabled());

Comment (artoonie) [reply src/main/java/network/brightspots/rcv/Tabulator.java:520]: That would lead to one candidate being eliminated, so it wouldn't be handled here -- since nobody has reached the threshold, the tiebreaker runs at the elimination stage.

The test for this is in `tiebreak_seed_test`.

Comment (tarheel) [reply src/main/java/network/brightspots/rcv/Tabulator.java:522]: Ah, so I think that test is wrong! It eliminates the tiebreak loser and runs a final round with only the winner continuing. I think we want to make that behavior consistent with what you've done here. @chughes297 can confirm.

To be specific, here's what that test is currently expecting:
```
{
  "config" : {
    "contest" : "Tiebreak test",
    "date" : "2017-12-03",
    "jurisdiction" : "Funkytown, USA",
    "office" : "Sergeant-at-Arms",
    "threshold" : "2"
  },
  "results" : [ {
    "round" : 1,
    "tally" : {
      "George Gervin" : "3",
      "Mookie Blaylock" : "3",
      "Yinka Dare" : "3"
    },
    "tallyResults" : [ {
      "eliminated" : "Yinka Dare",
      "transfers" : {
        "exhausted" : "3"
      }
    } ]
  }, {
    "round" : 2,
    "tally" : {
      "George Gervin" : "3",
      "Mookie Blaylock" : "3"
    },
    "tallyResults" : [ {
      "eliminated" : "George Gervin",
      "transfers" : {
        "exhausted" : "3"
      }
    } ]
  }, {
    "round" : 3,
    "tally" : {
      "Mookie Blaylock" : "3"
    },
    "tallyResults" : [ {
      "elected" : "Mookie Blaylock",
      "transfers" : { }
    } ]
  } ]
}
```

And I think what we actually want is for round 2 to be the final round, and in that round no one is eliminated but we elect Blaylock via the tiebreaker.

chughes297: > And I think what we actually want is for round 2 to be the final round, and in that round no one is eliminated but we elect Blaylock via the tiebreaker.

Yes!

HEdingfield (changes requested):
I'll plan to give this a quick review on the UI stuff once @tarheel is satisfied with everything.

src/main/java/network/brightspots/rcv/Tabulator.java
artoonie: To keep this issue well-scoped, and since that's a bug that exists in `develop` today, I've created a separate issue to address it: https://github.com/BrightSpots/rcv/issues/694

tarheel (approved):

HEdingfield (approved):
Approved for UI changes.

PR #649, opened 2023-04-03
Add generic CSV reader

artoonie:


src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -239,17 +239,20 @@ && stringAlreadyInUseElsewhereInSource(
...
+              "firstVoteColumnIndex",
+              provider,
+              source.getFilePath())) {
+            validationErrors.add(ValidationError.CVR_FIRST_VOTE_UNEXPECTEDLY_DEFINED);
+          }
+
+          if (fieldIsDefinedButShouldNotBeForProvider(
+              source.getFirstVoteRowIndex(), "firstVoteRowIndex", provider, source.getFilePath())) {
+            validationErrors.add(ValidationError.CVR_FIRST_VOTE_ROW_UNEXPECTEDLY_DEFINED);
+          }
```
tarheel, line 255: Instead of modifying this section, can we just change the line earlier in this method from
```
} else if (provider == Provider.ESS) {
```
to
```
} else if (provider == Provider.ESS || provider == Provider.CSV) {
```
?

(This suggestion is related to my comments about shared functionality between ESS and CSV.)

src/main/java/network/brightspots/rcv/CsvCvrReader.java
```diff
@@ -0,0 +1,111 @@
...
+                        null,
+                        rankings);
+        castVoteRecords.add(newCvr);
+      }
+    } catch (IOException exception) {
+      Logger.severe("Error parsing cast vote record:\n%s", exception);
+      throw exception;
+    }
+  }
+}
```
tarheel, line 111: missing newline at end

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -1041,6 +1037,18 @@ public LocalDate fromString(String string) {
           checkBoxCvrTreatBlankAsUndeclaredWriteIn
               .setSelected(ContestConfig.SUGGESTED_TREAT_BLANK_AS_UNDECLARED_WRITE_IN);
         }
+        case CSV -> {
+          buttonAddCvrFile.setDisable(false);
+          textFieldCvrFilePath.setDisable(false);
+          buttonCvrFilePath.setDisable(false);
+          textFieldCvrContestId.setDisable(false);
```
tarheel, line 1044: How is this used? It doesn't look like the reader has any concept of a CVR with multiple contests in it, so I'm not sure why the user would supply a contest ID.

src/main/java/network/brightspots/rcv/StreamingCvrReader.java
```diff
@@ -57,11 +57,11 @@ class StreamingCvrReader {
   private final String excelFilePath;
   // name of the source file
   private final String excelFileName;
-  // 1-based column index of first ranking
+  // 0-based column index of first ranking
```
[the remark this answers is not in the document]
tarheel: Why do you say that? All of these are interpreted as if the first row and column have indexes of 1. You can see in the constructor below that we then subtract 1 from each of these values.

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -277,7 +277,12 @@ private List<CastVoteRecord> parseCastVoteRecords(ContestConfig config, Set<Stri
         } else if (ContestConfig.getProvider(source) == Provider.CLEAR_BALLOT) {
           Logger.info("Reading Clear Ballot cast vote records from file: %s...", cvrPath);
           new ClearBallotCvrReader(cvrPath, config, source.getUndeclaredWriteInLabel())
-              .readCastVoteRecords(castVoteRecords, source.getContestId());
+                  .readCastVoteRecords(castVoteRecords, source.getContestId());
```
tarheel, line 280: this indentation change seems wrong

src/test/resources/network/brightspots/rcv/test_data/generic_csv_test/generic_csv_test_cvr.csv
```diff
@@ -0,0 +1,27 @@
...
+R_Aiy5eILJk4S5tIt,2,1,3,4,
+R_2xDiLvKgBaIxeFv,2,3,4,5,1
+R_124lgkhlQv9Ullj,5,4,3,2,1
+R_21EC0Nti8aqKdm9,5,3,2,4,1
+R_1pDpfl9azbgtFrC,1,2,3,4,5
+R_3EXs2zblwxj5olU,4,3,1,2,
+R_1I4UVe24fpZcqPY,2,4,5,3,1
+R_27v7BD6blfCf7BQ,3,1,4,5,2
+R_sUm5EhzNgJ9Fy7f,2,1,3,5,4
+R_33dwoHsyajiIzbF,4,2,1,3,
```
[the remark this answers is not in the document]
tarheel: Could you clarify your question?

tarheel, line 27: missing newline

src/main/java/network/brightspots/rcv/CsvCvrReader.java
```diff
@@ -0,0 +1,111 @@
...
+      CvrSource source) {
+    this.cvrPath = cvrPath;
+    this.contestConfig = contestConfig;
+    this.contestId = source.getContestId();
+    this.undeclaredWriteInLabel = source.getUndeclaredWriteInLabel();
+    this.firstVoteColumnIndex = Integer.parseInt(source.getFirstVoteColumnIndex()) - 1;
+    this.firstVoteRowIndex = Integer.parseInt(source.getFirstVoteRowIndex()) - 1;
+  }
+
+  // parse Cvr xml file into CastVoteRecord objects and add them to the input List<CastVoteRecord>
```
tarheel, line 58: not XML

tarheel (commented):
I had assumed that when we introduced a generic CSV reader, we would factor out all of the non-file-format-specific parts of `StreamingCvrReader` into an abstract `SpreadsheetReader` or whatever and then we'd have an XLS reader (for ESS) and a generic CSV reader as child classes. With this new `CsvCvrReader` that you've created, it has a fair amount of redundancy with `StreamingCvrReader`, but it's also missing a number of features (e.g. CVR IDs, precincts, overvote labels, overvote delimiters). In theory I'd expect the two readers to have identical functionality and differ only in the respect that the ESS one has to deal with a bunch of annoying Excel-specific stuff.

src/main/java/network/brightspots/rcv/StreamingCvrReader.java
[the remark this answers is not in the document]
artoonie: The constructor is:
```
    // to keep our code simple, we convert 1-indexed user-supplied values to 0-indexed here
    this.firstVoteColumnIndex = Integer.parseInt(source.getFirstVoteColumnIndex()) - 1;
    this.firstVoteRowIndex = Integer.parseInt(source.getFirstVoteRowIndex()) - 1;
```

Meaning that `this.firstVoteRowIndex` is obtained by subtracting 1 from a 1-indexed value, making it zero-indexed...both according to the logic and the comments below. Right? Forgive me if I'm not seeing something obvious here.

src/test/resources/network/brightspots/rcv/test_data/generic_csv_test/generic_csv_test_cvr.csv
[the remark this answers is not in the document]
artoonie: How do we explain to users how to create this new CSV format? Do we point them to this example file and say "do it like this"? Is there an existing document somewhere for How To Use RCTab, and if so, does an explainer belong in there?

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: That would add a bunch of invalid checks for CSV (firstVoteColumnIndex, firstVoteRowIndex, idColumnIndex, precinctColumnIndex).

Note that the only change here is REMOVING this existing check from CSVs. If you click "Hide Whitespace" on github, that becomes more clear.

src/main/java/network/brightspots/rcv/StreamingCvrReader.java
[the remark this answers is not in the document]
tarheel: Ah, so I think the comments are just poorly phrased. They were intended to convey that on the config side, these indexes are 1-based, and then internally we convert them.

src/main/java/network/brightspots/rcv/ContestConfig.java
tarheel: But shouldn't CSV have those checks? Are we validating `firstVoteColumnIndex` and `firstVoteRowIndex` for CSV somewhere else?

(Same for `idColumnIndex` and `precinctColumnIndex` if you add support for those.)

tarheel (commented):
Oh, sorry, I'm realizing now that the format of the generic CSV is different from ESS: the columns are candidates, not ranks. But, as you note, this is also how the Hart reader operates, so maybe there's an opportunity for shared logic there?

There's also just the general concept of taking a spreadsheet file and running through it one line at a time. I don't know whether there's enough commonality there to warrant shared code.

Note that (if memory serves) we created `StreamingCvrReader` because the initial implementation of the ESS reader pulled the whole file into memory and ran into problems if the source was huge. But it looks like you're already doing that for your new reader, yeah?

I guess the concept of overvote delimiters and overvote labels aren't relevant to this CSV format. But CVR IDs and precinct IDs still could be, I think?

> How do we explain to users how to create this new CSV format? Do we point them to this example file and say "do it like this"? Is there an existing document somewhere for How To Use RCTab, and if so, does an explainer belong in there?

Oh, I guess we could add this to the `README` or something?

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: Ah, you're right -- those checks should be in here.

It would be nice to automate this a little more: the same information needs to be synced here, in `GuiConfigController.java`, and in the `Reader` constructor. I think we can consolidate a bit -- a task for later! For now, I'll add the checks here.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1164,6 +1167,7 @@ enum Provider {
     DOMINION("dominion", "Dominion"),
     ESS("ess", "ES&S"),
     HART("hart", "Hart"),
+    CSV("genericCsv", "CSV"),
```
HEdingfield, line 1170: It occurred to me that we could do at least some barebones documentation on what this means in `hints_cvr_files.txt` under the `Provider` entry (e.g. break out a little * bulleted list of each one with a brief description). Can't really think of a better spot offhand.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -523,6 +523,14 @@ public void buttonClearDatePickerContestDateClicked() {
     datePickerContestDate.setValue(null);
   }
 
+  private List<File> chooseFile(Provider provider, ExtensionFilter filter) {
+    FileChooser fc = new FileChooser();
+    fc.setInitialDirectory(new File(FileUtils.getUserDirectory()));
+    fc.getExtensionFilters().add(new ExtensionFilter("JSON and XML files", "*.json", "*.xml"));
```
HEdingfield, line 529: I believe you should be subbing `filter` in here instead.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -1041,11 +1037,23 @@ public LocalDate fromString(String string) {
...
+          textFieldCvrFirstVoteRow.setDisable(false);
+          textFieldCvrFirstVoteRow
+                  .setText(String.valueOf(ContestConfig.SUGGESTED_CVR_FIRST_VOTE_ROW));
+        }
         case CLEAR_BALLOT, DOMINION, HART -> {
           buttonAddCvrFile.setDisable(false);
           textFieldCvrFilePath.setDisable(false);
           buttonCvrFilePath.setDisable(false);
-          textFieldCvrContestId.setDisable(false);
+          textFieldCvrContestId.setDisable(true);
```
HEdingfield, line 1056: Curious what's motivating this change? Was this a bug in there before? If so, how did you come to that conclusion?

Finally, if it is the case that this should be changed, just remove this line instead since the default is for it to be disabled.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -1041,11 +1037,23 @@ public LocalDate fromString(String string) {
           checkBoxCvrTreatBlankAsUndeclaredWriteIn
               .setSelected(ContestConfig.SUGGESTED_TREAT_BLANK_AS_UNDECLARED_WRITE_IN);
         }
+        case CSV -> {
+          buttonAddCvrFile.setDisable(false);
+          textFieldCvrFilePath.setDisable(false);
+          buttonCvrFilePath.setDisable(false);
+          textFieldCvrContestId.setDisable(true);
```
HEdingfield, line 1044: It looks like you use `contentId` in `CsvCvrReader`, so you probably want this set to `false` instead.

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -279,6 +279,11 @@ private List<CastVoteRecord> parseCastVoteRecords(ContestConfig config, Set<Stri
           new ClearBallotCvrReader(cvrPath, config, source.getUndeclaredWriteInLabel())
               .readCastVoteRecords(castVoteRecords, source.getContestId());
           continue;
+        } else if (ContestConfig.getProvider(source) == Provider.CSV) {
+          Logger.info("Reading Generic CSV cast vote records from file: %s...", cvrPath);
```
HEdingfield, line 283: Nit: generic (lowercase)

src/main/java/network/brightspots/rcv/CsvCvrReader.java
```diff
@@ -0,0 +1,111 @@
+/*
+ * RCTab
+ * Copyright (c) 2017-2022 Bright Spots Developers.
```
HEdingfield, line 3: Realized that we'll need to update the copyright dates to 2023. There's an auto setting for this when committing files in IntelliJ "Update copyright", but maybe you're not using IntelliJ.

Maybe create a new required issue for this before we launch 1.4.0?

src/main/java/network/brightspots/rcv/CsvCvrReader.java
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * Purpose: Read and parse generic CSV files
```
HEdingfield, line 11: Nit: add period at end.

src/main/java/network/brightspots/rcv/CsvCvrReader.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVParser;
+import org.apache.commons.csv.CSVRecord;
+
+class CsvCvrReader {
+  private final String cvrPath;
+  private final String contestId;
+  private final String undeclaredWriteInLabel;
+  private final ContestConfig contestConfig;
+  // 0-based column index of first ranking
```
HEdingfield, line 39: If you agree with what @tarheel is saying in the other thread, please update these comments here too.

src/main/java/network/brightspots/rcv/CsvCvrReader.java
```diff
@@ -0,0 +1,111 @@
...
+            Logger.severe("Row %s expected number at column %d, but got \"%s\" instead.",
+                    csvRecord.get(0), col, rankAsString);
+            throw new CastVoteRecord.CvrParseException();
+          }
+
+          String candidateId = candidateIds.get(col);
+          if (col == undeclaredWriteInColumn) {
+            candidateId = Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL;
+          } else if (!contestConfig.getCandidateCodeList().contains(candidateId)) {
+            unrecognizedCandidateCounts.merge(candidateId, 1, Integer::sum);
```
HEdingfield, line 91: I think you need to check this at the end and throw an error if it's > 0, as is done in `ClearBallotCvrReader` and the others as well.

HEdingfield (changes requested):

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: Good catch -- this was a mistake in the latest commit, @tarheel requested changing this a few lines above. Will revert.

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: I added a ticket to address this, since with the addition of the CSV checks, the function gets pretty hairy https://github.com/BrightSpots/rcv/issues/667

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: I'm not sure how to briefly describe these other than e.g. "Dominion: a Dominion CVR File". CSV could use better documentation, but I'm not sure what to write from the others...

Thoughts on just linking to example files for each? I think that may be more useful that trying to document it?

src/main/java/network/brightspots/rcv/CsvCvrReader.java
artoonie: Should we update all files, or only files as they're edited? I can find that intellij setting -- I'm using vim currently but I do have access to intellij.

src/main/java/network/brightspots/rcv/CsvCvrReader.java
artoonie: @HEdingfield @tarheel I think the comment is more confusing if we say "1-indexed" -- in my experience, comments above members refer to the members themselves, not their origins or how they were derived.

Thoughts on modifying this so everybody is happy? e.g. `0-based column index of the first ranking, derived by subtracting 1 from the 1-based config variable`? That feels quite long. Or, `1 less than the 1-based column index of the first ranking`? That feels kind of confusing, but is shorter.

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: It was set but had no impact on the output files, so @tarheel noted it should be removed. I removed all references to it from `CsvCvrReader` too.

src/main/java/network/brightspots/rcv/CsvCvrReader.java
```diff
@@ -0,0 +1,114 @@
...
+                new CastVoteRecord(
+                        "no contest ID",
+                        "no supplied ID",
+                        "no precinct",
+                        null,
+                        rankings);
+        castVoteRecords.add(newCvr);
+      }
+
+      if (unrecognizedCandidateCounts.size() > 0) {
```
HEdingfield, line 106: Nit: to keep this consistent with ClearBallot's (and maybe a touch more readable), you could move this to around line 112, after the `catch` for `IOException`.

src/main/resources/network/brightspots/rcv/hints_cvr_files.txt
```diff
@@ -4,7 +4,12 @@ The tabulator needs to know where each of your CVR files is and how to interpret
...
 
-Provider (required): The vendor/machine that generated (produced) the file. After you select the field, the tabulator will fill in as many of the other fields as it can based on what it knows about that provider. You can adjust those values as necessary.
+Provider (required): The vendor/machine that generated (produced) the file. After you select the field, the tabulator will fill in as many of the other fields as it can based on what it knows about that provider. You can adjust those values as necessary. The options are:
+  * Dominion. Example: https://github.com/BrightSpots/rcv/tree/develop/src/test/resources/network/brightspots/rcv/test_data/dominion_alaska/
+  * ES&S. Example: https://github.com/BrightSpots/rcv/blob/develop/src/test/resources/network/brightspots/rcv/test_data/aliases_ess_xlsx/
+  * Hart. Example: https://github.com/BrightSpots/rcv/tree/develop/src/test/resources/network/brightspots/rcv/test_data/hart_cedar_park_school_board/
+  * CSV. Example: https://github.com/BrightSpots/rcv/tree/develop/src/test/resources/network/brightspots/rcv/test_data/generic_csv_test/
```
HEdingfield, line 11: See notes from other conversation related to this -- I think we should remove these long example links and either replace with a single link to the Wiki at the bottom or, maybe better yet (since we won't have to worry about keeping it up to date), not link to examples at all.

Also, it seems like you're missing Clear Ballot here, so please be comprehensive even if it's just adding the word alone with the provider name (and maybe just CSV has a bit of an explanation after it per the other thread).

HEdingfield (commented):
Just about ready!

src/main/java/network/brightspots/rcv/CsvCvrReader.java
artoonie: I think it is, github is showing the "outdated" code here -- the "Files Changed" will show it truly resolved :)

src/main/java/network/brightspots/rcv/CsvCvrReader.java
HEdingfield: Created #668 for this.

src/main/resources/network/brightspots/rcv/hints_cvr_files.txt
HEdingfield: Yeah, I think that's great, thanks for setting that up! Only caveat is that the Qualtrics comment might not apply yet because you have to enable the checkbox that inverts the CSV format, right? Fine to leave as-is if you intend to add that feature as part of 1.4.0.

HEdingfield (approved):
LGTM, nice work on all this!

src/main/java/network/brightspots/rcv/CsvCvrReader.java
tarheel: Given how much confusion this has already generated, I think including the long version is well worth it.

artoonie:
Hi @brohithkr this info is available in the documentation here: https://rctab-docs.readthedocs.io/en/latest/user_guide/user_guide.html#cvr-files-tab

Under "CSV CVR".

Please let me know if you have any feedback on that documentation!

PR #652, opened 2023-04-06
Issue 60: Auto-Load candidates

artoonie:
image

The logic here is perhaps a little too pythonic. Would love advice on whether using exceptions in this way is acceptable.

What we do is use the existing `Read CVR` logic, and catch all `UnrecognizedCandidatesException`s to find any as-of-yet-unloaded candidate name. I believe it's frowned upon to use Exceptions for non-error-handling in Java, but this implementation felt like the cleanest way to go about it. Thoughts?


src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
```diff
@@ -312,55 +313,84 @@
...
+                  </VBox>
+                </VBox>
+                <VBox styleClass="bordered-box" maxWidth="400.0">
+                  <VBox alignment="TOP_CENTER">
+                    <padding>
+                      <Insets left="8.0" right="8.0" bottom="8.0" top="8.0"/>
+                    </padding>
+                    <Label text="Manually Add Candidate" style="-fx-font-weight: bold"/>
+                    <Text
+                      text="You can manually add candidates if auto-loading fails to produce acceptable results."
```
tarheel, line 346: Under what circumstances could auto-loading not produce the desired outcome? Is that specifically the case in which they want to map multiple strings (using aliases, once we've killed codes) to the same candidate? I guess ideally we would support that somehow, but obviously that's another layer of UX complexity on top of what you've added here.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -747,6 +749,41 @@ public void buttonDeleteCvrFileClicked() {
...
+    int successCount = 0;
+    for (String name : unloadedNames) {
+      Candidate candidate = new Candidate(name, "", false);
+      Set<ValidationError> validationErrors =
+              ContestConfig.performBasicCandidateValidation(candidate);
+      if (validationErrors.isEmpty()) {
+        tableViewCandidates.getItems().add(candidate);
+        successCount++;
+      } else {
+        Logger.warning("Candidate %s failed to load.", name);
```
tarheel, line 780: So this would only happen if the name is the empty string? Would the parser not have already choked on that?

src/main/java/network/brightspots/rcv/BaseCvrReader.java
```diff
@@ -32,6 +34,16 @@ abstract class BaseCvrReader {
     this.cvrPath = config.resolveConfigPath(source.getFilePath());
   }
 
+  // Gather candidate names from the CVR that are not in the config.
+  Set<String> gatherUnknownCandidateNames() throws CastVoteRecord.CvrParseException, IOException {
+    try {
+      readCastVoteRecords(new ArrayList<>(), new HashSet<>());
+    } catch (TabulatorSession.UnrecognizedCandidatesException unrecognizedCandidates) {
```
tarheel, line 41: Hmm. This is clever! But yeah, I think you're probably right that it's not really a kosher way to use exceptions. Could you just access the list of unrecognized candidates after calling `readCastVoteRecords` without using exception handling? Or maybe just have the parser expose a new method `getUnrecognizedCandidates` or whatever.

src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
artoonie: Yes, and perhaps cases where you want non-declared candidates to be listed as Write-Ins, or if `UnrecognizedCandidatesException` isn't implemented in some readers (which it's not required to be).

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: Not sure I follow -- this is checking if `validationErrors` is empty, not the `name`. I don't know if there are any checks preventing an empty-string candidate name -- perhaps it's an acceptable, unique name?

src/main/java/network/brightspots/rcv/BaseCvrReader.java
artoonie: It may be difficult to handle many types of errors in that case in a way that's resilient to future errors being added?

```
try {
  readCastVoteRecords(...)
} catch(ExceptionType0 e) {
} catch(ExceptionType1 e) {
} catch(ExceptionType2 e) {
}
```

to

```
readCastVoteRecords(...)
getErrorsType0(..)
getErrorsType1(..)
getErrorsType2(..)
```

Or, have all possible errors inherit from some base class, and return a list of that class' type?

A third option that feels the cleanest, but is perhaps the most work, is to create a `CVRParseResults` class:

```
class CVRParseResults() {
  List<CastVoteRecord> CastVoteRecords;
  bool HasErrors;
  Map<String, Integer> UnrecognizedCandidateCounts;
  ...
}
```

And rather than throwing any exceptions, we just check `HasErrors` here.

src/main/java/network/brightspots/rcv/GuiConfigController.java
tarheel: See `performBasicCandidateValidation`

artoonie: Ah, I see -- you're right, that's the only thing checked in `performBasicCandidateValidation`. The parsers do not check for it, best I can see.

It still seems good to check for this here in case `performBasicCandidateValidation` adds additional checks in the future, just so the auto- and manual- candidate loader use the same validation step.

HEdingfield (changes requested):
Placing a hold on this until we get the upstream PRs merged first. I'll review once those are in `develop`.

artoonie:
I've refactored this -- it's a more significant change, but I do think this is nicer.

It now separates out:
1. Load as much as possible, only erroring if we really can't load
2. Validate everything

The validation now has the added benefit of happening in the base class, so now all readers run the same validation code. There is also the opportunity for custom validation, which the dominion reader takes advantage of.

HEdingfield:
Nice, I'm going to try to get around to this this weekend.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -747,6 +749,43 @@ public void buttonDeleteCvrFileClicked() {
         .removeAll(tableViewCvrFiles.getSelectionModel().getSelectedItems());
   }
 
+  /**
+   * Action when "Auto-Load Candidates" clicked.
+   */
+  public void buttonAutoLoadCandidatesClicked() {
```
HEdingfield, line 747: These are some pretty hardcore calculations that can take a while, so you're going to need to `setGuiIsBusy` for this, and likely create a new service to handle it. See `menuItemValidateClicked` for an example.

Otherwise, users can spam the button and cause things to get locked up pretty badly.

HEdingfield:
Also submitted a commit with some clean-up, shout out if anything looks bad to you in there.

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: Neat! Didn't see that before. Done.

src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -194,12 +195,45 @@ private void gatherCvrsForContest(List<CastVoteRecord> castVoteRecords, String c
...
+      Set<String> candidates = contestIdToCandidateNames.get(cvr.getContestId());
+
+      for (Pair<Integer, CandidatesAtRanking> ranking : cvr.candidateRankings) {
+        for (String candidateId : ranking.getValue()) {
+          String candidateName = config.getNameForCandidate(candidateId);
+          if (candidateId.equals(source.getOvervoteLabel())) {
+            continue;
+          }
+          if (candidateId.equals(Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL)) {
+            // Note: candidateId is replaced with Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL
```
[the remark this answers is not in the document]
artoonie: The previous check should be removed entirely, actually -- dominion's reader doesn't support an overvote label.

HEdingfield (approved):
Looks great! Just added a little more logging.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -1388,47 +1388,51 @@ private static class AutoLoadCandidatesService extends Service<Void> {
...
-
-                  Logger.info("Auto-loaded %d candidates.", successCount);
-                  return null;
+          new Task<>() {
+            @Override
+            protected Void call() {
+              Logger.info("Auto-loading candidates from CVR files...");
+              if (sources.isEmpty()) {
+                Logger.warning("No CVR files specified!");
+                return null;
```
artoonie, line 1397: Is there an issue with multiple returns here @HEdingfield?

tarheel: Yeah, this isn't allowed. Also line 1433 below.

HEdingfield: Good catch. Soooo dumb. I'll clean this up. 1433 is necessary though, because we're overriding the `call` method so it needs some kind of return.

tarheel: Oops, right. Hard to read indentation when viewing in GitHub.

HEdingfield: Filed #693 for this.

Issue #282, opened 2019-06-10
Figure out how to create distributions for Windows, Linux, and MacOS on any system


HEdingfield:
Versions created using Gradle's `distZip` or `jlink` tasks are platform-specific due to the way JavaFX libraries are downloaded. If you try to run a version built on a Mac in Windows or Linux, for example, you get:

> Graphics Device initialization failed for :  d3d, sw
> Error initializing QuantumRenderer: no suitable pipeline found
> java.lang.RuntimeException: java.lang.RuntimeException: Error initializing QuantumRenderer: no suitable pipeline found

It should be possible to set up a separate zip dist for each OS even without actually being on that OS by doing something synthesized from the following:

- As outlined in this awesome post: https://stackoverflow.com/a/54065502/3846321 ...
- We should be able to use the `targetPlatform` method for `jlink` here: https://badass-jlink-plugin.beryx.org/releases/latest/#_methods

See also:
- https://stackoverflow.com/questions/53757807/build-java-module-runtime-image-for-other-os
- Creating separate Run/Debug configurations in IntelliJ and feeding in different arguments for OS, then following something like this to actually build them and pull the correct JavaFX libs: https://stackoverflow.com/questions/52569724/javafx-11-create-a-jar-file-with-gradle
- https://stackoverflow.com/questions/52584519/does-javafx-after-removing-it-from-jdk-cross-platform
- https://docs.gradle.org/4.10.3/userguide/application_plugin.html#sec:customizing_start_script_generation
- https://docs.gradle.org/4.10.3/userguide/distribution_plugin.html

HEdingfield:
Addressing #57 using `jpackage` may make this a moot point.

HEdingfield:
Attempted to use `targetPlatform` to generate for all three platforms after downloading the JDKs for each of them (see `build.gradle` excerpt I used below). The MacOS version didn't work when @tarheel tried. The one I generated on Linux gave:

> ~/Downloads/rcv-macOS/bin $ ./java
> -bash: ./java: cannot execute binary file

The one I generated on Windows gave:

> ~/Downloads/rcv-macOS $ bin/rcv
> bin/rcv: line 3: bin/java: cannot execute binary file

[...]

PR #658, opened 2023-04-22
Auto-build releases (#282)

artoonie:
Github action to:
1. Build assets on every merge into develop and master
2. Automatically add assets to every prerelease and release

Where assets include the zip files for windows, mac, and linux, built with jdk 17.0.2, along with their SHA512s.

The zips go in two places:
1. On every push to master or develop, they will show up [in the "artifacts" section](https://github.com/BrightSpots/rcv/actions/runs/4773447699) in the Action linked from the PR.
2. On every prerelease and release, they will show up [in the Releases section](https://github.com/BrightSpots/rcv/releases)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,12 @@
 name: Build Releases
 
-on: [push]
-# TODO: eventually, only build on master, and add a step to deploy when a Release is created.
+on:
+  release:
+    types: [published]
+  push:
+    branches:
+      - 'master'
+      - 'develop'
 
 jobs:
   release:
@@ -10,41 +15,38 @@ jobs:
         matrix:
             os: [ubuntu-latest, windows-latest, macos-latest]
     steps:
-      #- name: Create filename
-      #  id: fn
-      #  run: echo "FILENAME=build/rctab_${{ github.ref_name }}_${{ runner.os }}.zip" >> $GITHUB_OUTPUT
-      - name: Create filename
+      - name: "Create filename"
         id: fn
         shell: bash
-        run: echo "FILENAME=build/rctab_v1.3.1_${{ runner.os }}.zip" >> $GITHUB_OUTPUT
+        run: echo "FILENAME=build/rctab_${{ github.ref_name }}_${{ runner.os }}.zip" >> $GITHUB_OUTPUT
           
-
       - uses: actions/checkout@v3
-      - name: Set up JDK 17.0.2
+
+      - name: "Set up JDK 17.0.2"
         uses: actions/setup-java@v3
         with:
           java-version: '17.0.2'
           distribution: 'temurin'
 
-      - name: Validate Gradle wrapper
+      - name: "Validate Gradle wrapper"
         uses: gradle/wrapper-validation-action@ccb4328a959376b642e027874838f60f8e596de3
 
-      - name: Create zip with Gradle
+      - name: "Create zip with Gradle"
         uses: gradle/gradle-build-action@67421db6bd0bf253fb4bd25b31ebb98943c375e1
         with:
           arguments: jlinkZip
 
-      - name: "Copy zip file"
+      - name: "Rename zip file"
         run: cp build/rcv.zip ${{ steps.fn.outputs.FILENAME }}
 
-      - name: Linux SHA512
+      - name: "Generate SHA512 for Linux"
         if: runner.os == 'Linux'
         run: sha512sum ${{ steps.fn.outputs.FILENAME }} > ${{ steps.fn.outputs.FILENAME }}.sha512
-      - name: Windows SHA512
+      - name: "Generate SHA512 for Windows"
         if: runner.os == 'Windows'
         shell: bash
         run: certutil -hashfile ${{ steps.fn.outputs.FILENAME }} SHA512 >> ${{ steps.fn.outputs.FILENAME }}.sha512
-      - name: OSX SHA512
+      - name: "Generate SHA512 for OSX"
         if: runner.os == 'macOS'
         run: openssl dgst -sha512 > ${{ steps.fn.outputs.FILENAME }}.sha512
 
@@ -55,3 +57,14 @@ jobs:
           path: |
             ${{ github.workspace }}/${{ steps.fn.outputs.FILENAME }}
             ${{ github.workspace }}/${{ steps.fn.outputs.FILENAME }}.sha512
+          retention-days: 90
+
+      - name: "Upload binaries to release"
+        uses: svenstaro/upload-release-action@v2
+        if: ${{ github.event_name }} == 'release'
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: ${{ github.workspace }}/${{ steps.fn.outputs.FILENAME }}*
+          tag: ${{ github.ref_name }}
+          overwrite: true
+          file_glob: true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,57 @@
+name: Build Releases
+
+on: [push]
+# TODO: eventually, only build on master, and add a step to deploy when a Release is created.
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
+name: Build Releases
+
+on:
+  release:
+    types: [published]
```
HEdingfield, line 5: This means it'll automatically trigger when we save a draft of a new release? Or do we specifically have to choose the "pre-release" box?

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+  push:
+    branches:
+      - 'master'
```
HEdingfield, line 8: How does it know to build the v1.3.0 or whatever tag in master?

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+      - 'develop'
```
HEdingfield, line 9: What is develop doing in here? Sorry I was lazy and didn't look up the docs around this...

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17.0.2'
+          distribution: 'temurin'
+
+      - name: "Validate Gradle wrapper"
+        uses: gradle/wrapper-validation-action@ccb4328a959376b642e027874838f60f8e596de3
+
+      - name: "Create zip with Gradle"
+        uses: gradle/gradle-build-action@67421db6bd0bf253fb4bd25b31ebb98943c375e1
```
HEdingfield, line 35: Where'd you pull this hash from? Is there a list somewhere to choose from?

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+        shell: bash
+        run: echo "FILENAME=build/rctab_${{ github.ref_name }}_${{ runner.os }}.zip" >> $GITHUB_OUTPUT
+          
+      - uses: actions/checkout@v3
+
+      - name: "Set up JDK 17.0.2"
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17.0.2'
+          distribution: 'temurin'
```
HEdingfield, line 29: When we update the JDK, is it as simple as incrementing it here and GH will take care of that? Also, do we ever have to change this `temurin` thing?

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+
+      - name: "Validate Gradle wrapper"
+        uses: gradle/wrapper-validation-action@ccb4328a959376b642e027874838f60f8e596de3
```
HEdingfield, line 32: I assume it'll stop the auto-build if this fails? Will we get emailed a notification or how exactly does that work?

HEdingfield (commented):
This is potentially mindblowingly cool! Thanks so much for putting the work in to get this going!

.github/workflows/release.yml
artoonie: See the third info box here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

```Note: The prereleased type will not trigger for pre-releases published from draft releases, but the published type will trigger. If you want a workflow to run when stable and pre-releases publish, subscribe to published instead of released and prereleased.```

So this will trigger when you draft a release or a prerelease, but not a draft. That's intentional, because drafts would have access to all the same information via the branch tags, just not uploaded to the "Release" section.

I'll clarify this in the main PR comment.

.github/workflows/release.yml
artoonie: 1. If the trigger is a push to master or develop, the tag will be the branch name
2. If the trigger is a release, the tag will be the tag name, which is equal to the version number

.github/workflows/release.yml
artoonie: See the updated main comment above. This is Option (1), creating Assets, which won't go into the Releases section.

We can remove this, but it's designed to allow non-devs to quickly test the latest Develop or Master without needing to compile code.

.github/workflows/release.yml
artoonie: From the Getting Started demos from github: https://github.com/actions/starter-workflows/blob/main/ci/gradle-publish.yml

The list would be any commit here: https://github.com/gradle/gradle-build-action/commits/main

.github/workflows/release.yml
artoonie: 1. Yep, we just increment it here
2. Not sure about temurin

.github/workflows/release.yml
artoonie: 1. The release creator will get an email notification if it failed while creating a release
2. It will mark the branch as having a failed build + we'll get an email notification if it happens after a merge to master/develop, like so:
 
image

.github/workflows/release.yml
HEdingfield: I think it's a cool idea, but let's remove it for now and add it back in in the future if there's ever any demonstrated need for it.

.github/workflows/release.yml
HEdingfield: Cool, ok. The only thing I'm confused about is how you supposedly created a [`test-auto-release`](https://github.com/BrightSpots/rcv/releases/tag/test-auto-release) tag, but I don't see it in any of the branches when I run `git tag`. Know why this is? Does it hold off on pushing the tag until the non-pre-release is published?

.github/workflows/release.yml
HEdingfield: Apparently temurin is an anagram of "runtime", and it's associated with the Eclipse version of the JDK. Interesting. I guess we just keep this as-is.

.github/workflows/release.yml
HEdingfield: Perfect, that sounds good.

.github/workflows/release.yml
HEdingfield: Interesting, ok. Is this another thing we'd need to update regularly (i.e. before each release?).

.github/workflows/release.yml
artoonie: I created several test tags and then deleted them, so `test-auto-release` was a tag I created and deleted while debugging the workflow

.github/workflows/release.yml
artoonie: If I could add another reason to keep this: (1) it runs faster than the tests, and runs in parallel, so there are no downsides, and (2) it allows regular testing of this workflow, instead of only running the workflow on releases.

I'm a little afraid it will become stale if we only run it on release creation.

If you still think I should remove it, I'm happy to.

.github/workflows/release.yml
artoonie: I don't think so -- perhaps if gradle has major security updates?

We can also have it use the latest commit, but locking the version is more secure and more reliable.

.github/workflows/release.yml
artoonie: Some thoughts:
1. I could skip the artifact upload if that's an issue
2. Or, decrease the lifespan to 5 days

But, for public repos, [my understanding](https://github.com/orgs/community/discussions/26438) is that there is no storage limit for Github Actions.

Again, no strong opinions here, but given that it's free in terms of both money and time, I don't see downsides.

tarheel: If it's totally free, sure.

HEdingfield: Appreciate that, @artoonie. Sounds good re: the draft method.

@tarheel haha I knew that was coming! Probably relatively low. In all seriousness though, my main motivation for working on this project is the hope that it will one day unblock solutions to the climate crisis. I was also brain damaged by doing lots of LCA work in the past, so these are the things that keep me up at night.

HEdingfield (approved):
Awesome job on this, thanks so much for all the research that went into it!

Issue #57, opened 2018-04-25
Create installer / standalone executable


HEdingfield:
Determine the workflow needed to create a platform-independent installer and/or standalone executable for each serious release.

gngilbert:
This would be very good to have but I have no idea how much work is involved.

HEdingfield:
Now that we use Gradle, we can look into using `jpackage` to create separate installers for each platform.

If we can do this, it will probably effectively fix #282 also.

rrojas350:
Originally removed to make room for other issues but this request has popped back up. Updating as a lower priority item and a nice-to-have if possible.

HEdingfield:
@artoonie was able to close this after years with #660! 🥳

yezr:
trying to see if i can associate the PR if it is open

Issue #625, opened 2023-02-05
XML parsing fails when running built version


tarheel:
If you try to tabulate an XML file, you get this:
```
javax.xml.stream.FactoryConfigurationError: Provider for javax.xml.stream.XMLInputFactory cannot be found
```

This doesn't happen if you run directly from IntelliJ.

tarheel:
Btw, when I open Project Structure, the Problems section has a ton of issues listed:
```

Library Gradle: com.fasterxml.jackson.core:jackson-annotations:2.13.3 has broken sources paths:   /Users/tarheel/.gradle/caches/modules-2/files-2.1/com.fasterxml.jackson.core/jackson-annotations/2.13.3/ba85620155c9143d2f9e731008c94cf778ffe91e/jackson-annotations-2.13.3-sources.jar [Fix]
Library Gradle: com.fasterxml.jackson.core:jackson-core:2.13.3 has broken sources paths:   /Users/tarheel/.gradle/caches/modules-2/files-2.1/com.fasterxml.jackson.core/jackson-core/2.13.3/f86da50dc26b8f89ed97197be1e106589b6aa19d/jackson-core-2.13.3-sources.jar [Fix]
Library Gradle: com.fasterxml.jackson.core:jackson-databind:2.13.3 has broken sources paths:   /Users/tarheel/.gradle/caches/modules-2/files-2.1/com.fasterxml.jackson.core/jackson-databind/2.13.3/87fe580a4141d26b6036234fb0c240aa1d4592db/jackson-databind-2.13.3-sources.jar [Fix]
Library Gradle: com.fasterxml.jackson.dataformat:jackson-dataformat-xml:2.13.3 has broken sources paths:   /Users/tarheel/.gradle/caches/modules-2/files-2.1/com.fasterxml.jackson.dataformat/jackson-dataformat-xml/2.13.3/d58bebe4785dd2efd90e418416f7e40173d25c99/jackson-dataformat-xml-2.13.3-sources.jar [Fix]
Library Gradle: com.fasterxml.woodstox:woodstox-core:6.2.7 has broken sources paths:   /Users/tarheel/.gradle/caches/modules-2/files-2.1/com.fasterxml.woodstox/woodstox-core/6.2.7/2b5a0c908f0c485908bbc4f4d413126aa3dd9796/woodstox-core-6.2.7-sources.jar [Fix]
[...]

PR #660, opened 2023-04-25
Standalone Executable

artoonie:
Auto-create executables using jpackage.

Issues:
1. The version number is now in three different places. We need an easy way to sync this across the project, but I couldn't find a straightforward way yet. We'll need a ticket to track this.
2. We're now creating a ton of packages: the launchable binaries (jpackage), the java runtime image (jlinkZip), SHAs for each, x3 platforms --> the packages are around 350mb. When a release is created, this won't be an issue since people can just download the relevant file, but we should consider whether we want to support the jlinkZip image indefinitely, or if the jpackage will supercede it


HEdingfield:
Curious why we're merging this into `master` instead of `develop`? Just so we can test it out sooner?

> MacOS packages aren't signed, and seem impossible to install on newer macs. They are painful to install on older macs. We probably need to [sign the packages](https://docs.github.com/en/actions/deployment/deploying-xcode-applications/installing-an-apple-certificate-on-macos-runners-for-xcode-development), which I believe require paying Apple $100/year?

My knee-jerk reaction to this is helllllllll no. But I guess we can see what @chughes297 thinks.

Also, I'm happy to test the bins for Linux and Windows if you can point me to where they are.

> The version number is now in three different places. We need an easy way to sync this across the project, but I couldn't find a straightforward way yet. We'll need a ticket to track this.

We should be able to get rid of it in `release.yml`, right, defaulting to the GitHub tag variable instead? Looks like there are some merge conflicts to resolve with `release.yml` anyway.

> We're now creating a ton of packages: the launchable binaries (jpackage), the java runtime image (jlinkZip), SHAs for each, x3 platforms --> the packages are around 350mb. When a release is created, this won't be an issue since people can just download the relevant file, but we should consider whether we want to support the jlinkZip image indefinitely, or if the jpackage will supercede it

Yeah, I think jpackage will replace jlinkZip. However, we should discuss the ramifications of this with @chughes297 to make sure it doesn't cause any additional security issues having them use a compiled version. Also wondering if we need to provide SHA512 checksums for the bins too.

chughes297:
>> MacOS packages aren't signed, and seem impossible to install on newer macs. They are painful to install on older macs. We probably need to [sign the packages](https://docs.github.com/en/actions/deployment/deploying-xcode-applications/installing-an-apple-certificate-on-macos-runners-for-xcode-development), which I believe require paying Apple $100/year?
> 
> My knee-jerk reaction to this is helllllllll no. But I guess we can see what @chughes297 thinks.
> 
Why hell no? I need to confirm this but I assume RCVRC can cover those costs, especially if it makes install that much easier. 
> Also, I'm happy to test the bins for Linux and Windows if you can point me to where they are.
> 
>> We're now creating a ton of packages: the launchable binaries (jpackage), the java runtime image (jlinkZip), SHAs for each, x3 platforms --> the packages are around 350mb. When a release is created, this won't be an issue since people can just download the relevant file, but we should consider whether we want to support the jlinkZip image indefinitely, or if the jpackage will supercede it
> 
> Yeah, I think jpackage will replace jlinkZip. However, we should discuss the ramifications of this with @chughes297 to make sure it doesn't cause any additional security issues having them use a compiled version. Also wondering if we need to provide SHA512 checksums for the bins too.

I need a bit more info about what is being discussed - not sure I understand what the potential security issues are if using a compiled version? And yes I believe we'll need to continue providing checksums for each release (if that's what's being asked lol).

src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -26,6 +26,7 @@
 @SuppressWarnings("WeakerAccess")
 public class Main extends GuiApplication {
 
+  // TODO - sync these values with build.gradle
```
HEdingfield, line 29: Any ideas on how we can create a common spot to access these vars for both `Main.java` and `build.gradle`?

I don't think it's a deal-breaker if we can't, as I typically just Ctrl-Shift-F in IntelliJ and search for the current version number any time I want to update it (which would catch both spots). Probably safe to get rid of both TODOs in that case, and maybe create an issue for the backlog if you think it's worth it.

build.gradle
```diff
@@ -98,19 +98,32 @@ javafx {
 
 def JLINK_DIR = "$buildDir/rcv"
 // Necessary to explicitly include due to https://github.com/BrightSpots/rcv/issues/625
-def MODULES_TO_ADD = "com.ctc.wstx"
+def MODULE_TO_ADD = "com.ctc.wstx"
```
HEdingfield, line 101: Nit: I lean towards keeping this as-is since it matches the option `add-modules`, and it also hints that we can add more if necessary.

build.gradle
```diff
@@ -98,19 +98,32 @@ javafx {
...
-        name = "rcv"
+        // TODO sync these values with Main.java
+        name = "RCTab"
+        version = "1.3.1"
+    }
+
+    jpackage {
+        installerOptions += [
+          '--verbose',
+          '--vendor', 'The Ranked Choice Voting Resource Center'
```
HEdingfield, line 122: Nit: I think we can get rid of "The" in here.

.github/workflows/release.yml
```diff
@@ -0,0 +1,103 @@
+name: Build Releases
```
HEdingfield, line 1: Looks like merge conflicts in here -- I'll look it over once those are resolved.

.github/actions/gradle-and-sha/action.yml
```diff
@@ -0,0 +1,36 @@
+name: Run gradle command and create SHA512
```
HEdingfield, line 1: Nice refactor!

.github/actions/gradle-and-sha/action.yml
```diff
@@ -0,0 +1,36 @@
...
+    type: string
+
+runs:
+  using: "composite"
+  steps:
+    - name: "Run gradle command"
+      uses: gradle/gradle-build-action@67421db6bd0bf253fb4bd25b31ebb98943c375e1
+      with:
+        arguments: ${{ inputs.gradle-command }}
+    - name: "Rename zip file"
```
HEdingfield, line 21: Nit: Rename output file

.github/actions/gradle-and-sha/action.yml
```diff
@@ -0,0 +1,36 @@
...
+      shell: bash
+      run: ls build/jpackage & mv ${{ inputs.intermediate-filepath }} ${{ inputs.final-filepath }}
```
HEdingfield, line 23: Is `jpackage` supposed to be explicitly referenced in this refactored action?

HEdingfield:
> @chughes297: Why hell no? I need to confirm this but I assume RCVRC can cover those costs, especially if it makes install that much easier.

I guess because I'm allergic to costs associated with open source software (especially recurring ones) but, more importantly, I thought all the election tabulation stations ran Windows anyway?

> @chughes297: I need a bit more info about what is being discussed - not sure I understand what the potential security issues are if using a compiled version? And yes I believe we'll need to continue providing checksums for each release (if that's what's being asked lol).

I guess I was thinking because it's just an extra level of opacity, which probably also creates a registry entry on their machine (correct me if I'm wrong, @artoonie).

src/main/java/network/brightspots/rcv/Main.java
artoonie: I think having a standalone `VERSION` or `version.txt` would be sufficient, but I need to read more about best practices in Java/Gradle to make that call.

My plan was to (1) create a ticket, (2) add the ticket number to the TODO. Whenever I need to copy/paste changes across a codebase, I like to also copy/paste a comment nearby that tells me all the places it needs to be synced.

chughes297:
> > @chughes297: Why hell no? I need to confirm this but I assume RCVRC can cover those costs, especially if it makes install that much easier.
> 
> I guess because I'm allergic to costs associated with open source software (especially recurring ones) but, more importantly, I thought all the election tabulation stations ran Windows anyway?
> 
Most do. So far in my experience, though, some offices have a Linux machine handy and others may have a Mac workstation they can use. I lean towards being as comprehensive as possible with OSes so we can be flexible and jurisdictions can ideally use a piece of hardware they already have sitting around instead of having to buy a new one. 
>>I guess I was thinking because it's just an extra level of opacity, which probably also creates a registry entry on their machine (correct me if I'm wrong, @artoonie).
>
>Yes, that is true -- when installing the software, you cede some control to the operating system which could be seen as reducing transparency. For example: If you try installing 1.3.0 but you previously had 1.4.0 installed, will your operating system decline to downgrade, but show you that it succeeded? The answer is No today, but that's the type of thing we lose control over.
>
>(This may not be the best example of what can go wrong, but it's the only thing I can think of right now.)
>
>However, assuming other election software follows the same model (which I expect they do), I think we can feel safe following the same best practices?
I'm curious to hear more about the security/opacity argument (maybe easier to talk out on the phone?). And I would need to ask vendors what their standard practice is but I imagine it is similar to what Armin described.

.github/workflows/release.yml
HEdingfield: Merge conflicts still need resolving, even after branch change.

src/main/java/network/brightspots/rcv/Main.java
HEdingfield: That sounds perfect.

HEdingfield (commented):
>> Also, I'm happy to test the bins for Linux and Windows if you can point me to where they are.
>
> https://github.com/BrightSpots/rcv/actions/runs/4801866529

Unfortunately I can't install this package on my Linux machine due to org policy. We'll have to find another way to test (e.g. VirtualBox).

I was, however, able to verify that the zip built by the current release.yml works on Linux though!

.github/workflows/release.yml
artoonie: Yeah, the merge conflict is due to `develop` removing build-on-branches and this branch adding it back so I can test it on every push. I didn't plan to resolve until all testing was complete, since it's a nice reminder to delete that line and prevents us from accidentally keeping it. I can resolve if you prefer though, there are other ways to make sure it doesn't get merged (e.g. a Request Changes review on GH).

.github/actions/gradle-and-sha/action.yml
artoonie: No, this was just debugging code since every OS spat out several file options, so this was a way to inspect and see which we wanted (e.g. Windows has MSI and EXE; Mac has DMG and .App). I've removed it now.

build.gradle
artoonie: Agreed. For some reason, I thought the modules would be space-separated rather than comma-separated, so I thought only one module was allowed. I'll update the comment for future travelers.

src/main/java/network/brightspots/rcv/Main.java
artoonie: https://github.com/BrightSpots/rcv/issues/662 and also noted in the comments now, feel free to add any context needed to that issue

.github/workflows/release.yml
HEdingfield: Ok, I'll just review by diffing in IntelliJ against this file, but I'll hold off on doing the final approval and merge until the conflict is resolved, because I think it's a good idea to have another pair of eyes when resolving merge conflicts.

.github/workflows/release.yml
```diff
@@ -0,0 +1,103 @@
...
+        shell: bash
+        run: echo "FILEPATH=build/${{ steps.basefn.outputs.FILEPATH }}.zip" >> $GITHUB_OUTPUT
+      - name: "Get extension"
+        id: ext
+        shell: bash
+        run: |
+          if [ ${{ runner.os }} == 'Windows' ]; then
+            echo "EXT=.exe" >> $GITHUB_OUTPUT
+          elif [ ${{ runner.os }} == 'Linux' ]; then
+            echo "EXT=" >> $GITHUB_OUTPUT
```
HEdingfield, line 37: In my failed Linux check today, the lack of extension was a little confusing, but the OS seemed to at least recognize it was a package when I tried to run it (and rejected it due to security policy, as mentioned). Hopefully it's correct to keep it without an extension (e.g. no `.deb`). Will leave it up to your Virtualbox testing to verify.

.github/workflows/release.yml
```diff
@@ -0,0 +1,103 @@
...
+          else
+            echo "EXT=.dmg" >> $GITHUB_OUTPUT
+          fi
+      - name: "Get jpackage output filepath"
+        id: jpackagefn
+        shell: bash
+        run: |
+          # TODO Issue-662 sync version number with Main.java and build.gradle
```
HEdingfield, line 45: Can you leave a comment about why these numbers need to be hard-coded in here? I have a feeling the question will come up repeatedly in the future.

.github/workflows/release.yml
```diff
@@ -0,0 +1,103 @@
...
+            ${{ github.workspace }}/${{ steps.exefn.outputs.FILEPATH }}
+            ${{ github.workspace }}/${{ steps.exefn.outputs.FILEPATH }}.sha512
+          retention-days: 90
+
+      - name: "Upload binaries to release"
+        uses: svenstaro/upload-release-action@v2
+        if: ${{ github.event_name }} == 'release'
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: ${{ github.workspace }}/${{ steps.zipfn.outputs.FILEPATH }}*
```
HEdingfield, line 100: Is it only uploading the jlinkzip version since this is zipfn here? I think we'd also want to upload the jpackage versions.

I also noticed that the artifact file you linked to for Linux testing was a single file with no extension that turned out to be a zip with every single platform and SHA in it. Hopefully we're not uploading something like that to the release, as it was fairly confusing. I'd prefer individual files as the current release.yml does.

src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -26,6 +26,7 @@
 @SuppressWarnings("WeakerAccess")
 public class Main extends GuiApplication {
 
+  // TODO Issue-662 - sync these values with release.yml build.gradle
```
HEdingfield, line 29: Here and in the other two spots: I recommend linking the full URL just in case this gets forked, e.g.:

// TODO: Sync these values with release.yml and build.gradle (https://github.com/BrightSpots/rcv/issues/625)

artoonie:
@chughes297 the logo needs to be a perfect square. The two logos I see are either slightly landscape or slightly portrait. Of these three variations, which do you prefer? Or something in-between? Whitespace would not look great on the logo, thus Option 3, but Option 3 doesn't match your other iconography.

image
image
image

.github/workflows/release.yml
artoonie: 1. Before we merge this, we need to decide whether we want to replace jlinkzip with jpackage, release both, or continue just releasing jlinkzip until we've had more time/feedback on jpackage. I believe that decision is [pending feedback from vendors](https://github.com/BrightSpots/rcv/pull/660#issuecomment-1523536314).
2. Good catch on the linux release! No, that's not intentional. I'll fix it.

.github/workflows/release.yml
artoonie: I've corrected this: previously, I just shipped the binary (thus no extension), but changed it to use .deb now and forgot to sync that change.

The .deb seems to be correct -- however, I haven't booted my linux partition in several years and it seems like testing on my end will be pretty hard. The machine has no access to the internet (seems like the wifi hardware is too new), and when manually copying the .deb via a USB stick, it's unable to open it because it uses a newer compression format not supported by Ubuntu 16.04.

All this to say: if somebody else has faster access to Linux, could you test it fully? Otherwise, I'll need to install virtualbox or reformat my linux partition, both of which will take some time to set up.

.github/workflows/release.yml
HEdingfield: Oh, and I didn't notice the RCVRC logo at any point. The installer just had a red CD-looking thing, and the program icon was the Java Star Trek looking icon.

.github/workflows/release.yml
artoonie: Ah, that does remind me-- I've got some Raspberry Pis ready-to-go laying around. I'll see if I can boot one of those up.

.github/workflows/release.yml
artoonie: Ah, I understand: yes, that is a limitation of Github Artifacts on Actions. In a release, they will be individual files, but Github forces them all into one giant package on Github Actions Artifacts.

i.e. this won't be an issue when we create releases, only in development.

HEdingfield:
I believe this would resolve #57 as well.

artoonie:
> I believe this would resolve #57 as well.

Indeed :) that's the branch name.

HEdingfield:
> > I believe this would resolve #57 as well.
> 
> Indeed :) that's the branch name.

Ah yeah! I was just tagging the issue in GH so this PR would actually appear in its history to help keep things organized.

.github/workflows/release.yml
artoonie: i've reinstalled a more recent Ubuntu and confirmed that this works, though the install doesn't put the binary in the $PATH -- you have to manually launch it from `/opt/`. I can't figure out how to make the package more easily launchable, so I'm happy to call this "good enough" for now given the lack of linux users, and the easy workaround. I've added info to the README. We can revisit later if we want to make this simpler, but for now, just run `/opt/rctab/bin/RCTab` to launch after installing the .deb.

artoonie:
This is almost ready for review. I've tested on Linux and added an app icon since the last convo. Last step will be to sign the mac application.

I am still worried about this never getting tested until we run a release -- but, as requested, I have set this to only run on release.

HEdingfield:
Nice! Shout out when you think it's ready.

> I am still worried about this never getting tested until we run a release -- but, as requested, I have set this to only run on release.

I have no issue at all with temporarily making it run during events other than release for testing purposes. Then we can just revert it to run only on release once we're satisfied.

build.gradle
```diff
@@ -97,20 +97,78 @@ javafx {
...
+               '--strip-debug', '--compress', '2',
+               '--no-header-files', '--no-man-pages'
     mergedModule {
         requires "java.xml"
     }
     launcher {
-        name = "rcv"
+        // TODO Sync version number with release.yml and Main.java (github.com/BrightSpots/rcv/issues/662)
+        name = "RCTab"
+        version = "1.3.1"
```
HEdingfield, line 118: I assume it was 1.3.1 just for testing purposes? Can we set to 1.4.0 now?

build.gradle
```diff
@@ -97,20 +97,78 @@ javafx {
...
+            installerOptions += [
+              '--type', 'dmg', 
+              '--icon', 'src/main/resources/network/brightspots/rcv/launcher.icns',
+              '--resource-dir', 'src/main/resources/network/brightspots/rcv/',
+              '--mac-sign',
+              '--mac-signing-key-user-name', 'Developer ID Installer: Election Administration Resource Center (A257HB4NS4)',
+              '--mac-signing-keychain', 'build.keychain',
+              '--mac-package-name', 'RCTab',
+              '--mac-package-identifier', 'network.brightspots.rcv',
+              '--mac-package-signing-prefix', 'network.brightspots.rcv'
```
HEdingfield, line 169: I appreciate how painful it must have been to get all these params just right!

.github/workflows/release.yml
```diff
@@ -27,40 +68,76 @@ jobs:
...
+          xcrun stapler staple ${{ steps.exefn.outputs.FILEPATH }}
 
       - uses: actions/upload-artifact@v3
         with:
           name: Package
           if-no-files-found: error
           path: |
-            ${{ github.workspace }}/${{ steps.fn.outputs.FILENAME }}
-            ${{ github.workspace }}/${{ steps.fn.outputs.FILENAME }}.sha512
+            ${{ github.workspace }}/${{ steps.zipfn.outputs.FILEPATH }}
```
HEdingfield, line 129: I think I asked about this previously but forgot the answer... I see in the test run [here](https://github.com/BrightSpots/rcv/actions/runs/5213169580) that it creates a single file called "Package", with all three versions in it. Is it possible to upload the individual versions and their hashes separately instead of all bundled together?

.github/workflows/release.yml
```diff
@@ -1,8 +1,11 @@
-name: Build Releases
+name: Generate Releases
 
 on:
   release:
     types: [published]
+  push:
+    branches:
+      - 'feature/issue-57_standalone-executable'
```
HEdingfield, line 8: Do we update this to `master`?

.github/workflows/release.yml
```diff
@@ -11,10 +14,48 @@ jobs:
...
+      - name: "Get jpackage output filepath"
+        id: jpackagefn
+        shell: bash
+        run: |
+          # TODO Sync version number with Main.java and build.gradle (github.com/BrightSpots/rcv/issues/662)
+          # The version numbers are hardcoded because the files below include the version number in them,
+          # and while we could use some regex to figure out the version number automatically, it seems cleaner
+          # to know the expected version number upfront.
+          if [ ${{ runner.os }} == 'Windows' ]; then
+            echo "FILEPATH=build/jpackage/RCTab-1.3.1.exe" >> $GITHUB_OUTPUT
```
HEdingfield, line 49: Update 1.3.1 in these three spots to 1.4.0 as well?

.github/workflows/release.yml
```diff
@@ -11,10 +14,48 @@ jobs:
...
+          fi
+      - name: "Get jpackage output filepath"
+        id: jpackagefn
+        shell: bash
+        run: |
+          # TODO Sync version number with Main.java and build.gradle (github.com/BrightSpots/rcv/issues/662)
+          # The version numbers are hardcoded because the files below include the version number in them,
+          # and while we could use some regex to figure out the version number automatically, it seems cleaner
+          # to know the expected version number upfront.
+          if [ ${{ runner.os }} == 'Windows' ]; then
```
HEdingfield, line 48: I noticed the file I downloaded inside `Package.zip` is called `rctab_feature_issue-57_standalone-executable_Windows.exe`, which seems off... it should be the version number, shouldn't it?

HEdingfield, line 48: Not a huge deal, but the start menu icon in Windows still has the Java icon instead of the launcher icon you're including in this PR. Not sure if that's possible to fix easily?

HEdingfield (commented):
Some light changes for you [here](https://github.com/BrightSpots/rcv/pull/660/commits/af18a07eba4d9b0df0ace77a9cf115beb6ffbec0) to review too, per our discussion.

src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -27,7 +27,10 @@
 public class Main extends GuiApplication {
 
   public static final String APP_NAME = "RCTab";
-  public static final String APP_VERSION = "1.4.0.alpha";
+
+  // TODO Sync version number with release.yml and build.gradle:
+  // github.com/BrightSpots/rcv/issues/662
+  public static final String APP_VERSION = "1.4.0-alpha";
```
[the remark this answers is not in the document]
artoonie: `jpackage` required in one of the OSs, I forget which, maybe all. It's standard semantic versioning to use a dash instead of a dot here: https://semver.org/

build.gradle
artoonie: I'm a little afraid to set it to 1.4.0 prematurely, as we'd have fake but signed `1.4.0` versions floating around. I guess the same is true with `1.3.1`?

The issue stems from the fact that Mac won't allow a bundle to have a `-alpha` at the end, so we need some way of signifying a fake/development version number. See: https://stackoverflow.com/questions/70066384/javafx-gradle-jlink-plugin-jpackageimage-fails-because-of-1-0-snapshot-conta

Perhaps we should use 0.0.0 unless we are building a release? Or maybe this just isn't a real issue, and we're okay having "false" 1.4.0s around?

build.gradle
artoonie: The a-ha moment was realizing I need two different keys, one for the installer, one for the application. *headdesk*

.github/workflows/release.yml
artoonie: Only in a Release, but not in an Artifact. So here, it would be possible: https://github.com/BrightSpots/rcv/releases/tag/

And it is indeed what happens there -- just not in the Artifact.

.github/workflows/release.yml
artoonie: I think we delete this entirely when we're ready to merge, and create a Release instead.

.github/workflows/release.yml
artoonie: It will be in the final release! Just not in development builds, since it uses the ref name. The ref name is the branch name for these dev builds, and the release name for release builds.

artoonie: Done! Windows needed a `ico` file instead of `png`.

.github/workflows/release.yml
HEdingfield: Let's update all these to 1.4.0-alpha too.

.github/workflows/release.yml
HEdingfield: SGTM, let's keep this convo open and I'll make my final approval contingent upon deleting this.

build.gradle
HEdingfield: Dumb Apple. That may be a blessing in disguise in a small way, though, because we don't want to be creating these builds / releasing alpha versions. I'd say we set it to `1.4.0-alpha` here now so the find-replace is easier later (plus I guess it'll block building).

src/main/java/network/brightspots/rcv/Main.java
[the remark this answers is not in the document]
HEdingfield: Fair point. I think we're going to need to update `isVersionNewer` in this PR too though if you want to do this here now. Alternatively, you could leave it as-is and we can file another P1 issue to deal with this.

.github/workflows/release.yml
HEdingfield: Good catch. I'll test this out and verify one last time before final approval.

src/main/java/network/brightspots/rcv/Main.java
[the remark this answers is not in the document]
artoonie: https://github.com/BrightSpots/rcv/issues/700 i'll deal with it separately

.github/workflows/release.yml
HEdingfield: Worked perfectly!

HEdingfield (approved):
Fantastic job! This would have taken me forever to get going.

PR #665, opened 2023-05-14
Fixes `DominionCvrReader` bad merge

HEdingfield:
Fixed bug where `contestId` was being rendered as an `int`; clarified logic around `dominionCandidateId`.

The bad merge occurred between PR #639 and #640.


artoonie (approved):
thanks for catching this!

PR #670, opened 2023-05-23
fix compilation caused by uncaught merge conflicts

artoonie:
Merging #649 caused some compilation issues.

### DIFF
--- a/src/main/java/network/brightspots/rcv/CsvCvrReader.java
+++ b/src/main/java/network/brightspots/rcv/CsvCvrReader.java
@@ -86,7 +86,7 @@ class CsvCvrReader {
           String candidateId = candidateIds.get(col);
           if (col == undeclaredWriteInColumn) {
             candidateId = Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL;
-          } else if (!contestConfig.getCandidateCodeList().contains(candidateId)) {
+          } else if (contestConfig.getNameForCandidate(candidateId) == null) {
             unrecognizedCandidateCounts.merge(candidateId, 1, Integer::sum);
           }
           rankings.add(new Pair<>(rankAsInt, candidateId));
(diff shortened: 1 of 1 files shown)


HEdingfield (approved):
Very confused by how this could have happened... I thought it required all tests to pass before we could merge it?

artoonie:
There's a setting in github that requires every PR to be "up to date" with the main branch before merging, but that's not enabled.

This means that if you have two pull requests, `Main < A` and `Main < B`, the tests run for `Main Merged With A` and `Main Merged With B`, but only runs for `Main Merged with A + B` when both pull requests have landed.

Enabling the setting is annoying because 98% of the time, it just adds busy work, but it catches instances like this where `Main+A` passes, `Main+B` passes, but `Main+A+B` fails.

HEdingfield:
Ok, interesting. If we were to turn this on, I don't quite understand what the additional busy work would be most of the time?

artoonie:
If you have 3 PRs, you'd have to:
1. Merge PR1 into Main
2. Merge Main into PR2
3. Merge PR2 into Main
4. Merge Main into PR3
5. Merge PR3 into Main

In practice, you'd end up merging Main into each PR frequently to get the green check, so steps 2 and 4 often happen several times, just so all PRs are ready to be merged in any order:

1. Merge PR1 into Main
2. Merge Main into PR2
3. [etc for every PR 3..n]
4. Merge PR2 into Main
5. Merge Main into PR3
6. Merge PR3 into Main
7. [etc for PR 4..n]

I think it's worth enabling then reconsidering later if it was too painful. If there aren't many live PRs at once, it's not a problem.

tarheel:
I'm inclined to agree that this setting would cause more annoyance than it would solve. But I don't feel strongly about it.

HEdingfield:
I think this is actually what we use at work (where we do squash commits), so I think it'd be great to set it up this way and see how it works for us.

Is it the setting "Require branches to be up to date before merging"? If so, it looks like that's already set in `develop`. Not sure how this situation would have happened if that's true? Or is it another setting?

Issue #645, opened 2023-03-22
Handle multiple Candidate IDs in CDF


artoonie:
A CDF can have multiple candidate IDs. Currently, there is a warning:
```
"CandidateSelection \"%s\" has multiple CandidateIds. Only the first one will be processed.",
```
Once we support aliases (https://github.com/BrightSpots/rcv/pull/639), we won't need this limitation.

```
  "Election" : [ {
    "Contest" : [ {
      "ContestSelection" : [ {
        "CandidateIds" : [ "id0", "id1" ]
      },
```

artoonie:
Note: will wait for https://github.com/BrightSpots/rcv/pull/639 to merge before working on this to avoid creating a too-long PR chain.

HEdingfield:
@chughes297 could you please comment on the questions raised in #684?

HEdingfield:
@chughes297 commented in #684:

> If I'm understanding things correctly, the only time a candidate will have multiple IDs in the CDF is in a party ticket context? If my understanding is correct, I'd say yes we don't need to support this. Nowhere using RCV now (or in the near term) combines it with party ticket, so this doesn't seem necessary/valuable to include.

@artoonie if what Chris says is the case, should we close that PR and this issue?

artoonie:
Note: abandoned this; not something we need. See discussion in #684

PR #684, opened 2023-06-05
support multiple candidate IDs in CDF

artoonie:
The CDF format can have multiple candidate IDs to support party ticket voting options. A comment says "in practice this is always a single candidate ID".

I'm not sure we actually want to do this -- I thought the candidate IDs were aliases, but they're not quite the same.

This PR looks for the first-found candidate ID, which is a reasonable thing to do, but also adds potentially-unnecessary complexity since we'd never encounter this situation (right?).

Thoughts on ignoring this ticket entirely vs merging this PR?

### DIFF
--- a/src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
+++ b/src/main/java/network/brightspots/rcv/CommonDataFormatReader.java
@@ -29,6 +29,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


HEdingfield:
This looks to address #645. Would like to get @chughes297's thoughts on this before reviewing / merging.

chughes297:
If I'm understanding things correctly, the only time a candidate will have multiple IDs in the CDF is in a party ticket context? If my understanding is correct, I'd say yes we don't need to support this. Nowhere using RCV now (or in the near term) combines it with party ticket, so this doesn't seem necessary/valuable to include.

artoonie:
Sounds like we're aligned -- I'll go ahead and close this and mark the ticket as resolved.

PR #686, opened 2023-06-08
Create codeql.yml

artoonie:
Thoughts on using this free vulnerability scanner?

File is auto-generated by github. It seems like it could help. I use something similar over on RCVis and it's nice.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,77 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


tarheel (commented):
Not familiar with this, but happy to give it a try if you think it might be useful. I'll let @HEdingfield comment also.

HEdingfield:
Found more info on it [here](https://codeql.github.com/docs/codeql-overview/about-codeql/). It says it scans the built files -- what kind of errors does it actually show? I'm wondering how, if it's scanning compile stuff, it could link you back to specific lines of could that need addressing?

Also, roughly how long does it take to run in your experience? I assume it'd run every time you push a commit to a PR?

artoonie:
1. Here's an example: https://devblogs.microsoft.com/cppblog/code-scanning-with-github-actions/ 
2. This run took 2 minutes, which is less than the `Test` action, so it won't add any time to our end

It's helped catch a few things in RCVis; always found it nice to have layers of safety nets. Though -- I'm surprised it didn't find anything on its first run, that seems suspicious.

HEdingfield:
> Though -- I'm surprised it didn't find anything on its first run, that seems suspicious.

Agreed -- can you test spiking our code with something you know should trigger it in this PR and then see if it works?

If not, it might not be worth including.

tarheel:
It's a get-nothing-for-nothing PR!

PR #687, opened 2023-06-08
make all output files read-only

artoonie:
WIP: Depends on whether this satisfies requirements


HEdingfield:
@chughes297 per @artoonie's comment above, does this satisfy requirements and should we go forward with it?

chughes297:
haven't heard either way but i see no harm in adding this.

HEdingfield (approved):
Works after fixing asserts. 👍

PR #688, opened 2023-06-08
Don't force re-saving of TEST files

artoonie:
This turned out to be a little trickier than expected.

*The naive solution* was to just not save a file whenever the version was TEST.
This didn't work because:
1. Too many places assume we're reading from a file, so I couldn't just use the in-memory RawContestConfig
2. There was no easy way to tell if there have been any modifications in the UI, i.e. truly force a re-save

I opted to have this, which has a pretty nice UX to me:
1. User has the option of saving before tabulation, or proceeding without saving
2. The latter creates a temporary file that is cleaned up on program close
3. The temporary file is placed next to the original so all relative paths still exist


src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -953,29 +955,35 @@ private boolean checkForSaveAndContinue() {
     return willContinue;
   }
 
-  private boolean checkForSaveAndExecute() {
-    boolean willContinue = false;
+  private String commitConfigToFileAndGetFilepath() {
+    String filename = null;
     if (checkIfNeedsSaving()) {
       ButtonType saveButton = new ButtonType("Save", ButtonBar.ButtonData.YES);
+      ButtonType dontSaveButton = new ButtonType("Proceed without Saving", ButtonBar.ButtonData.NO);
```
HEdingfield, line 962: 1) We need a cancel button too.
2) If a user clicks the X in the top right corner of the dialog, that should also act as cancel (it currently proceeds with the tabulation).

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -453,10 +453,11 @@ public void menuItemValidateClicked() {
    * changes, and creates and launches TabulatorService from the saved config path.
    */
   public void menuItemTabulateClicked() {
-    if (checkForSaveAndExecute()) {
+    String filepath = commitConfigToFileAndGetFilepath();
```
HEdingfield, line 456: > I opted to have this, which has a pretty nice UX to me

Agree that it's a cool idea, but I think it poses a problem in that we need to ensure that the config file exactly matches what was actually used to run a tabulation. If we could make this pop-up _only_ appear if the version is `TEST`, that would be more palatable I think.

HEdingfield (changes requested):

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: Three buttons seems pretty confusing (and is [poor UX practice](https://ux.stackexchange.com/a/131474) if it can be avoided).

If the user hits tabulate and there are no changes, there is no opportunity to cancel. Why would it be different if there are changes? Just in case the changes were unintended?

I'm okay hooking up the `X` to cancel` but hoping we can avoid a three-button dialog.

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: Perhaps this is the solution to what I mentioned above: if version is `TEST`, then it follows this workflow. If it's anything else, it reverts to the old workflow (`save` or `cancel`).

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: There existing dialog warning users that they need to save it has a cancel button; and yeah, in case the changes were unintended is a good reason.

Sure, I guess I'm cool with still allowing the user to cancel by pressing the X if you don't want to add a third button.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Yep, that sounds good to me.

artoonie: Alright, done!

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: The X button is still proceeding with the tabulation.

HEdingfield: BTW, it appears we do have existing dialogs that have three buttons; see `checkForSaveAndContinue`.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -953,29 +969,41 @@ private boolean checkForSaveAndContinue() {
...
         if (fileToSave != null) {
           saveFile(fileToSave);
-          willContinue = true;
+          filename = fileToSave.getAbsolutePath();
         }
+      } else if (result.get() == dontSaveButton) {
+        File tempFile = new File(selectedFile.getAbsolutePath() + "_temp");
+        filename = tempFile.getAbsolutePath();
+        saveFile(tempFile);
+        tempFile.deleteOnExit();
```
HEdingfield, line 999: Curious why we wait until the program is exited to delete the config?

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -953,29 +969,41 @@ private boolean checkForSaveAndContinue() {
...
+    if (comparisonResult != ConfigComparisonResult.SAME) {
       ButtonType saveButton = new ButtonType("Save", ButtonBar.ButtonData.YES);
+      ButtonType dontSaveButton = new ButtonType("Proceed without Saving", ButtonBar.ButtonData.NO);
+      ButtonType cancelButton = new ButtonType("Cancel", ButtonBar.ButtonData.NO);
+      ButtonType noButton = comparisonResult
+          == ConfigComparisonResult.DIFFERENT_BUT_VERSION_IS_TEST ? dontSaveButton : cancelButton;
       Alert alert =
           new Alert(
               AlertType.WARNING,
               "You must either save your changes before continuing or load a new contest config!",
```
HEdingfield, line 984: This message also appears when using test files, which is a little misleading now that it's possible to bypass it. Maybe we make the message say this for tests:

"You are using a test config, and must either save the file or proceed with creating a temporary one!"

HEdingfield: And then maybe we also change the "Proceed without Saving" button to "Proceed with Temporary File".

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: Fair enough -- I'll have three buttons only for developers, which makes me feel better about it. The third button only appears if version is TEST.

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: We could also do it when tabulation completes, or when another file is selected, but that would require following this around and adding some complexity since Java doesn't have a destructor. `deleteOnExit` was the cleanest one-line solution.

artoonie:
@HEdingfield comments addressed!

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Gotcha, thanks for clarifying. Generally speaking, this SGTM, but I did find a corner case accidentally that's a little annoying for development: if you terminate the program early via the CLI (or in IntelliJ by clicking the stop button), it won't actually clean up after itself and it'll just leave the temp files there indefinitely.

Could you investigate deleting right after tabulation a little bit? If it looks like it'll take more than 10-20 minutes of work, or increase the code complexity significantly / in an ugly fashion, I'm fine leaving it this way and just trying to remember not to stop the tabulator in this way.

HEdingfield (commented):
Submitted a commit to fix a few nits / linting issues, and have one outstanding investigation request, but I think this about ready to approve. Great work!

tarheel:
Just a nice to have, probably, but ideally the app would also check for old temp files and clear them out every time it starts up and/or every time it takes some other action like running a tabulation. That would more thoroughly address the edge case that @HEdingfield raised.

tarheel:
I don't know about storing state between runs, but I think using a temp directory would be ideal if we can figure out a clean solution for the file path issue (which I imagine we could). But again, just a nice to have.

HEdingfield:
I personally think this is not worth the effort, as 1) it's super unlikely to happen again now that it deletes the files right after the run, and 2) you can just manually delete them. Plus it'd be a pain to deal with relative pathing if we put them in a separate folder, etc.

Issue #642, opened 2023-03-18
Re-save all test config files


HEdingfield:
And upload the new versions to reflect latest changes once 1.4.0 development is drawing to a close (see [this thread](https://github.com/BrightSpots/rcv/pull/636#discussion_r1141068580) for context).

HEdingfield:
TBD how or if #664 will affect this issue.

HEdingfield:
Before resolving, review https://github.com/BrightSpots/rcv/pull/690/files/1a1f0a39fd5451944c7d041e90e928881ddbf36c#r1227285308.

HEdingfield:
Probably rename the `testPortlandMayorCodes` to testPortlandMayorAliases` (and any other tests named like this) as part of this process.

HEdingfield:
Relevant, and might want to wait until this is done as well: #723.

artoonie:
Will get to this after everything else is merged to avoid conflicts

PR #690, opened 2023-06-09
deprecate the concept of "codes" (almost) entirely

### LINKED ISSUE #663: Remove notion of "Candidate Codes" altogether
Based on commentary throughout #639, we should remove the notion of candidate codes altogether, and replace with candidate name wherever a unique name is needed.

Commentary:
https://github.com/BrightSpots/rcv/pull/639#discussion_r1179977900
https://github.com/BrightSpots/rcv/pull/639#discussion_r1179996601

artoonie:
resolves #663 

This removes the concept of "codes" entirely. In doing so, all dominion tests have been updated to use candidate names instead of codes.

In order to ensure that the tests were only changed for exactly what we expect, I didn't simply replace test files with their updated version; I used the script below to update each test. It's not documented because it's a one-time throwaway, but let me know if you have questions on it:

```python
import json
import sys

# To run: python3 test.py <test-directory-name> [--save].
base = sys.argv[1]
doSave = len(sys.argv) > 2 and sys.argv[2] == "--save"
print(f"working on {base} and doSave={doSave}")

# Map candidate codes to candidate names
codeToName = {}
with open(f"{base}/{base}_config.json", 'r') as f:
    data = json.loads(f.read())
for candidate in data['candidates']:
    codeToName[candidate['code']] = candidate['name']

# Undeclared write-in maps to itself, e.g. 12=12. Output will have the numerical code.
uwiLabel = data['cvrFileSources'][0]['undeclaredWriteInLabel']
codeToName[uwiLabel] = uwiLabel
contestId = data['cvrFileSources'][0]['contestId']

# For each line, replace candidate codes with their names instead.
lines = []
with open(f"{base}/{base}_contest_{contestId}_expected.csv", 'r') as f:
    for rawline in f.readlines():
        if rawline.startswith('Contest'):
            lines.append(rawline.strip())
        else:
            line = rawline.strip().split(',')
            for i,item in enumerate(line[6:]):
                if item.strip() in ['undervote', 'overvote']:
                    continue
                line[i+6] = codeToName[item.strip()]
            lines.append(','.join(line))

print('\n-----\n'.join(lines))

if doSave:
    with open(f"{base}/{base}_contest_{contestId}_expected.csv", 'w') as f:
        f.write('\n'.join(lines))
```


src/test/resources/network/brightspots/rcv/test_data/dominion_multi_file/dominion_multi_file_contest_13_expected.csv
```diff
@@ -1,26 +1,26 @@
...
-13,18,504,51,,PCT 7335,28,undervote,undervote,undervote
-13,18,504,37,,PCT 7311,28,29,undervote,undervote
-13,18,504,20,,PCT 7331,27,28,30,29
-13,18,504,3,,PCT 7348,27,undervote,undervote,undervote
-13,18,504,22,,PCT 7317,undervote,undervote,undervote,undervote
-13,18,504,25,,PCT 7315,29,30,27,28
-13,18,504,4,,PCT 7348,27,28,undervote,undervote
-13,19,664,45,,PCT 7326,27,28,29,30
-13,19,664,67,,PCT 7321,28,29,30,undervote
+Contest Id,Tabulator Id,Batch Id,Record Id,Precinct,Precinct Portion,Rank 1,Rank 2,Rank 3,Rank 4
```
HEdingfield, line 1: On the topic of tests, does `testPortlandMayorCodes` need to be updated somehow as well?

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1063,10 +1063,6 @@ String getNameForCandidate(String nameOrAlias) {
     return candidateAliasesToNameMap.get(nameOrAlias);
   }
 
-  String getCodeForCandidate(String nameOrAlias) {
```
HEdingfield, line 1066: In [this comment](https://github.com/BrightSpots/rcv/pull/639#discussion_r1141238943), you mentioned possibly reverting / getting rid of some lines in `TabulatorSession`. Is that no longer feasible?

HEdingfield (commented):
Please review the commit I added to clean up a bit more and make sure it looks good to you.

src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -217,11 +217,6 @@ public static class Candidate {
     private boolean excluded;
     private List<String> aliases = new ArrayList<String>();
 
-    // The code is a special alias which is used in the output files instead of
-    // the display name. Other than output displays, it is not handled specially:
-    // it is just another alias.
-    private String code;
```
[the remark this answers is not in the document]
artoonie: Exactly -- thus the (almost) entirely. Dominion has a concept of a `code`, which gets translated to an `alias` everywhere outside of dominion's reader.

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: I pushed a commit showing the effect of that. Depends on what we want -- we can get rid of `getNameForCandidate` in `ResultsWriter`, which generates results that use the candidate code instead of the candidate name. See the diff [here](https://github.com/BrightSpots/rcv/pull/690/commits/3f121b7a69aefc25d749838e64d82244458b07dc).

I'm not sure how the `_expected.csv`s are used, so I don't know which format is more useful, but it seems likely that it's more useful with canonical names (i.e. reverting #3f121b7, the next commit) and keeping the code as it was.

Let me know which you prefer.

HEdingfield: The `_expected.csv`s are just used for our tests.

Hard for me to answer this question... might be best for @tarheel or @chughes297 to weigh in? It appears that with the above commit you linked, it leaves our tests as-is, which is probably preferred?

src/test/resources/network/brightspots/rcv/test_data/dominion_multi_file/dominion_multi_file_contest_13_expected.csv
HEdingfield: Heads-up that #642 will swap these over to `aliases` whether we like it or not. I can make a note in there to keep it as-is if we care about testing old file formats, but I don't really think we do.

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: Are the outputted CSVs never used by election administrators? If so, then I suppose it doesn't matter.

tarheel: The real output CSV are definitely used by admins. @HEdingfield is just referring specifically to the `_expected.csv` files that are in the test directories.

Apologies if I missed a relevant part of the discussion -- I'm just looking at this comment thread -- but I think ideally the output CSV would have the canonical names in one column and then perhaps a separate column that lists any alias that are used for each candidate (separated by a delimiter).

artoonie: Hm-- let's separate adding both canonical names and aliases for another task?

For now, I think the question is: should the output CSVs have the candidate code or the canonical name?

The current PR and the existing codebase use the candidate code.
Reverting the last commit in this PR will use the canonical name, and simplify `ResultsWriter` a smidge.

chughes297: We should use canonical names wherever possible, I don't believe it's very useful for jurisdictions to get results using candidate codes.

artoonie: Great. I think that makes sense too. I reverted the last commit, and this PR should be ready to go now.

HEdingfield: > Hm-- let's separate adding both canonical names and aliases for another task?

@chughes297 @tarheel do y'all think it's worth creating a new issue for this? If so, is it something we'd need to include in 1.4.0?

HEdingfield (approved):
LGTM!

src/main/java/network/brightspots/rcv/ContestConfig.java
tarheel: Yes, I think it would be useful to list aliases in the output spreadsheet. I doubt it's a requirement for v1.4, but @chughes297 can comment.

HEdingfield: Created #705 for this.

PR #697, opened 2023-06-13
Adds note in UI that names must match exactly

### LINKED ISSUE #677: Notify users that candidate names must match names in CVR files exactly
I imagine this should be right at the start of the info box on the candidates tab. 

Not sure how this interacts with #60 (or #591 ,#628 ,  #663 , and #645) (I just looked for all issues mentioning candidates that seemed relevant).

HEdingfield:
Fixes #677.

image

### DIFF
--- a/src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
+++ b/src/main/resources/network/brightspots/rcv/GuiConfigLayout.fxml
@@ -322,8 +322,12 @@
                     </padding>
                     <Label text="Manually Add Candidate" style="-fx-font-weight: bold"/>
                     <Text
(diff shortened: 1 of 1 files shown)


artoonie:
I wasn't sure what this meant until I saw the original email. I wonder if it's worth expounding a tiny bit:
`Candidate names are case-sensitive and must match the CVR exactly, including any HTML tags or other formatting present in the CVR.`?

SInce the real issue was not adding `<B></B>`, and this text doesn't quite tell me that's needed.

Also: might be nice to have a slightly smaller font? That's a lot of words that are visually equally weighted, which makes my eyes glaze over all of it and makes me less likely to read any of it.

HEdingfield:
How's this?

image

artoonie (approved):

PR #698, opened 2023-06-13
Converts non-test asserts to error checks

### LINKED ISSUE #695: Convert assert statements to error checks
Per [this SO post](https://stackoverflow.com/a/7865255/3846321), `assert` statements in Java are generally ignored when running a program. All existing statements in the codebase outside of tests should be converted to proper error checks.

(We bumped up against this while working on #688; an attempt to delete a file executed as an assert statement wasn't happening at all.)

HEdingfield:
Fixes #695.


src/main/java/network/brightspots/rcv/CandidateRankingsList.java
```diff
@@ -23,31 +23,19 @@
...
+   * Precondition: must be called with a 1-indexed ranking number. Note: just because this returns
+   * false at ranking N doesn't mean it will also return false at N+1 -- specifically, that will not
+   * be true if there are skipped rankings.
+   *
+   * @param num A value >= 1.
+   * @return Whether the candidate has a ranking at the given value.
+   */
+  boolean hasRankingAt(int num) {
+    if (num < 1) {
+      throw new IllegalArgumentException();
```
artoonie, line 36: Curious what motivated this move?

artoonie (approved):
Seems fine, though we could also enable assertions in build.gradle: https://stackoverflow.com/questions/48396274/how-to-enable-assertions-in-the-gradle-run-task

Something about throwing an assertion error feels incorrect to me, like we're abusing assertionerrors, but I don't have strong opinions here.

src/main/java/network/brightspots/rcv/CandidateRankingsList.java
HEdingfield: Ah, that was automatic IntelliJ stuff. The original three collaborators worked out years ago that we'd keep all the boxes checked when committing in IntelliJ:

image

HEdingfield:
> Seems fine, though we could also enable assertions in build.gradle: https://stackoverflow.com/questions/48396274/how-to-enable-assertions-in-the-gradle-run-task

Ah ok, I'd be ok with that too if we're sure that would hold for the standalone executable version too? However, I do tend to like the more explicit coding style of actually throwing the errors, since it gives us more control over the type / behavior in these situations.

> Something about throwing an assertion error feels incorrect to me, like we're abusing assertionerrors, but I don't have strong opinions here.

Agreed, but that _is_ what assert does under the hood, and I think this is just more explicit about it. Whether or not we should actually be throwing all these `AssertionError`s is a good question, maybe a bit orthogonal to this issue.

Will hold off on merging until hearing back on  these points.

src/main/java/network/brightspots/rcv/CandidateRankingsList.java
artoonie: Ah, perhaps worth adding this to the contributor guide? https://github.com/BrightSpots/rcv/blob/develop/contributor_guide.md

artoonie:
I like explicit errors too, unless I can nearly guarantee that the assertions are superfluous. Throwing a lot of exceptions can make code unreadable in a way that assertions don't -- but that's purely theoretical, as I think these changes are reasonable. My only concern was using `AssertionErrors` instead of something a little more precise, but -- I'm fine with this as-is, so feel free to merge!

HEdingfield:
> My only concern was using `AssertionErrors` instead of something a little more precise, but -- I'm fine with this as-is, so feel free to merge!

Yeah, that's a good point. I made another pass making all these more specific. Let me know of that looks good to you (I did get rid of one check that I didn't think was really useful).

src/main/java/network/brightspots/rcv/CandidateRankingsList.java
HEdingfield: Good idea; created #699 for this.

artoonie:
Looks good! Appreciate the logging messages too.

PR #701, opened 2023-06-13
support semantic versions with snapshots (e.g. 1.3.0-alpha)

### LINKED ISSUE #700: Support snapshot level info in semantic versioning
`isVersionNewer` should support formats like `1.3.0-alpha`

artoonie:
closes #700


src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -193,6 +194,22 @@ static void setup() {
...
+    assertFalse(isVersionNewer("1.2.0", "1.3.0"));
+    assertFalse(isVersionNewer("1.3.0", "1.3.1"));
+    assertTrue(isVersionNewer("1.3.1", "1.3.0"));
+
+    // Version comparison ignores the -alpha specifier
+    assertFalse(isVersionNewer("1.2.0-alpha", "1.3.0"));
+    assertTrue(isVersionNewer("1.4.0-alpha", "1.3.0"));
+    assertFalse(isVersionNewer("1.4.0-alpha", "1.4.0"));
+    assertFalse(isVersionNewer("1.4.0", "1.4.0-alpha"));
+  }
```
artoonie, line 211: this isn't really a tabulator test -- is there anywhere else unit tests might go? Is here okay or should I break out a second file?

src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -193,6 +194,22 @@ static void setup() {
...
+    assertFalse(isVersionNewer("1.3.0", "1.3.0"));
+    assertFalse(isVersionNewer("1.2.0", "1.3.0"));
+    assertFalse(isVersionNewer("1.3.0", "1.3.1"));
+    assertTrue(isVersionNewer("1.3.1", "1.3.0"));
+
+    // Version comparison ignores the -alpha specifier
+    assertFalse(isVersionNewer("1.2.0-alpha", "1.3.0"));
+    assertTrue(isVersionNewer("1.4.0-alpha", "1.3.0"));
+    assertFalse(isVersionNewer("1.4.0-alpha", "1.4.0"));
+    assertFalse(isVersionNewer("1.4.0", "1.4.0-alpha"));
```
HEdingfield, line 210: Please also add a check that makes sure:

`assertTrue(isVersionNewer("1.4.0-beta", "1.4.0-alpha"));`

And also curious what happens if a version is just a string like "foo"? Maybe we'd expect that to throw an exception?

src/test/java/network/brightspots/rcv/TabulatorTests.java
HEdingfield: I'm cool with kicking off a `ContestConfigMigrationTests` class and putting this in there if it's not too much trouble.

HEdingfield (commented):
Wow, nice find on that helper library!

artoonie:
> Wow, nice find on that helper library!

There are a lot with nicer APIs, but this one already exists in the codebase, so it's nice that we don't have to add any dependencies for it :)

src/test/java/network/brightspots/rcv/TabulatorTests.java
artoonie: Alas, it is not true. I've opened an Issue with Jackson: https://github.com/FasterXML/jackson-core/issues/1050

Currently, `alpha` and `beta` are considered the same version.

src/test/java/network/brightspots/rcv/ContestConfigTests.java
```diff
@@ -0,0 +1,62 @@
...
+    assertFalse(isVersionNewer("1.4.0-alpha", "1.4.0"));
+    assertFalse(isVersionNewer("1.4.0", "1.4.0-alpha"));
+
+    // -beta and -alpha are considered the same version
+    assertFalse(isVersionNewer("1.4.0-beta", "1.4.0-alpha"));
+    assertFalse(isVersionNewer("1.4.0-alpha", "1.4.0-beta"));
+
+    // Sane things with false versions
+    assertFalse(isVersionNewer("goober", "1.0.0"));
+    assertTrue(isVersionNewer("1.0.0", "goober"));
```
HEdingfield, line 54: Maybe we should rig it up so it throws an exception in either of these cases instead? I really don't know what we should expect here... :/ Maybe that's going too deep?

src/main/java/network/brightspots/rcv/ContestConfigMigration.java
```diff
@@ -17,61 +17,25 @@
...
-          isNewer = true;
-          break;
-        } else if (version2Num > version1Num) {
-          break;
-        }
-      }
-    }
-
-    return isNewer;
+  public static boolean isVersionNewer(String version1, String version2) {
```
HEdingfield, line 35: This is all kind of making me wonder if `checkConfigVersionMatchesApp` is actually doing anything besides logging?? It doesn't seem like it's stopping the code from running, so I'm a bit confused by this. Would you mind giving `checkConfigVersionMatchesApp` a sanity check?

src/test/java/network/brightspots/rcv/ContestConfigTests.java
```diff
@@ -0,0 +1,62 @@
...
+
+    // Version comparison ignores the -alpha specifier
+    assertFalse(isVersionNewer("1.2.0-alpha", "1.3.0"));
+    assertTrue(isVersionNewer("1.4.0-alpha", "1.3.0"));
+    assertFalse(isVersionNewer("1.4.0-alpha", "1.4.0"));
+    assertFalse(isVersionNewer("1.4.0", "1.4.0-alpha"));
+
+    // -beta and -alpha are considered the same version
+    assertFalse(isVersionNewer("1.4.0-beta", "1.4.0-alpha"));
+    assertFalse(isVersionNewer("1.4.0-alpha", "1.4.0-beta"));
```
HEdingfield, line 50: Another critical test we should add is:
```
assertFalse(isVersionNewer("1.4.0-alpha", "1.4.0"));
assertTrue(isVersionNewer("1.4.0", "1.4.0-alpha"));
```

src/test/java/network/brightspots/rcv/ContestConfigTests.java
```diff
@@ -0,0 +1,62 @@
...
+    assertFalse(isVersionNewer("1.3.0", "1.3.1"));
+    assertTrue(isVersionNewer("1.3.1", "1.3.0"));
+
+    // Version comparison ignores the -alpha specifier
+    assertFalse(isVersionNewer("1.2.0-alpha", "1.3.0"));
+    assertTrue(isVersionNewer("1.4.0-alpha", "1.3.0"));
+    assertFalse(isVersionNewer("1.4.0-alpha", "1.4.0"));
+    assertFalse(isVersionNewer("1.4.0", "1.4.0-alpha"));
+
+    // -beta and -alpha are considered the same version
```
HEdingfield, line 48: Please add a comment linking to your issue here as well (thanks for opening that).

I'm undecided on whether or not we should add bespoke code to handle this situation ourselves yet... maybe take a look at how `isVersionNewer` is used in practice and let me know what you think?

src/test/java/network/brightspots/rcv/ContestConfigTests.java
artoonie: `goober` turns into an Unknown Version object, and every versioned object is considered newer than Unknown Versions. This will allow TEST to be marked as older as any other version, which feels right? I'm also inclined to just do "whatever Jackson does, we do" rather than building our own scaffolding around it, since with enough changes, we might as well just create our own `class Version`.

src/main/java/network/brightspots/rcv/ContestConfigMigration.java
artoonie: It doesn't do anything besides log. Based on the comments, this isn't a problem in the GUI because the GUI tries to migrate versions without any care as to the existing version (which #706 should address). For now, all it does is spit out a SEVERE and continue tabulation.

src/test/java/network/brightspots/rcv/ContestConfigTests.java
artoonie: I'm inclined to leave as-is, since the version isn't really strongly used at all -- at least not yet. Once we implement #706, we might decide the current version is fine? But I don't feel strongly, it won't be hard to support this.

src/test/java/network/brightspots/rcv/ContestConfigTests.java
HEdingfield: Ok, I think that makes sense. 👍

src/main/java/network/brightspots/rcv/ContestConfigMigration.java
HEdingfield: Yeah, this is kind of weird, and I wonder if this code was removed unintentionally... in the past I _think_ we had it block tabulation if the config file version was newer than the tabulator version. Does that make sense to you?

Conversely, if the config file version is older than the tabulator version, I'd think attempting to run via the CLI should block it and say "you need to migrate this via the GUI". What do you think?

@tarheel @chughes297 do either of you have any idea why we apparently don't block tabulation in either of these cases?

src/test/java/network/brightspots/rcv/ContestConfigTests.java
```diff
@@ -0,0 +1,67 @@
...
+    // Partial versions are parsed correctly and have implicit zeros
+    assertTrue(isVersionNewer("1.0.1", "1.0"));
+    assertFalse(isVersionNewer("1.0.0", "1.0"));
+    assertFalse(isVersionNewer("1.0.0", "1"));
+    assertFalse(isVersionNewer("1", "1.0"));
+
+    // snapshot info is ignored, though it shouldn't be.
+    // The following tests check the current state, though we have an Issue open with
+    // jackson-core to address.
+    // https://github.com/FasterXML/jackson-core/issues/1050
```
HEdingfield, line 61: Could you please add one commented-out versions of the lines we're hoping to be able to change these to once your issue is resolved?

Same applies to many of the tests under this header above as well:

`// Version comparison ignores the -alpha specifier`

Maybe it's worth kind of grouping these two sections together more (looks like you have a couple duplicates up there anyway.

I'd say put all 6 non-duplicated tests between these two sections under the same header here, and add commented-out versions of what the values _should_ be after your issue is resolved.

artoonie: Good catch on the duplicate tests. Removed those.

Added a comment in this section saying the values should all flip from false to true once the issue is resolved.

HEdingfield: Not all should be true though; just pushed a commit to clarify.

src/main/java/network/brightspots/rcv/ContestConfigMigration.java
HEdingfield: This PR might be good to go, depending on what @chughes297 and / or @tarheel think about this comment.

tarheel: I'm not sure about the history, but I think I agree that 1) it should fail if the config is newer than the app and 2) it should  warn but continue and try to auto-migrate if the app is newer than the config.

HEdingfield: Right on, agreed. Opened #712 for this.

HEdingfield (approved):
Awesome job outsourcing this and submitting a PR for Jackson!

PR #704, opened 2023-06-14
editable tables

artoonie:
Has two major fixes from the previous iteration of Editable Tables:
1. Commits the edit when focus is lost (e.g. clicking another cell)
2. Ability to edit, save, then edit again 

The former led to the addition of a couple files which I need to heavily clean up and trim down.
The latter led to a bunch of code changes, along with many safety guardrails because a lot of magic happens with reflection behind-the-scenes and it was too easy to "miss" a required function, causing a silent failure where the UI would update but it wouldn't commit that update back to disk.

Alias editing happens as so:
image


HEdingfield:
Oh yeah, this one's gonna be gnaaaaaaarly. Props to you for picking it up! I'll keep an eye out for when you say it's ready for review.

artoonie:
I'll spend a little longer on aliases, I think it can be pretty straightforward with the semicolon-separated plan. Adding aliases will be one of the highest-priority adjustments, since the Auto Candidate Loader with Dominion loads codes as canonical names, so I expect election administrators will want Aliases working more than anything else.

HEdingfield:
Definitely agree, I'm just suggesting you break it up into two PRs since
this one is already going to be a challenge to review. Your call though.

artoonie:
This is ready for review!

HEdingfield:
Awesome! Just want to confirm that running a tabulation from the CLI still
WAI given the changes to RawContestConfig?

artoonie:
Hm, I've never tested the CLI interface, and it seems to hang indefinitely on `develop` for me. Not sure if that's a bug or my setup...probably my setup on WSL (since WSL acted funny in the past). I'll test on a mac later.

artoonie:
Confirmed that the CLI version still works.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -1130,35 +1131,50 @@ public LocalDate fromString(String string) {
...
+    // is cleaner anyway since it nicely disables and enables supported fields.
+    tableColumnCvrProvider.setEditable(false);
+
     tableViewCvrFiles.getSelectionModel().setSelectionMode(SelectionMode.MULTIPLE);
-    tableViewCvrFiles.setEditable(false);
+    tableViewCvrFiles.setEditable(true);
+
+    EditableColumn[] candidateStringColumnsAndProperties = new EditableColumn[]{
+        new EditableColumnString(tableColumnCandidateName, "name"),
+        new EditableColumnList(tableColumnCandidateAliases, "aliases"),
```
HEdingfield, line 1162: We talked about having this work as a pop-up in [this thread](https://github.com/BrightSpots/rcv/pull/639/files#r1141150539). Right now, there's zero guidance to the user that they need to put semicolons in there if they expect to break out separate aliases.

A seriously confusing UX is, if they enter aliases with semicolons in the box above the table ( (e.g. "test1;test2" or even "test1; test2"), it doesn't treat them as separate aliases (you can verify this because validation passes); the expectation there is that it's supposed to be entered with a line break.

However, if they then double-click the cell and press enter, it will subsequently convert them into separate aliases.

src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -96,162 +100,233 @@ public static class CvrSource {
...
 
-    public void setProvider(String provider) {
-      this.provider = provider;
+    /**
+     * Set the provider by either its GUI label or its internal label.
+     * Supporting both means we don't have to translate between internal and GUI labels
+     * each time we serialize/deserialize in the table.
+     *
+     * @param providerString The provider's internal or GUI label.
+     */
```
HEdingfield, line 183: I noticed that in the GUI table, it now shows the non-user-friendly versions, like "ess" and "clearBallot" instead of "ES&S" and "Clear Ballot". Is that an artifact of this? Can we revert to that more user-friendly behavior?

src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -96,162 +100,233 @@ public static class CvrSource {
...
     }
 
-    public void setTreatBlankAsUndeclaredWriteIn(boolean treatBlankAsUndeclaredWriteIn) {
-      this.treatBlankAsUndeclaredWriteIn = treatBlankAsUndeclaredWriteIn;
+    public void setTreatBlankAsUndeclaredWriteIn(Boolean treatBlankAsUndeclaredWriteIn) {
+      this.treatBlankAsUndeclaredWriteIn.set(treatBlankAsUndeclaredWriteIn);
+    }
+
+    /**
+     * The following properties are marked as unused by an editor, but
```
HEdingfield, line 229: A little confused by this comment. What is "an editor" referring to?

src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -271,38 +346,41 @@ private void setCode(String code) {
...
-        name = name.trim();
+        name.setValue(getName().trim());
       }
       if (aliases != null) {
         aliases.replaceAll(s -> s.trim());
       }
     }
+
+    /**
+     * The following properties are marked as unused by an editor, but
```
HEdingfield, line 369: Same comment here about not understanding this comment.

HEdingfield (changes requested):
Solid work so far. This one is pretty tough.

src/main/java/network/brightspots/rcv/RawContestConfig.java
artoonie: IntelliJ and vim -- an IDE with some code analysis. I can replace with "IDE"

HEdingfield: Ahh gotcha, yeah, that would make more sense. "The following properties might be flagged as unused by an IDE, but..."

src/main/java/network/brightspots/rcv/EditableTableCell.java
```diff
@@ -0,0 +1,216 @@
...
+
+        // requesting focus so that key input can immediately go into the
+        // TextField (see RT-28132)
+        textField.requestFocus();
+      }
+    }
+  }
+
+  @Override
+  public void commitEdit(T newValue) {
```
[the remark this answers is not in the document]
artoonie: Hm, that feels pretty strange too -- if you:
1. Begin edit
2. Hit tabulate

Should it cancel or commit your edit? I can see arguments both ways. I've opted to just disable the menu bar and tab bar until the edit is committed or canceled.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -1256,6 +1272,38 @@ private void setUpHintsForTab(Tab tab, String hintsFilename) {
...
+          + editableColumn.propertyName.substring(0, 1).toUpperCase()
+          + editableColumn.propertyName.substring(1);
+      String property = editableColumn.propertyName + "Property";
+
+      try {
+        rowType.getMethod(getter);
+        rowType.getMethod(setter, editableColumn.propertyType());
+        rowType.getMethod(property);
+      } catch (NoSuchMethodException e) {
+        Logger.severe("Could not find required function %s", e.getMessage());
```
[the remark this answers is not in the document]
artoonie: It's actionable! `EDT SEVERE: Could not find required function network.brightspots.rcv.RawContestConfig$CvrSource.getXYZ()`

src/main/java/network/brightspots/rcv/RawContestConfig.java
artoonie: done

artoonie:
All comments addressed! Primary change since the last review is the creation of `EditableTableCellPopup`.

src/main/java/network/brightspots/rcv/EditableTableCell.java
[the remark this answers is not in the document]
HEdingfield: Yeah, that's a good point, and I like the direction you went. Could you modify this further to disable everything outside the table? For example, if I edit a cell, I can still click the auto-load button, which re-enables the menu at the top.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Nice, I think that's a big improvement! One thing I noticed that's still a bit off is if you end a line with a semi-colon... it kind of just vanishes after you commit it.

HEdingfield (changes requested):
Almost there!

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: Good catch -- that's an issue with the existing aliases code. Split that out just to make sure: https://github.com/BrightSpots/rcv/pull/726

src/main/java/network/brightspots/rcv/EditableTableCell.java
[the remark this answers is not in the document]
artoonie: Done -- there's not an existing clean way to grab everything outside of the table since many of the buttons aren't hooked up to variables in the class, so the best way is to traverse through the scene graph and find things to disable.

Let me know if you're okay with how I did it!

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: Scratch my previous [deleted] comment.

It's an issue in the existing code too, when creating aliases, and when editing. Fixed for both -- I had the regex eat all non-word characters `\W` instead of all non-whitespace characters `\s`

src/main/java/network/brightspots/rcv/EditableTableCell.java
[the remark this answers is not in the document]
HEdingfield: On the CVR Files tab, it doesn't disable the buttons there. There's another edge case where you can edit a cell and keep it open, then go through and add another CVR. As soon as you click add or clear, it commits the edit down below (so it's making the decision you mentioned above that's kind of ambiguous).

Can we just disable an entire box and everything in it (in this case the "Add CVR File" box), including all buttons and inputs? If not, I guess we just need more bespoke code targeting the buttons on this pane as well.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Seems to work well now, nice work.

HEdingfield (changes requested):

src/main/java/network/brightspots/rcv/EditableTableCell.java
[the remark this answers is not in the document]
artoonie: I cannot disable the entire panel since that would disable the edit button too. I think I have an idea for a cleaner solution -- I might be able to tag CSS selectors for everything that needs to be disabled, and then just loop over that selector. Will look into this.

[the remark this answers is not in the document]
artoonie: Done -- I now use CSS style selectors to identify what to disable.

HEdingfield:
Great! One last request: could you put a small note at the top of each
table saying something like "Double-click a cell to edit, and press enter
to save or escape to cancel"?

I know they can also click out to save, but I think it'd be nice to promote
the more straightforward UX (and also signal to users that they can edit
cells again).

artoonie:
Done. Adjusted text to be short and sweet: 
image

src/main/java/network/brightspots/rcv/Utils.java
```diff
@@ -83,4 +83,8 @@ static String getUserName() {
     }
     return user;
   }
+
+  static String[] splitByNewline(String s) {
+    return s.split("\\s*\\r?\\n\\s*");
```
HEdingfield, line 88: One last quirk with this: if you make the first line or first few lines blank, it'll count it as a null alias (happens in both the original alias box and in the edit pop-up).

src/main/java/network/brightspots/rcv/EditableTableCell.java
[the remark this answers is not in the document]
HEdingfield: This is a great solution, and exactly what I was hoping for!

The only other corner case I can find now is, if you're editing and you click the X to close the program, it won't commit the edit and it'll end up effectively ignoring the edit if you save it. Probably fine to leave it this way though.

HEdingfield (approved):
Amazing job!! I just tweaked the helper text a bit and found one last quirk to deal with, but pre-approving.

src/main/java/network/brightspots/rcv/Utils.java
artoonie: Fixed!

PR #708, opened 2023-06-16
consolidate all duplicate CVRs

### LINKED ISSUE #692: Create a common test suite for shared CVRs
See comment https://github.com/BrightSpots/rcv/pull/650#discussion_r1224591566

In the `test_data` directory, create a `common` directory in which multiple tests can share a single CVR, to avoid unnecessary duplication of tens of thousands of lines of CVR data.

artoonie:
closes #692 

Note: one of the original goals was to specifically get rid of the duplicate files in #650, but I opted to skip that: while most files are the same, one test removes the Precinct file, and the other doesn't, meaning the test sets aren't identical. I'm sure we could solve that with some added complexity, but I'm not sure that complexity is worth it.


HEdingfield (changes requested):
Placing a hold on this until we get the parent branch merged. Will review once that's done.

src/test/resources/network/brightspots/rcv/test_data/minneapolis_multi_seat_threshold/minneapolis_multi_seat_threshold_config.json
```diff
@@ -11,7 +11,7 @@
   },
   "cvrFileSources": [
     {
-    "filePath" : "minneapolis_multi_seat_threshold_cvr.xlsx",
+    "filePath" : "../_shared/minneapolis_multi_seat_cvr.xlsx",
```
HEdingfield, line 14: It seems like this one might not actually be shared between any other tests? Are we pre-emptively moving CVRs into the `_shared` folder, or was this a mistake?

src/test/resources/network/brightspots/rcv/test_data/skip_to_next_test/skip_to_next_test_config.json
```diff
@@ -10,7 +10,7 @@
     "generateCdfJson" : false
   },
   "cvrFileSources" : [ {
-    "filePath" : "skip_to_next_test_cvr.xlsx",
+    "filePath" : "../_shared/skipped_ranking_cvr.xlsx",
```
HEdingfield, line 13: Same question on this one -- seems like this isn't used in any other tests.

src/test/resources/network/brightspots/rcv/test_data/test_set_2_overvote_skip_to_next/test_set_2_overvote_skip_to_next_cvr.xlsx
HEdingfield, line 1: Kind of weird that this wasn't used at all... was this just junk all along, or could it possibly serve some value?

src/test/resources/network/brightspots/rcv/test_data/test_set_multi_winner_fractional_threshold/test_set_multi_winner_fractional_threshold_cvr.xlsx
HEdingfield, line 1: Same question on this... wondering why this was around if it apparently wasn't being used?

HEdingfield (commented):
This is great clean-up! Just a few questions before merging.

src/test/resources/network/brightspots/rcv/test_data/test_set_multi_winner_fractional_threshold/test_set_multi_winner_fractional_threshold_cvr.xlsx
artoonie: Hummm -- for all four comments here, the two *unused* files were duplicates, which is why the two *unshared* CVRs were moved to `_shared`.

In particular, `../_shared/skipped_ranking_cvr.xlsx` == `test_set_2_overvote_skip_to_next_cvr.xlsx` and 
`../_shared/minneapolis_multi_seat_cvr.xlsx ` == `../minneapolis_multi_seat_threshold/minneapolis_multi_seat_cvr.xlsx`.

It seems as though the xlsx files are indeed never used -- that they've been converted to CDF already. So I removed the duplicate files and placed them in shared, but one of the duplicate files were never used.

I have two options:
1. Delete the CDF CVR and point to the shared XLSX CVR (which seems to be the ground truth anyway), or
2. Leave these XLSX files deleted and move the two files in `_shared` out of the shared directory.

I think (1) is cleaner -- let's use the source files? Unless there was some reason why they weren't being used in the first place.

HEdingfield (approved):
Awesome job! Left a few last nit comments, but pre-approving.

PR #709, opened 2023-06-16
track session count to avoid race conditions in tests

artoonie:
While this hasn't happened with milliseconds appended to the output filenames, it's theoretically possible:

1. Output files now go into the `_shared` directory for some tests
2. If two tests that output the same file run within the same millisecond, they'll have the same output filename, and one will overwrite the other
3. This could lead to flaky tests in the future

I seems exceedingly rare for two tests to run at the same millisecond unless we're running on multiple threads -- and in that case, this code won't work anyway, since `sessionCount` isn't thread-safe. Still, wanted to open this PR to start a discussion: is this something we should handle?

### DIFF
--- a/src/main/java/network/brightspots/rcv/TabulatorSession.java
+++ b/src/main/java/network/brightspots/rcv/TabulatorSession.java
@@ -42,6 +42,9 @@ import network.brightspots.rcv.Tabulator.TabulationAbortedException;
 
 @SuppressWarnings("RedundantSuppression")
 class TabulatorSession {
+  // In order to allow multiple tests to write to the same output file at the same millisecond,
+  // we track the session count and append it to the timestamp string.
+  private static int sessionCount = 0;
 
   private final String configPath;
   private final String timestampString;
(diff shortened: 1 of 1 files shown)


HEdingfield:
Are there any downsides to doing this?

artoonie:
Just a touch of complexity. Tests will need to update to take in the session count and timestamp, not just session count. Filenames will be longer. Static variables are kind of yucky, and not thread-safe. This is a mediocre solution to a problem that may not exist.

Still, flaky tests are frustrating, and this is very likely to prevent them.


I'd recommend punting on this unless we discover that the tests are actually flaky.

HEdingfield:
Gotcha, sounds good. Let's plan on keeping this PR around until we're ready to release 1.4.0. If we haven't seen any flakiness by then, we can go ahead and close.

HEdingfield:
Actually, out of curiosity, does this issue only exist because we're now potentially outputting to a common `output` folder for multiple tests in the `_shared` directory?

Would this be a non-issue if we continued to output to an `output` directory in the test directory itself instead of in a sub of the `_shared` folder? I like this design better anyway, as mentioned in https://github.com/BrightSpots/rcv/pull/707#discussion_r1233352053.

artoonie:
Closing this in favor of #707 (allowing duplicate config files)

Issue #633, opened 2023-02-17
Updated Results Output


rrojas350:
RCTab's result file could use some improvement, specifically a breakdown of inactive ballots. See this mock-up for an example: 

[RCTab Results Report Draft](https://docs.google.com/spreadsheets/d/1YPScnGZnQvJw6klg6CiIXRBCN4v9SEaa_pf7Hsj-ITk/edit?usp=sharing)

rrojas350:
Column C and D have additional information as well. Column C has a percentage of the vote in that round and D has the amount of votes transferred over. Ideally each round would have both the percentage and votes transferred column.

chughes297:
And the creation of the undervote category in B14. This is ballots that have no ranking at all in a contest. Lots of statute coming online now that treats undervotes as a separate category from ballots that have any kind of marking/ranking in an RCV contest (even if it's just an overvote in the first round).

ggilber1:
Also, Column F should be "% of Vote" and G "Transfered"

ggilber1:
Not sure exactly what "Truncated Ballots" are. ?? @rrojas350

chughes297:
@ggilber1 Truncated ballot is a ballot that does not use all of its rankings and goes inactive when it runs out of those rankings. Open to other terms!!

ggilber1:
I would include "truncated" with the "Exhausted Choices" category.  I've
never seen the two separated and don't think the increased complexity is
worth the information gained.

George Gilbert
***@***.***
336-906-0047

chughes297:
Also not sure if this is clear but I'm assuming we'd update the .json file with this sort of info as well, I just didn't make a mockup of that format.

artoonie:
Would this be a good time to include a per-round threshold too, since we're making this update?
I'd like to also add a version number to the output summary files.

chughes297:
> Would this be a good time to include a per-round threshold too, since we're making this update? I'd like to also add a version number to the output summary files.

Yes! That's on row 26. (bottom of candidate totals). Given your question, though, is that easy for a reader to miss? 

Great idea on adding version number too.

artoonie:
Ah, I see it now -- I saw the original threshold on Line 7 and assumed there was no other. The top one on Line 7 is useful for a quick glance, so let's keep both.

I'm making some additional minor adjustments:
[...]

PR #710, opened 2023-06-16
Add "Total Votes Cast" to results summary file

artoonie:
When tabulating by precinct, "totalVotesCast" is the number of ballots in that precinct.

If this looks good, I can go ahead and update all expected summary JSONs in the tests. Didn't want to do that prematurely in case `numBallots` isn't the right value here.

I don't see any tests that check the summary CSV files -- I'll take a closer look after I updated all JSONs, and if everything is passing, I'll add a test that does compare the summary CSVs.

I see a strong dependency between this PR and #633, so I'll do both in the same PR to avoid updating the config files twice.


artoonie:
Closing this; will merge as part of #711

PR #711, opened 2023-06-19
Issue 633 (New CSV Format) -- Part 2: Scaffolding

### LINKED ISSUE #633: Updated Results Output

### LINKED ISSUE #678: Add "Total votes cast" to summary results files.
This specific information is included in the results report suggested in #633. This is a much smaller request to just add a single additional piece of info with the total votes counted in the contest (all active + inactive ballots + undervoted ballots). I'm not sure if we want to do this separately because it's very simple or just do the whole results report overhaul.

artoonie:
Makes progress on #633, #678 :

Build the scaffolding needed to write the undervote breakdown in the output summary files

This is part 2 of a 3-part PR Chain:

1. Add tests to ensure no CSVs change [#713]
2. Update tabulation code, ensuring no adjustments to results writing [#711]
3. Update results writing code, ensuring no adjustments to tabulation code [#714]

### DIFF
--- a/src/main/java/network/brightspots/rcv/ResultsWriter.java
+++ b/src/main/java/network/brightspots/rcv/ResultsWriter.java
@@ -20,6 +20,7 @@
 package network.brightspots.rcv;
 
 import static java.util.Map.entry;
+import static network.brightspots.rcv.CastVoteRecord.BallotStatus;
 import static network.brightspots.rcv.Utils.isNullOrBlank;
 
 import com.fasterxml.jackson.core.util.DefaultPrettyPrinter;
@@ -32,6 +33,7 @@ import java.io.BufferedWriter;
 import java.io.File;
 import java.io.IOException;
 import java.math.BigDecimal;
+import java.math.MathContext;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -267,34 +269,21 @@ class ResultsWriter {
   // param: precinctTallyTransfers is a map from precinct to tally transfers for that precinct
   // param: numBallotsByPrecinct is the total count of ballots per precinct
   void generatePrecinctSummaryFiles(
-      Map<String, Map<Integer, Map<String, BigDecimal>>> precinctRoundTallies,
+      Map<String, Map<Integer, RoundTally>> precinctRoundTallies,
       Map<String, TallyTransfers> precinctTallyTransfers,
-      Map<String, Integer> numBallotsByPrecinct,
       Map<String, Integer> numUndervotesByPrecinct)
       throws IOException {
     Set<String> filenames = new HashSet<>();
     for (var entry : precinctRoundTallies.entrySet()) {
       String precinct = entry.getKey();
-      Map<Integer, Map<String, BigDecimal>> roundTallies = entry.getValue();
+      Map<Integer, RoundTally> roundTallies = entry.getValue();
       String precinctFileString = getPrecinctFileString(precinct, filenames);
       String outputPath =
           getOutputFilePathFromInstance(String.format("%s_precinct_summary", precinctFileString));
-      Integer numBallotsObj = numBallotsByPrecinct.get(precinct);
-      if (numBallotsObj == null) {
-        Logger.warning("No ballots found in precinct \"%s\"", precinct, numBallotsByPrecinct);
-        numBallotsObj = 0;
-      }
-      Integer numUndervotesObj = numUndervotesByPrecinct.get(precinct);
-      if (numUndervotesObj == null) {
-        Logger.warning("No undervote tally found in precinct \"%s\"",
-            precinct, numUndervotesByPrecinct);
-        numBallotsObj = 0;
-      }
-      int numBallots = numBallotsObj;
-      int numUndervotes = numUndervotesObj == null ? 0 : numUndervotesObj;
-      generateSummarySpreadsheet(roundTallies, numBallots, numUndervotes, precinct, outputPath);
+      Integer numUndervotes = numUndervotesByPrecinct.get(precinct);
+      generateSummarySpreadsheet(roundTallies, numUndervotes, precinct, outputPath);
       generateSummaryJson(roundTallies, precinctTallyTransfers.get(precinct),
-            numBallots, precinct, outputPath);
+            precinct, outputPath);
     }
   }
 
@@ -303,29 +292,13 @@ class ResultsWriter {
   // param: ballotStatusTallies is the round-by-count count of ballot statuses
   // param: precinct indicates which precinct we're reporting results for (null means all)
   private void generateSummarySpreadsheet(
-      Map<Integer, Map<String, BigDecimal>> roundTallies,
-      Map<Integer, Map<BallotStatus, BigDecimal>> ballotStatusTallies,
-      int numBallots,
-      int numUndervotes,
+      Map<Integer, RoundTally> roundTallies,
+      Integer numUndervotes,
       String precinct,
       String outputPath)
       throws IOException {
     String csvPath = outputPath + ".csv";
     Logger.info("Generating summary spreadsheet: %s...", csvPath);
-
-    // totalActiveVotesPerRound is a map of round to active votes in each round
-    Map<Integer, BigDecimal> totalActiveVotesPerRound = new HashMap<>();
-    for (int round = 1; round <= numRounds; round++) {
-      // tally is map of candidate to tally for the current round
-      Map<String, BigDecimal> tallies = roundTallies.get(round);
-      // total will contain total votes for all candidates in this round
-      // this is used for calculating other derived data
-      BigDecimal total = BigDecimal.ZERO;
-      for (BigDecimal tally : tallies.values()) {
-        total = total.add(tally);
-      }
-      totalActiveVotesPerRound.put(round, total);
-    }
 
     CSVPrinter csvPrinter;
     try {
@@ -339,7 +312,7 @@ class ResultsWriter {
     }
 
     addContestInformationRows(csvPrinter, precinct);
-    addContestSummaryRows(csvPrinter, numBallots, numUndervotes);
+    addContestSummaryRows(csvPrinter, roundTallies.get(1), numUndervotes);
     csvPrinter.print("Rounds");
     for (int round = 1; round <= numRounds; round++) {
       csvPrinter.print(String.format("Round %d Votes", round));
@@ -354,15 +327,14 @@ class ResultsWriter {
     }
 
     // Get all candidates sorted by their first round tally. This determines the display order.
-    Map<String, BigDecimal> firstRoundTally = roundTallies.get(1);
-    List<String> sortedCandidates = sortCandidatesByTally(firstRoundTally);
+    List<String> sortedCandidates = roundTallies.get(1).getSortedCandidatesByTally();
 
     // For each candidate: for each round: output total votes
     for (String candidate : sortedCandidates) {
       String candidateDisplayName = config.getNameForCandidate(candidate);
       csvPrinter.print(candidateDisplayName);
       for (int round = 1; round <= numRounds; round++) {
-        BigDecimal thisRoundTally = roundTallies.get(round).get(candidate);
+        BigDecimal thisRoundTally = roundTallies.get(round).getCandidateTally(candidate);
         // not all candidates may have a tally in every round
         if (thisRoundTally == null) {
           thisRoundTally = BigDecimal.ZERO;
@@ -372,15 +344,20 @@ class ResultsWriter {
         csvPrinter.print(thisRoundTally);
 
         // Vote %
-        csvPrinter.print(thisRoundTally / totalActiveVotesPerRound.get(round));
+        BigDecimal activeBallots = roundTallies.get(round).numActiveBallots();
+        if (activeBallots != BigDecimal.ZERO) {
+          csvPrinter.print(thisRoundTally.divide(activeBallots, MathContext.DECIMAL32));
+        } else {
+          csvPrinter.print("");
+        }
 
         // Transfer
         if (round < numRounds) {
-          BigDecimal nextRoundTally = roundTallies.get(round + 1).get(candidate);
-          if (thisRoundTally == null) {
+          BigDecimal nextRoundTally = roundTallies.get(round + 1).getCandidateTally(candidate);
+          if (nextRoundTally == null) {
             nextRoundTally = BigDecimal.ZERO;
           }
-          csvPrinter.print(nextRoundTally - thisRoundTally);
+          csvPrinter.print(nextRoundTally.subtract(thisRoundTally));
         } else {
           csvPrinter.print(0);
         }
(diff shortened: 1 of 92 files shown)


artoonie:
@HEdingfield can I get an early review on the architecture here?

In order to support the new CSV and JSON formats, we need to track the ballot statuses for each round.

Instead of a `Map<Integer, Map<String, BigDecimal>>` to track the per-round vote counts for each candidate, I refactored to use `Map<Integer, RoundTally>`. `RoundTally` includes both a `Map<String, BigDecimal>` as before, but also includes a `Map<BallotStatus, BigDecimal>` to track how ballots became inactive on each round.

The math isn't accurate yet (known issue: `ACTIVE` ballots are only properly counted in Round 1; other issues may exist) -- but was hoping to get an architecture review before I go too far down this path.

One decision of particular note: the previous structure was both written to and read from at various times, and assumptions were made about whether the content was still being written or whether it was finalized. I opted to make that explicit: when the new structure is being written to, it cannot be read from, and vice versa. This changes no functionality, but it is a nice safety net. An exception is made for surplus transfers, where both reading and writing must occur.

Since this is a pretty large change, in order to ensure the tabulation logic doesn't change at all, my plan is:
1. Add tests for the current CSV format (it seems none exist?)
2. Complete this PR, ensuring no changes to both the JSONs and CSVs
3. Chain a second PR to this one that writes the new test files and updates test files

tarheel:
I'll try to look through this tomorrow also.

HEdingfield:
This one is pretty far out of my wheelhouse, so absolutely defer to @tarheel's review on this one!

I think your steps 1 through 3 seem like a good way to approach it. 👍

tarheel (commented):
I know this is still in progress, but I made a few comments inline anyway; apologies if they're flagging things that you're already planning to change. I didn't look closely at most of the code, though.

In terms of the general approach... I like the idea of creating a new `RoundTally` class so you can keep track of the different statuses. I haven't thought hard about whether there might be a way to achieve this with a smaller memory footprint, but since the main change is just that you're changing a boolean to an enum in the CVR class, I don't know if that makes any difference in Java, and if it does it's definitely not meaningful.

One thing I'm not sure about from my brief reading of this PR is whether you're storing enough information to be able to handle fractional vote transfers properly. I think maybe it works the way it is? The fractions of a ballot that have been allocated will be reflected in those candidates' totals, and then when the ballot is exhausted, its remaining fractional value will go into one of the inactive buckets, right?

src/main/java/network/brightspots/rcv/RoundTally.java
```diff
@@ -0,0 +1,175 @@
...
+    this.roundNumber = roundNumber;
+    candidateTallies = new HashMap<>();
+    candidateNames.forEach((String candidateName) -> {
+      candidateTallies.put(candidateName, BigDecimal.ZERO);
+    });
+
+    ballotStatusTallies = new HashMap<>();
+    for (CastVoteRecord.StatusForRound statusForRound : CastVoteRecord.StatusForRound.values()) {
+      ballotStatusTallies.put(statusForRound, BigDecimal.ZERO);
+    }
```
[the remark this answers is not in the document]
artoonie: `Stream` doesn't allow a `for item : list` loop, it requires a `.forEach`

src/main/java/network/brightspots/rcv/RoundTally.java
```diff
@@ -0,0 +1,175 @@
...
+  int getRoundNumber() {
+    return roundNumber;
+  }
+
+  // Call this to prevent future modifications to the round, and allow retrieval of data
+  // from the round.
+  void lockInRound() {
+    ensureNotFinalized();
+    isFinalized = true;
+    this.numBallots = countBallots();
```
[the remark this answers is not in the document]
artoonie: Just not cleaned up yet -- I moved this from a constructor where `this.` is more standard.

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -58,9 +59,12 @@ class Tabulator {
...
   // e.g. roundTallies[1] contains a map of all candidate ID -> votes for each candidate in round 1
   // this structure is computed over the course of tabulation
-  private final Map<Integer, Map<String, BigDecimal>> roundTallies = new HashMap<>();
+  private final Map<Integer, RoundTally> roundTallies = new HashMap<>();
   // precinctRoundTallies is a map from precinct to roundTallies for that precinct
-  private final Map<String, Map<Integer, Map<String, BigDecimal>>> precinctRoundTallies =
+  private final Map<String, Map<Integer, RoundTally>> precinctRoundTallies =
+      new HashMap<>();
+  // precinctBallotStatusTallies is a map from precinct to roundTallies for that precinct
+  private final Map<String, Map<Integer, RoundTally>> precinctBallotStatusTallies =
```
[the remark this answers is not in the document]
artoonie: Unused! (A previous iteration of the code had two separate maps instead of using `RoundTally`.)

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -296,9 +300,14 @@ private void logSummaryInfo() {
...
-  private void updatePastWinnerTallies() {
-    Map<String, BigDecimal> roundTally = roundTallies.get(currentRound);
-    Map<String, BigDecimal> previousRoundTally = roundTallies.get(currentRound - 1);
+  private void updateWinnerTallies() {
+    RoundTally roundTally = roundTallies.get(currentRound);
+    RoundTally previousRoundTally = roundTallies.get(currentRound - 1);
+    roundTally.unlockForSurplusCalculation();
+    if (previousRoundTally != null) {
+      previousRoundTally.unlockForSurplusCalculation();
+    }
```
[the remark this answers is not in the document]
artoonie: This is one of the pieces I'd specifically like feedback on:

"The previous structure was both written to and read from at various times, and assumptions were made about whether the content was still being written or whether it was finalized. I opted to make that explicit: when the new structure is being written to, it cannot be read from, and vice versa. This changes no functionality, but it is a nice safety net. An exception is made for surplus transfers, where both reading and writing must occur."

I can remove all the lock/unlocks and checks for it if it feels too messy, but the safety net feels kind of nice: make sure you're not computing a threshold before all votes have been tallied, for example.

artoonie:
Re: fractional votes. Yes, it should work that way, and we'll validate by ensuring that this PR doesn't change anything in the output JSONs or CSVs.

src/main/java/network/brightspots/rcv/Tabulator.java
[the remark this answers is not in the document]
tarheel: Yeah, I think this sanity check is reasonable. The part I don't understand is why you're unlocking `previousRoundTally` even though it's not being updated.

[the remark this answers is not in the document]
artoonie: Ah, gotcha. You're right, it doesn't need to be unlocked, I just unlocked everything in the function.

src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -18,6 +18,7 @@
 
 package network.brightspots.rcv;
 
+import static org.apache.commons.io.FileUtils.copyFile;
```
artoonie, line 21: [note: this unused import will be removed when #714 is merged]

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -889,13 +914,14 @@ private OvervoteDecision getOvervoteDecision(CandidatesAtRanking candidates)
   //  logs the results to audit log
   //  update tallyTransfers counts
   private void recordSelectionForCastVoteRecord(
-      CastVoteRecord cvr, int currentRound,
-      String selectedCandidate, String outcomeDescription) throws TabulationAbortedException {
+      CastVoteRecord cvr, RoundTally currentRound,
```
tarheel, line 917: I would rename `currentRound` to `currentRoundTally` for clarity

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -918,29 +944,52 @@ private void recordSelectionForCastVoteRecord(
...
+          outcomeDescription = "underote" + additionalLogText;
+      case INACTIVE_BY_OVERVOTE ->
+          outcomeDescription = "overvote" + additionalLogText;
+      case INACTIVE_BY_SKIPPED_RANKING ->
+          outcomeDescription = "exhausted by skipped ranking" + additionalLogText;
+      case INACTIVE_BY_REPEATED_RANKING ->
+          outcomeDescription = "duplicate candidate" + additionalLogText;
+      case INACTIVE_BY_EXHAUSTED_CHOICES ->
+          outcomeDescription = "no continuing candidate" + additionalLogText;
+      // When we encounter a continuing candidate, we record it with this string in the output logs
```
tarheel, line 968: I'm not sure what this comment is referring to

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -918,29 +944,52 @@ private void recordSelectionForCastVoteRecord(
...
+      default ->
+        throw new RuntimeException("Unexpected ballot status: " + statusForRound);
```
tarheel, line 970: this is basically an assertion, right? I think you guys changed our assertions to throws in a recent PR, but I think in cases where we're trying to catch a bug in our code vs. handling bad input data, we should at least make that clear with an inline comment (or in the exception message)

src/main/java/network/brightspots/rcv/RoundTally.java
```diff
@@ -0,0 +1,220 @@
...
+  void setWinningThreshold(BigDecimal winningThreshold) {
+    this.winningThreshold = winningThreshold;
+  }
+
+  // Gets the winning threshold for this round.
+  BigDecimal getWinningThreshold() {
+    return winningThreshold;
+  }
+
+  // Adds to the votes for this candidate
```
tarheel, line 117: fix

src/main/java/network/brightspots/rcv/RoundTally.java
```diff
@@ -0,0 +1,220 @@
...
+  BigDecimal addInactiveBallot(StatusForRound statusForRound, BigDecimal value) {
+    ensureNotFinalized();
+    if (statusForRound == StatusForRound.ACTIVE) {
+      throw new RuntimeException("Cannot add an active ballot as inactive");
+    }
+    return addBallotWithStatus(statusForRound, value);
+  }
+
+  // Adds to the votes for this candidate
```
tarheel, line 126: fix

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -1084,40 +1144,34 @@ && isCandidateContinuing(cvr.getCurrentRecipientOfVote())) {
...
-        tally.put(candidateName, BigDecimal.ZERO);
-      }
-    }
-    return tally;
+  private RoundTally getNewTally(int roundNumber) {
+    return new RoundTally(roundNumber, candidateNames.stream().filter(this::isCandidateContinuing));
   }
 
   // add a vote (or fractional share of a vote) to a tally
   private void incrementTally(
```
tarheel, line 1164: no longer seems useful to have this helper method

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -918,29 +944,52 @@ private void recordSelectionForCastVoteRecord(
...
+    if (statusForRound != StatusForRound.ACTIVE) {
+      currentRound.addInactiveBallot(statusForRound, cvr.getFractionalTransferValue());
+    }
+
+    String outcomeDescription;
+    switch (statusForRound) {
+      case ACTIVE ->
+          outcomeDescription = selectedCandidate;
+      case INACTIVE_BY_UNDERVOTE ->
+          outcomeDescription = "underote" + additionalLogText;
```
tarheel, line 959: typo ("underote")

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -267,54 +269,37 @@ ResultsWriter setTimestampString(String timestampString) {
...
 
   // create a summary spreadsheet .csv file
   // param: roundTallies is the round-by-count count of votes per candidate
+  // param: ballotStatusTallies is the round-by-count count of ballot statuses
   // param: precinct indicates which precinct we're reporting results for (null means all)
   private void generateSummarySpreadsheet(
-      Map<Integer, Map<String, BigDecimal>> roundTallies,
-      int numBallots,
+      Map<Integer, RoundTally> roundTallies,
+      Integer numUndervotes,
```
tarheel, line 296: So has the code been using the term "undervote" inaccurately? It looks like now you're considering a ballot to be an undervote only if it doesn't contain any rankings at all. And I guess the idea is that these won't be included in the first round's results, and we'll instead present them as a single independent number? In the past I think we've also used this term to refer to cases in which a ballot skipped too many rankings.

If your definition is the right one, there are probably some more places in the code (comments, at least?) that need updating.

tarheel (commented):
Looks great! Just some nits and minor questions inline.

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -340,37 +324,38 @@ private void generateSummarySpreadsheet(
...
-          new BigDecimal(numBallots).subtract(totalActiveVotesPerRound.get(round));
-
-      if (precinct == null) {
-        // We don't have the concept of residual surplus at the precinct level (see comment below),
-        // so we'll just incorporate that part (if any) into the inactive count.
-        thisRoundInactive = thisRoundInactive.subtract(roundToResidualSurplus.get(round));
+      BigDecimal inactiveBallots = roundTallies.get(round).numInactiveBallots();
+
+      // The previous method of calculating inactive ballots was to sum values that included
+      // active ballots, leading to additional but unnecessary precision.
```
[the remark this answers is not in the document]
artoonie: Basically, this is needed to ensure the tests have the exact same text.
This turns some values from `2` into `2.000` in the output CSVs, which was necessary to match the old math exactly.

This block of code gets removed in the next PR -- it's purely here to ensure the tests are character-for-character the same.

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -184,15 +183,18 @@ Set<String> tabulate() throws TabulationAbortedException {
           || currentRound == 1;
       }
       if (shouldRecomputeThreshold) {
-        setWinningThreshold(currentRoundCandidateToTally, config.getMinimumVoteThreshold());
+        setWinningThreshold(currentRoundTally, config.getMinimumVoteThreshold());
+      } else {
+        BigDecimal lastRoundThreshold = roundTallies.get(currentRound - 1).getWinningThreshold();
+        currentRoundTally.setWinningThreshold(lastRoundThreshold);
```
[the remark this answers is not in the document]
artoonie: Yes, I should have done that here, but missed it. It's done in the next PR in the chain.

src/main/java/network/brightspots/rcv/ResultsWriter.java
artoonie: You're right -- this is particularly confusing in the GUI, where there's an explicit undervote label, and that's used to label skipped rankings. To fix that, we'd need to rename config variables, update the GUI, and update a lot of comments...

Let's discuss in #633

artoonie:
All comments addressed except for the Undervote ambiguity

tarheel (approved):

PR #714, opened 2023-06-21
Issue 633 (New CSV Format) -- Part 3: Implementation

artoonie:
Closes https://github.com/BrightSpots/rcv/issues/633, https://github.com/BrightSpots/rcv/issues/678 :

Before merging, `TabulatorTests.java` must be reverted to avoid committing the code which overwrites all summary CSV and JSON files.

Note: this PR has no changes to any tabulation code.

This is part 3 of a 3-part PR Chain:

1. Add tests to ensure no CSVs change [#713]
2. Update tabulation code, ensuring no adjustments to results writing [#711]
3. Update results writing code, ensuring no adjustments to tabulation code [this]


src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -339,23 +333,81 @@ private void generateSummarySpreadsheet(
 
         // Vote count
         csvPrinter.print(thisRoundTally);
+
+        // Vote %
+        BigDecimal activeBallots = roundTallies.get(round).numActiveBallots();
+        if (activeBallots != BigDecimal.ZERO) {
+          BigDecimal divDecimal = thisRoundTally.divide(activeBallots, MathContext.DECIMAL32);
+          csvPrinter.print(divDecimal.scaleByPowerOfTen(4).intValue() / 100.0 + "%");
```
tarheel, line 341: add an inline comment explaining what this is doing

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -339,23 +333,81 @@ private void generateSummarySpreadsheet(
...
       }
+      csvPrinter.println();
+    }
+
+    csvPrinter.print("Inactive Ballots Total");
+    BigDecimal numUndervotes = roundTallies.get(1).getBallotStatusTally(
+        StatusForRound.INACTIVE_BY_UNDERVOTE);
+    for (int round = 1; round <= numRounds; round++) {
+      BigDecimal inactiveBallots = roundTallies.get(round).numInactiveBallots()
+          .subtract(numUndervotes);
```
tarheel, line 406: Could use an inline comment here also. If I'm understanding it: the undervote count never changes, so if you want to show the sum of all of the other types of inactive ballots for each round, you need to subtract that undervote count from the round's overall inactive ballot count. And you could compute that same number for each round in the preceding nested `for` loops if you kept a running tally as you looped through the different inactive statuses.

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -339,23 +333,81 @@ private void generateSummarySpreadsheet(
...
 
       csvPrinter.print(inactiveBallots);
+      csvPrinter.print("");
+      csvPrinter.print("");
```
tarheel, line 410: add comment (here and below) clarifying why you do this (filling in unused cells)

src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -208,6 +208,18 @@ private static void compareFiles(
...
+    // Overwrite expected file with the generated one
+    Boolean rewriteAllExpectedFiles = false;
+    if (rewriteAllExpectedFiles && !outputType.equals("cvr_cdf")) {
+      try {
+        new File(expectedPath).setWritable(true);
+        copyFile(new File(actualOutputPath), new File(expectedPath));
+      } catch (IOException e) {
+        fail(e);
+      }
+    }
```
tarheel, line 221: OK, so you're going to delete this block of code.

tarheel (approved):

src/main/java/network/brightspots/rcv/ResultsWriter.java
artoonie: Done -- and actually, I should have been filling out the transfer value here like above, so that's done too.

src/main/java/network/brightspots/rcv/ResultsWriter.java
artoonie: Done

src/main/java/network/brightspots/rcv/ResultsWriter.java
artoonie: Done

src/test/java/network/brightspots/rcv/TabulatorTests.java
artoonie: Deleted since the files shouldn't change any more

PR #717, opened 2023-06-23
Option to prevent final round from having just one candidate

### LINKED ISSUE #694: Fix final-round tiebreaker
https://github.com/BrightSpots/rcv/pull/646#discussion_r1224709054

As seen in `tiebreak_seed_test`, the tiebreaker should occur on the winner, not the elimination, when it occurs in the final round of a single-winner election

artoonie:
Closes #694 
image

Option to prevent final round from having just one candidate. Only available in single-winner.

This has been implemented and tested in two cases: first, when both candidates have the same number of votes (tiebreaker tests) and when both candidates have a different number of votes ("first round determines threshold" case)


artoonie:
Done!

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -693,7 +703,7 @@ private List<String> doRegularElimination(
     LinkedList<String> lastPlaceCandidates = currentRoundTallyToCandidates.get(minVotes);
     if (lastPlaceCandidates.size() > 1) {
       // there was a tie for last place
-      // create new Tiebreak object to pick a loser
+      // create new Tiebreak object to pick a winner
```
tarheel, line 706: Not clear why you changed this comment. This Tiebreak object is selecting a loser, not a winner.

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -548,14 +551,21 @@ private List<String> identifyWinners(
       // * If this is a single-winner election in which it's possible for no candidate to reach the
       //   threshold (i.e. "first round determines threshold" is set), the tiebreaker will choose
       //   the only winner.
```
tarheel, line 553: I think this comment is a bit out of place now.

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -548,14 +551,21 @@ private List<String> identifyWinners(
...
+      boolean needsTiebreakNoWinners = selectedWinners.size() == 0
+          && (currentRoundTally.numActiveCandidates() == 2
+              && config.doPreventOneCandidateInFinalRound());
+      if (needsTiebreakMultipleWinners || needsTiebreakNoWinners) {
         // currentRoundTallyToCandidates is sorted from low to high, so just look at the last key
         BigDecimal maxVotes = currentRoundTallyToCandidates.lastKey();
         selectedWinners = currentRoundTallyToCandidates.get(maxVotes);
         // But if there are multiple candidates tied for the max tally, we need to break the tie.
-        if (selectedWinners.size() > 1) {
+        if (selectedWinners.size() > 1 || (selectedWinners.size() == 0)) {
```
tarheel, line 568: nit: extraneous parens around second half of this

tarheel: But also, will `selectedWinners` ever be empty on this line? The previous statement just set it to `currentRoundTallyToCandidates.get(maxVotes)`.

I think this section of the code would be easier to follow if you kept the checks for the different scenarios distinct and then ended with `if (needTiebreak) { doTiebreak }`

src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
```diff
@@ -24,6 +24,8 @@ Continue until Two Candidates Remain: Single-winner ranked-choice voting electio
 
 First Round Determines Threshold: Single-winner ranked-choice voting elections can use the first round to determine the winning threshold, rather than the last round. This will not affect the eventual winner. This may elect a winner that does not reach the threshold.
 
+Prevent Final Round From Having Just One Candidate: If two candidates remain and neither has met the threshold, select one as the winner instead of running an elimination round.
```
tarheel, line 27: I'm curious what @chughes297 thinks here. My instinct is that we would want the default behavior to be that we prevent the final round from having just one candidate. And if that's the case, perhaps this config setting should be "Allow Final Round to Have Just One Candidate" and it should default to false.

Also, we should clarify which types of contests this is relevant for. (And disable it in the UI/raise an error when it's used inappropriately.)

src/main/java/network/brightspots/rcv/Tabulator.java
artoonie: Is it? The logic is the same on the next line.

src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
artoonie: I defer to @chughes297 on what the default behavior should be.

I don't know if any legislation specifies this explicitly, so it seems that it comes down to which statute is seen more frequently:
1. Tabulate until somebody crosses the threshold
2. Tabulate until two candidates remaining, then pick the top vote-getter

In the first case, that would indicate that we _should_ go down to one candidate. In the second case, we _shouldn't_.



As for clarifying contest types: do you mean disabling this option for STV, or are there other cases where it should be disabled?

chughes297: Agree that we should default to including two candidates in the final round. As I look at this now though I'm not sure we want to give any users an option here - the default tabulation rule of having two people in the final round is all we should do, imo. I don't think there's any value in giving people the option of going to a single-candidate final round (and I can't think of a law that requires it? Sorry if I gave confusing input earlier that implied that...)

artoonie:
https://github.com/BrightSpots/rcv/pull/731 replaces this one; this should be closed without merging if #731 is approved

PR #721, opened 2023-06-26
Prompt for operator name

### LINKED ISSUE #631: Login Feature to Track Users
RCTab needs the ability to track who's conducting the tabulation for auditing purposes. Ideally, this would involve a user login and password. 

Thoughts on the feasibility of including in 1.4?

artoonie:
Closes #631 

In GUI:
image

In CLI:
image

CLI can still work interactively if `--name` is specified, but the audit log will show that the name wasn't interactively entered.

Given the addition of the CLI argument, we needed the support of a framework to avoid clutter and stringent argument order.


build.gradle
```diff
@@ -29,11 +30,11 @@ application {
...
-// Uncomment below to simulate running from the CLI
-//run {
-//    standardInput = System.in
-//    args = ["-cli", "path/to/config"]
-//}
+// Required to accept a name and tiebreak interactively via CLI
+run {
+    standardInput = System.in
+    args = ["-cli", "path/to/config"]
+}
```
HEdingfield, line 37: What are the ramifications of leaving this uncommented now? Before it was a quick way for me to swap in having the `run` commend launch the CLI instead of the GUI.

How can we easily switch to this mode now using Gradle?

build.gradle
```diff
@@ -29,11 +30,11 @@ application {
...
 
-// Uncomment below to simulate running from the CLI
-//run {
-//    standardInput = System.in
-//    args = ["-cli", "path/to/config"]
-//}
+// Required to accept a name and tiebreak interactively via CLI
+run {
+    standardInput = System.in
+    args = ["-cli", "path/to/config"]
```
HEdingfield, line 36: Do we not need to switch to `--cli` here too?

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -945,6 +948,21 @@ private ConfigComparisonResult compareConfigs() {
...
+    dialog.setTitle("Enter your name");
+    dialog.setHeaderText("For auditing purposes, enter the name(s) of everyone currently "
+        + "operating this machine.");
+    dialog.setContentText("Name:");
+    Optional<String> result = dialog.showAndWait();
+
+    if (result.isPresent()) {
+      Logger.info("Name(s) of operators, as entered interactively: " + result.get());
+    } else {
+      Logger.info("Operator(s) did not enter a name.");
```
HEdingfield, line 962: Should we block this / throw an exception here instead?

src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -42,41 +51,36 @@ public static void main(String[] args) {
...
+      } else {
+        Logger.info("Name(s) of operators, as entered in a command-line argument: " + name);
       }
-      // Path to either: config file for configuring the tabulator, or Dominion JSONs
-      String providedPath = argsCli.get(0);
-      // Session object will manage the tabulation process
-      TabulatorSession session = new TabulatorSession(providedPath);
-      if (argsCli.size() == 2 && argsCli.get(1).equals("convert-to-cdf")) {
+
+      Logger.info("Tabulator is being used via the CLI");
```
HEdingfield, line 80: Why not keep this at the very start of the `else` statement (line 59)?

src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -42,41 +51,36 @@ public static void main(String[] args) {
...
-            argsCli.size());
-        System.exit(2);
+      CommandLine cmd = parseArgsForCli(args);
+      String path = cmd.getOptionValue("cli");
+      String name = cmd.getOptionValue("name");
+      Boolean convertToCdf = cmd.hasOption("convert-to-cdf");
+
+      if (name == null || name.isEmpty()) {
+        // Prompt user for their name
+        System.out.println("Enter name(s) of operators for auditing purposes:");
```
HEdingfield, line 66: Can we provide a name of "Automated Test" any time we run a test somewhere? And, if so, would it simplify any of this logic?

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -945,6 +948,21 @@ private ConfigComparisonResult compareConfigs() {
     return comparisonResult;
   }
 
+  private void askUserForName() {
+    TextInputDialog dialog = new TextInputDialog();
+    dialog.setTitle("Enter your name");
+    dialog.setHeaderText("For auditing purposes, enter the name(s) of everyone currently "
+        + "operating this machine.");
+    dialog.setContentText("Name:");
+    Optional<String> result = dialog.showAndWait();
```
HEdingfield, line 957: The cancel button doesn't properly cancel out of this, it just runs it with no name provided. Same problem with the X button in the corner.

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -945,6 +948,21 @@ private ConfigComparisonResult compareConfigs() {
...
+
+    if (result.isPresent()) {
+      Logger.info("Name(s) of operators, as entered interactively: " + result.get());
```
HEdingfield, line 960: Note that, a few lines below this in the logs, it also gives:

```
2023-06-26 14:49:56 PDT INFO: Computer name: abc
2023-06-26 14:49:56 PDT INFO: User name: xyz
```

This could be confusing to people. Maybe at the very least we should change "user name" to "Computer user name" or something?

HEdingfield (changes requested):

build.gradle
artoonie: Ah, I didn't read this too closely. The `standardInput` is needed and has no ramifications, best I can tell. the fact that `-cli` was not `--cli` meant the `args` were ignored. Will fix and leave `args` commented.

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: I prefer this, since when developing, I often hit tabulate a lot and can just hit cancel.

My reasoning is that we can't guarantee they'll enter a correct name, so why pretend to validate? They could enter `foo` instead of empty string. Just let them do empty string.

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: Yeah, I was using the cancel functionality while developing -- I can just hit `esc` and tabulation continues. I can adjust to stop tabulation instead.

src/main/java/network/brightspots/rcv/Main.java
artoonie: We can provide any name any time we test. If provided via cli argument, you won't be prompted here. Not sure if that's what you mean by simplify?

artoonie:
All comments addressed -- entering no name or canceling out of the name prompt cancels tabulation now

src/main/java/network/brightspots/rcv/Main.java
HEdingfield: Nevermind, I didn't realize that all this is just bypassed for our tests.

src/main/java/network/brightspots/rcv/GuiConfigController.java
HEdingfield: Ok, but if that's the case, why do we force the user to provide a value in the CLI version? Shouldn't we treat this consistently?

I'm ok with allowing the user to bypass without entering a name but, if they do, we should probably log a warning (or maybe just an info) saying the user didn't provide a name.

artoonie: We already log an INFO that user didn't provide a name (`Operator(s) did not enter a name.`)

artoonie: I can update the CLI to not fail on empty string, but I fear it's more hidden there. Especially when running via gradle, the line `<===========--> 85% EXECUTING [20s]` shows up after the request for a name, so it could be hidden.

In other words, empty string in CLI is more likely to be accidental, whereas in tabulation, it's more obvious.

Happy to make changes in either direction!

HEdingfield: I think an argument in favor of throwing an exception / requiring this in both cases is that the user might accidentally bypass the message box with a mis-click or mis-key. I say we should go forward with it. It's really easy to just enter "asdf" when testing here anyway.

src/main/java/network/brightspots/rcv/Main.java
```diff
@@ -42,41 +49,53 @@ public static void main(String[] args) {
...
-      // Check for unexpected input
-      if (argsCli.size() == 0) {
+      Logger.info("Tabulator is being used via the CLI");
+
+      CommandLine cmd = parseArgsForCli(args);
+      String path = cmd.getOptionValue("cli");
+      String name = cmd.getOptionValue("name");
+      boolean convertToCdf = cmd.hasOption("convert-to-cdf");
+
+      boolean validNameProvided = false;
```
HEdingfield, line 64: I went ahead and redesigned this CLI logic so it handles corner cases more robustly. However, in doing so, I also found a critical flaw in this PR: the name that's entered and logged doesn't actually make it into the audit logs!

I think we need to make all this happen deeper in the app, i.e. immediately after this line:

```
Logger.info("Computer user name: %s", Utils.getUserName());
```

Also think it's probably fine if we don't prompt the user for their name when converting to CDF, but you could confirm with @chughes297 about this.

HEdingfield (changes requested):

src/main/java/network/brightspots/rcv/Main.java
artoonie: Okay, I've made these changes: Convert to CDF doesn't require a name, and entering a blank name prevents tabulation.

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: Or aoeu, depending on keyboard layout ;)

HEdingfield (approved):
Nice work! Just fixed a couple last things.

tarheel:
> Whoa whoa whoa, are you one of those? I'm impressed... 👽

I heard that truly 3lit3 engineers only use [Colemak](https://colemak.com/).

PR #726, opened 2023-06-27
allow aliases to start and end with non-word characters

artoonie:
Regex was eating all non-word characters instead of all non-space characters, so while the first alias could start and end with `.;,` etc, the second alias could not.


artoonie:
Nevermind! Better to do it in the other PR.

PR #727, opened 2023-06-27
use dictionary-based comparison in all JSON tests

### LINKED ISSUE #723: Switch tests over to use `dict` comparisons
I noticed a new line appearing in all our test summaries / outputs: `"generatedBy" : "RCTab 1.4.0-alpha",` which seems brittle to me.

Discussed with @artoonie and he recommended that we switch our tests over to doing a `dict` comparison rather than a line-by-line file comparison. Advantages include:

* You can comment test files
* Different OSs might have different HashMap implementations, which would output different orders in dictionaries, leading to different line-by-line JSONs that are actually the same
* You can easily specify fields to ignore, without accidentally doing something silly, like saying "ignore all lines that have generatedBy then outputting all JSONs on one line so the entire comparison is ignored

artoonie:
closes #723 
we can easily add `generatedBy` later, and even have different ignore keywords based on what type of JSON we're comparing


moldover:
Could I ask one of you to remove the early return in TabulatorTests.java around line 112 while this code is being updated?  That's done to avoid multiple returns per vvsg coding standards.   Thanks.

artoonie:
done -- hopefully it's fine just having that change as part of this PR

tarheel:
> Could I ask one of you to remove the early return in TabulatorTests.java around line 112 while this code is being updated? That's done to avoid multiple returns per vvsg coding standards. Thanks.

@moldover's passion for adhering to VVSG standards burns like an eternal flame.

src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -109,36 +206,36 @@ private static void runTabulationTest(String stem, String expectedException) {
...
+          compareFiles(config, stem, timestampString, Integer.toString(i));
+        }
+      } else {
+        compareFiles(config, stem, timestampString, null);
+      }
 
-    if (config.isMultiSeatSequentialWinnerTakesAllEnabled()) {
-      for (int i = 1; i <= config.getNumberOfWinners(); i++) {
-        compareFiles(config, stem, timestampString, Integer.toString(i));
+      // If this is a Dominion tabulation test, also check the converted output file.
```
tarheel, line 223: Does it still make sense to limit this to Dominion in light of #715?

tarheel (approved):
This is a great improvement. Only thing I'm wondering is whether there's a third-party library that's already solved this problem. But obviously that's kind of moot now that you've implemented it.

src/test/java/network/brightspots/rcv/TabulatorTests.java
artoonie: Nope -- removed since #715 merged first!

artoonie:
> This is a great improvement. Only thing I'm wondering is whether there's a third-party library that's already solved this problem. But obviously that's kind of moot now that you've implemented it.

I originally implemented with a jackson feature, but there was no library we already had installed that would give a detailed error message. I didn't think this was worth adding a true additional dependency, though they do exist (e.g. [guava](https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/Maps.html#difference-java.util.Map-java.util.Map-) or [jackson-databind](https://github.com/FasterXML/jackson-databind))

PR #731, opened 2023-07-05
Don't allow one candidate in the final round

### LINKED ISSUE #694: Fix final-round tiebreaker

artoonie:
Closes #694 

In #717, we had an option to prevent the final round from having just one candidate.

In this PR, we mandate it, based on [this comment thread](https://github.com/BrightSpots/rcv/pull/717#discussion_r1253175454).

All of the tests below had just two candidates on the last round and the tiebreaker and had an elimination tiebreaker, except for `stop tabulation early` test, which has been updated to stop tabulation before the tiebreaker would have hit.

In that test, we requested tabulation stop on Round 2, but Round 2 was a tiebreaker and tabulation would now end there regardless (just without choosing a winner). I now stop tabulation on Round 1.

### DIFF
--- a/src/main/java/network/brightspots/rcv/Tabulator.java
+++ b/src/main/java/network/brightspots/rcv/Tabulator.java
@@ -520,9 +520,9 @@ class Tabulator {
       }
     } else {
       // We should only look for more winners if we haven't already filled all the seats.
-      if (winnerToRound.size() < config.getNumberOfWinners()) {
-        if (currentRoundTally.numActiveCandidates()
-            == config.getNumberOfWinners() - winnerToRound.size()) {
+      int numSeatsUnfilled = config.getNumberOfWinners() - winnerToRound.size();
+      if (numSeatsUnfilled > 0) {
+        if (currentRoundTally.numActiveCandidates() == numSeatsUnfilled) {
           // If the number of continuing candidates equals the number of seats to fill,
           // everyone wins.
           selectedWinners.addAll(currentRoundTally.getCandidates());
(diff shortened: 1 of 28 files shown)


src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -548,9 +548,19 @@ private List<String> identifyWinners(
...
+          || config.isFirstRoundDeterminesThresholdEnabled());
+      // Edge case: there are two candidates remaining. To avoid having just one candidate in the
+      // final round, we break the tie here. Happens when we have unfilled seats, two candidates
+      // remaining, neither meets the threshold, and both have more than the minimum vote threshold.
+      boolean needsTiebreakNoWinners = config.getNumberOfWinners() == 1
+          && selectedWinners.size() == 0
+          && currentRoundTally.numActiveCandidates() == 2
+          && numSeatsUnfilled == 1
+          && currentRoundTallyToCandidates.keySet().stream().allMatch(
+              x -> x.compareTo(config.getMinimumVoteThreshold()) >= 0);
```
tarheel, line 562: This doesn't look logically wrong, but are these last two parts of the condition actually necessary? In other words, could either of these ever be false when the first three conditions are true?

tarheel (commented):
Just one question...

src/main/java/network/brightspots/rcv/Tabulator.java
artoonie: Yes, without this we had one failing test (`minimum_threshold_test`) -- since everyone under the threshold is still considered active until they're eliminated.

tarheel: I was confused by this until I realized that you meant to refer to the `no_one_meets_minimum` test. That makes sense. Maybe add an inline comment explaining why we need this condition?

It would also be helpful to explain why we need `numSeatsUnfilled == 1`, since it's also not obvious (at least to me) at first glance why it's necessary in addition to the other conditions.

tarheel (approved):

src/main/java/network/brightspots/rcv/Tabulator.java
artoonie: Added comments explaining it all! The numSeatsUnfilled is a check to see if the winner was selected in a previous round.

PR #751, opened 2023-09-15
make Convert to CDF filenames more likely to be unique

### LINKED ISSUE #734: When tabulating multiple CVR files post-tabulation CVR conversion export does not include all data
And for some reason the errors aren't appearing in any of my log files, either? I was running the scale test [from here](https://github.com/BrightSpots/rcv/tree/develop/src/test/resources/network/brightspots/rcv/test_data/2013_minneapolis_mayor_scale) and got the attached CVR conversion on the other end. 
[2023-07-13_12-44-52_dominion_conversion_contest_.csv](https://github.com/BrightSpots/rcv/files/12069595/2023-07-13_12-44-52_dominion_conversion_contest_.csv)

I'm guessing this is because we've made files generated by RCTab read-only - it looked like RCTab created the CVR conversion file after reading in the first CVR and then couldn't update that file after it had made it read only.

artoonie:
closes #734 

not all contests have contest IDs, so relying on them for uniqueness causes files to be overwritten. Don't do that anymore.

It's still possible for this bug to pop up if all CVR files have the same filename and there is no contest ID. Do you think that's acceptable? Alternatively, I could (a) force a Contest ID; or (b) increment until a unique filename is found -- but I think this is fine?

After merging into develop, I will cherry-pick onto 1.3.2.

### DIFF
--- a/src/main/java/network/brightspots/rcv/ResultsWriter.java
+++ b/src/main/java/network/brightspots/rcv/ResultsWriter.java
@@ -553,17 +553,28 @@ class ResultsWriter {
       List<CastVoteRecord> castVoteRecords,
       Integer numRanks,
       String csvOutputFolder,
+      String inputFilepath,
       String contestId,
       String undeclaredWriteInLabel)
       throws IOException {
     String fileWritten;
+    // Get input filename with extension
+    String inputFileBaseName = new File(inputFilepath).getName();
+    // Remove the extension if it exists
+    int lastIndex = inputFileBaseName.lastIndexOf('.');
+    if (lastIndex != -1) {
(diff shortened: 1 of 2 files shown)


tarheel (approved):
Maybe include the contestId (when present) also? And if you want to make it truly robust, add a random string at the end?

HEdingfield:
@artoonie Was this cherry-picked to 1.3.2 yet? I don't see a record of it.

artoonie:
It's not needed in 1.3.2 (see commentary here: https://github.com/BrightSpots/rcv/issues/753)

HEdingfield:
Got it, updated description to reflect this.

PR #752, opened 2023-09-18
include SHA256 for output files

### LINKED ISSUE #746: Generate hash for output files
To increase security, along with #696, we've been asked to generate a hash of each output file to verify that the output files have not been edited.  For now, can RCTab generate a read only hash for the following output files

- each audit\__N_.log
- summary.csv
- summary.json

Does it make sense to also include the hash of each summary file into the text of the audit log?

artoonie:
1. Every file that's written should now use AuditableFile, instead of File. When the file writing is complete, you should call `finalizeAndHash` on it. That will mark it as read-only and add its hash to the audit log.
2. An exception is made for files in the Logger to avoid a circular dependency: how can you write your own hash to your file contents? In this case, we add the hash to the filename.

Alternatively, you could:
```
String hash;
do {
   hash = getHash(logFilePath);
   ReplaceLastLineOfFileWith("The hash of this file is %s".formatted(hash));
} while (getHash(logFilePath) != lastSeenHash)
```
But that will take around 10^60 years to complete by my math.

Resolves #746

### DIFF
--- a/src/main/java/network/brightspots/rcv/AuditableFile.java
+++ b/src/main/java/network/brightspots/rcv/AuditableFile.java
@@ -32,8 +32,8 @@ final class AuditableFile extends File {
     Logger.info("File %s written with hash %s".formatted(getAbsolutePath(), hash));
 
     // Write hash to hash file
-    File hashFile = new File(getAbsolutePath() + ".sha");
-    writeStringToFile(hashFile, hash);
+    File hashFile = new File(getAbsolutePath() + ".hash");
+    writeStringToFile(hashFile, "sha512: " + hash);
 
     // Make both file and its hash file read-only
     makeReadOnlyOrLogWarning(this);
--- a/src/main/java/network/brightspots/rcv/AuditableFile.java
+++ b/src/main/java/network/brightspots/rcv/AuditableFile.java
@@ -17,6 +17,8 @@
 package network.brightspots.rcv;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
 
 final class AuditableFile extends File {
   public AuditableFile(String pathname) {
@@ -24,12 +26,32 @@ final class AuditableFile extends File {
   }
 
   public void finalizeAndHash() {
-    boolean readOnlySucceeded = setReadOnly();
+    String hash = FileUtils.getHash(this);
+
+    // Write hash to audit log
+    Logger.info("File %s written with hash %s".formatted(getAbsolutePath(), hash));
+
+    // Write hash to hash file
+    File hashFile = new File(getAbsolutePath() + ".sha");
+    writeStringToFile(hashFile, hash);
+
+    // Make both file and its hash file read-only
+    makeReadOnlyOrLogWarning(this);
+    makeReadOnlyOrLogWarning(hashFile);
+  }
+
+  private void writeStringToFile(File file, String string) {
+    try {
+      Files.writeString(file.toPath(), string);
+    } catch (IOException e) {
+      Logger.severe("Could not write to file %s: %s", file.getAbsoluteFile(), e.getMessage());
+    }
+  }
+
+  private void makeReadOnlyOrLogWarning(File file) {
+    boolean readOnlySucceeded = file.setReadOnly();
     if (!readOnlySucceeded) {
       Logger.warning("Failed to set file to read-only: %s", getAbsolutePath());
     }
-
-    String hash = FileUtils.getHash(this);
-    Logger.info("File %s written with hash %s".formatted(getAbsolutePath(), hash));
   }
 }
--- a/src/main/java/network/brightspots/rcv/AuditableFile.java
+++ b/src/main/java/network/brightspots/rcv/AuditableFile.java
@@ -1,0 +1,35 @@
+/*
+ * RCTab
+ * Copyright (c) 2017-2023 Bright Spots Developers.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * Purpose: Create a file that, on close, is read-only and has its hash added to the audit log.
+ * Design: Overrides the File class
+ * Conditions: Always.
+ * Version history: see https://github.com/BrightSpots/rcv.
+ */
+
+package network.brightspots.rcv;
+
+import java.io.File;
+
+final class AuditableFile extends File {
+  public AuditableFile(String pathname) {
+    super(pathname);
+  }
+
+  public void finalizeAndHash() {
+    boolean readOnlySucceeded = setReadOnly();
+    if (!readOnlySucceeded) {
+      Logger.warning("Failed to set file to read-only: %s", getAbsolutePath());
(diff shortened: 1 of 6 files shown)


src/main/java/network/brightspots/rcv/AuditableFile.java
```diff
@@ -0,0 +1,35 @@
...
+
+final class AuditableFile extends File {
+  public AuditableFile(String pathname) {
+    super(pathname);
+  }
+
+  public void finalizeAndHash() {
+    boolean readOnlySucceeded = setReadOnly();
+    if (!readOnlySucceeded) {
+      Logger.warning("Failed to set file to read-only: %s", getAbsolutePath());
```
tarheel, line 29: Any chance that we want this type of error to trigger `severe` instead of `warning`?

src/main/java/network/brightspots/rcv/Utils.java
```diff
@@ -86,4 +91,41 @@ static String getUserName() {
   static String[] splitByNewline(String s) {
     return s.trim().split("\\s*\\r?\\n\\s*");
   }
+
+  static byte[] readFileToByteArray(File file) {
+    byte[] fileBytes = null;
+    try {
+      fileBytes = Files.readAllBytes(file.toPath());
```
tarheel, line 98: Probably not necessary for now, but ideally we would have a way to compute this without reading the entire file into memory at once. Maybe something like what's discussed here would work? https://stackoverflow.com/questions/76013476/java-inputstream-copy-to-and-calculate-hash-at-the-same-time

tarheel (approved):
> But that will take around 10^60 years to complete by my math.

We could make an M2 processor a requirement for running the tabulator. I bet it could complete this in more like 10^59 years.

src/main/java/network/brightspots/rcv/AuditableFile.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * Purpose: Create a file that, on close, is read-only and has its hash added to the audit log.
+ * Design: Overrides the File class
+ * Conditions: Always.
```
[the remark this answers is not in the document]
artoonie: `Conditions: Always`? I assume it means "when is this used," since other values include "during tabulation". Since "Always" is used by Logger, I assume it was appropriate to use here too -- i.e. no matter what you do (tabulate, log, launch gui, or do something on CLI), this will be used.

src/main/java/network/brightspots/rcv/AuditableFile.java
artoonie: I opted for warning since I think there are valid cases where this will fail, though I'm not sure of them. For example, if saving to a mounted Google Drive folder, are you allowed to set a file to readonly?

I haven't tested any of this, but given that the function returns a boolean rather than throwing an exception, I assumed there would be valid, non-severe cases where this fails.

src/main/java/network/brightspots/rcv/Utils.java
artoonie: Good call -- I imagine at some point we may decide to hash the input files too, so I've updated this function to read the file in 1kb chunks. Can you take another quick look before I merge?

src/main/java/network/brightspots/rcv/AuditableFile.java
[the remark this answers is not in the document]
yezr: Is this from some certification requirements?

src/main/java/network/brightspots/rcv/FileUtils.java
```diff
@@ -50,6 +56,29 @@ static void createOutputDirectory(String dir) throws UnableToCreateDirectoryExce
...
+          // Read in 1kb chunks -- don't need to do anything in the body here
+        }
+      }
+    } catch (IOException e) {
+      Logger.severe("Failed to read file: %s", file.getAbsolutePath());
+      return "[hash not available]";
+    }
+
+    return Utils.bytesToHex(digest.digest());
+  }
```
tarheel, line 80: Oh, cool. So this means that the footprint is O(1) instead of O(n)? Also curious why you chose 1024 instead of, say, that number squared or whatever. Not sure it makes any meaningful difference in perf.

artoonie: I assume it doesn't affect performance much either way, though we can revisit if we end up hashing large CVRs and performance becomes a concern.

artoonie:
Ready for review with new changes: using a .sha file instead of including the hash in the filename

src/main/java/network/brightspots/rcv/Logger.java
```diff
@@ -95,17 +95,19 @@ static void setup() {
   // adds file logging for a tabulation run
   static void addTabulationFileLogging(String outputFolder, String timestampString)
       throws IOException {
-    // log file name is: outputFolder + timestamp + log index
-    // FileHandler requires % to be encoded as %%.  %g is the log index
+    // log file name is: outputFolder + timestamp + log index + hash + .log
```
tarheel, line 98: revert?

src/main/java/network/brightspots/rcv/Utils.java
```diff
@@ -86,4 +93,27 @@ static String getUserName() {
   static String[] splitByNewline(String s) {
     return s.trim().split("\\s*\\r?\\n\\s*");
   }
+
+  static byte[] readFileToByteArray(File file) {
```
tarheel, line 97: no longer used, right?

tarheel (approved):

artoonie:
1. PR comments fixed
2. Changed `.sha` files to `.hash` and included hash algo in file body
3. Changed signing algo from sha256 to 512

Final review I think! @tarheel

src/main/java/network/brightspots/rcv/Utils.java
```diff
@@ -86,4 +93,16 @@ static String getUserName() {
...
     return s.trim().split("\\s*\\r?\\n\\s*");
   }
+
+  static String bytesToHex(byte[] hash) {
+    StringBuilder hexString = new StringBuilder(2 * hash.length);
+    for (byte b : hash) {
+      String hex = Integer.toHexString(0xff & b);
+      if (hex.length() == 1) {
+        hexString.append('0');
+      }
```
tarheel, line 103: could you explain this part?

tarheel (approved):

src/main/java/network/brightspots/rcv/Utils.java
artoonie: `toHexString` strips all leading zeros for some reason, so byte 0 would be hex `0` instead of `00`

tarheel: I'm probably just confused here, but why aren't we prepending '0' instead of appending it?

artoonie: We append `0` before we append the remainder of the string, so it's effectively prepending!

tarheel: Sigh. Reading comprehension!

PR #761, opened 2023-10-24
Don't allow output directory in user directory

### LINKED ISSUE #757: Targeting Windows user account folders for output path should fail config validation
*This issue is directly related to #748 and the 1.3.2 build. Along with #747, this should be part of the two-build system specific to CA*

On Windows, one of the requirements for ensuring read-only access for RCTab output is setting the right folder permissions on the output folder RCTab is configured for. Step by step instructions for setting the necessary folder permissions are described in the installation instructions of Section 22 of the TDP. 

From testing in Windows, I found that #748 successfully prevents modifications, but user accounts retain delete and create permissions to their user account folders (e.g. Desktop, My Documents) in `c:\users\` despite following the installation instructions to restrict permissions on those folders. 

Can we add a check into the config verification step before tabulation begins to throw a halting error if the configured output path resolves to any path under `c:\users\`?

artoonie:
Closes #757 

After approval, I will toggle both security flags and turn them off before merging into `develop`
Then, I will cherry-pick into 1.3.2 and toggle the security flags back to what they are here


tarheel (commented):
Seems reasonable to me. Will let @HEdingfield look also.

artoonie:
Sounds good. I cherry-picked the commits particular to this PR to resolve the merge conflicts, and swapped the booleans so both this and the previous PR are no-ops. Once this lands, I'll cherry-pick into 1.3.2 and swap the booleans again.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -515,6 +515,26 @@ private void validateOutputSettings() {
...
       Logger.severe("contestName is required!");
     }
+
+    if (!SecurityConfig.canOutputFilesSaveToUsersDirectory()) {
+      try {
+        String homeDirectory = new File(System.getProperty("user.home")).getCanonicalPath();
+        // Strip the username off the file path to get the root Users directory path
+        String rootUsersDirectory = homeDirectory.substring(0, homeDirectory.lastIndexOf('\\'));
+        String outputDirectory = new File(getOutputDirectory()).getCanonicalPath();
+        if (outputDirectory.startsWith(rootUsersDirectory)) {
```
HEdingfield, line 525: I'm struggling with the entire premise of this PR -- please see https://github.com/BrightSpots/rcv/issues/757#issuecomment-1793502225.

src/main/java/network/brightspots/rcv/SecurityConfig.java
```diff
@@ -22,7 +22,12 @@ class SecurityConfig {
   // Only the unit test modules should ever set this to false, if it is initially set as true.
   // Note: On some builds, this will be configured to false by default. We will need some
   // formalized method of toggling this for two versions of builds, which has yet to be determined.
-  private static boolean IS_HART_SIGNATURE_VALIDATION_ENABLED = true;
+  private static boolean IS_HART_SIGNATURE_VALIDATION_ENABLED = false;
+
+  // Is the user allowed to save output files to their Users directory?
+  // Since user accounts retain delete and create permissions to their user account folders,
+  // this should be disallowed to truly ensure output files are read-only.
+  private static boolean CAN_OUTPUT_FILES_SAVE_TO_USERS_DIRECTORY = true;
```
HEdingfield, line 30: > After approval, I will toggle both security flags and turn them off before merging into develop
> Then, I will cherry-pick into 1.3.2 and toggle the security flags back to what they are here

Does this mean we're going to have to remember to do this every single time we do a launch? It'd be pretty annoying to keep this `false` for development cycles, but it also seems easy to forget about doing this whenever we do a prod release.

HEdingfield (changes requested):

src/main/java/network/brightspots/rcv/ContestConfig.java
artoonie: Let me know if your concerns were addressed there

src/main/java/network/brightspots/rcv/SecurityConfig.java
artoonie: Yes, and we need a better way to manage these flags. For now, it's nice that they're centrally located so we can turn on strict mode. Eventually, we absolutely need a better way to approach this than hoping to remember to change flags.

HEdingfield: What if we made a helper function somewhere that checks if the current version is *999* (dev) and, if so, it flips them accordingly. Otherwise, it flips them the other way?

We could use this same function to test whether or not to render the developer version banner in the GUI too.

HEdingfield: If we do go this route, perhaps we could create the helper in #719 and then merge that in, then make use of it in this PR?

yezr: I'm looking forward to #719 because I think it's going to make sure we have solid version control. But in the interest of getting a v1.3.2 release candidate out can we defer that work until later? I've run past a couple of my estimates for this build release and the pressure is building to get it out

src/main/java/network/brightspots/rcv/ContestConfig.java
HEdingfield: Yep. It's a lot to test, so  I'm just going to assume this will solve the issue as expected.

src/main/java/network/brightspots/rcv/SecurityConfig.java
HEdingfield: It would be trivial to do this, but from what @artoonie said in Slack, these flags aren't something we'll necessary want flipped the same way for 1.3.2 as in future releases, so I guess it's best to handle them ad hoc now.

HEdingfield (approved):
LGTM

PR #764, opened 2023-10-30
RSA Validation: Cherry-Picked ont 1.3.2 release

artoonie:
Cherry-picking #755 onto the 1.3.2 release


HEdingfield:
Will leave this to @tarheel to approve, since he approved #755 (also because I'm unsure of how to actually test that this is WAI on my end).

HEdingfield:
Also seems like we'd want to merge #765 first...

artoonie:
We can merge either first. If this one goes in first, I'll cherry-pick the relevant commits in #765 so there are no merge conflicts.

To check if this works:
1. Run the tests
2. Download the verity files with .sig.xml hashes (available on Slack) and create a config that loads them. Then hit "Auto-Load Candidates." If the signature is valid, it will work. Change one of the signatures or data files. It will then fail.

HEdingfield:
> To check if this works:
> 
> 1. Run the tests
> 2. Download the verity files with .sig.xml hashes (available on Slack) and create a config that loads them. Then hit "Auto-Load Candidates." If the signature is valid, it will work. Change one of the signatures or data files. It will then fail.

Trying out 2... there is no auto-load candidates option. Wasn't this something we added in 1.4.0?

yezr:
> Wasn't this something we added in 1.4.0?
yep! You should also be able to run a tabulation to verify that the signature verification works. It won't tabulate if it can't verify the signature. 

Here are some of the illustrious candidates from that test data that you can put in manually
- Mickey Mouse
- Bugs Bunny
- Homer Simpson
- Woody Woodpecker
- Scrooge McDuck
- Toucan Sam

tarheel (approved):

artoonie:
> Here are some of the illustrious candidates from that test data that you can put in manually

Yeah, sorry, this was the testing process before the cherry-pick.

HEdingfield:
Tested this out locally with some of the candidates and it failed with "no cast vote records found". Tried it again after modifying one of the sig files and it failed with "Failure while trying to verify hash", so I guess that worked?  Ah well, merged now anyway!

Issue #767, opened 2023-11-07
[1.4.0] Research continued FIPS 140-2 Compliance


yezr:
#765 added the BouncyCastle library to RCTab 1.3.2 in order to satisfy FIPS 140-2 compliance mandated in California Voting System Standards (CVSS). 1.4.0 upgrades Java to Java 20. Those BouncyCastle libraries are not compatible with Java 20. 

We need to make a decision on how to move forward in 1.4.0: do we find another FIPS 140-2 compliant library? Do we downgrade to an earlier Java version that is compatible with BouncyCastle? Do we somehow implement a two build system, one for CA and one for other jurisdictions?

yezr:
We also need to understand the ramifications of the changes in #769 that were required to build the jar with BouncyCastle. SO has some details [here](https://github.com/BrightSpots/rcv/pull/769)

HEdingfield:
From Slack convo:

@HEdingfield 
> RE: FIPS, what does it do if you try to run that code with 20? Does it fail silently, or just refuse to compile or what?


@artoonie 
> I haven't tried running FIPS on 20, I expect it will neither fail silently nor refuse to compile -- it just doesn't have the FIPS certification. That means everything should work, but we won't be FIPS compliant.

@yezr 
> like the bouncycastle library silently isn't FIPS compliant when running with java 20?


@artoonie 
> Right -- if you look here: https://www.bouncycastle.org/fips_java_roadmap.html
> You'll see the release notes of 1.0.2.4:
> > Patch release to add Java 17 to BC-FJA 1.0.2.3 as an operational environment. Removes the risk of CVE 2022-45146 and also deals with the end of the transition periods for PKCS 1.5 RSA encryption and TDES encryption. Module is now certified with certificate #4616.
>
> So we are good on 1.3.2 which uses Java 17, but we kind of got lucky that we weren't on -- we just happened to be using the exact right version of Java here.

HEdingfield:
@artoonie created #770, pending https://github.com/BrightSpots/rcv/issues/767#issuecomment-1806078003.

PR #769, opened 2023-11-10
Allow bouncycastle to build on non-Oracle JDKs

artoonie:
See: https://stackoverflow.com/questions/50597926/signed-modular-jar-with-crypto-provider-cannot-be-linked-into-run-time-image/68295740#68295740 for the error we were getting on openjdk

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -106,7 +106,7 @@ jlink {
     imageDir = file(JLINK_DIR)
     imageZip = file(JLINK_DIR + ".zip")
     addOptions '--add-modules', MODULES_TO_ADD, '--strip-debug', '--compress', '2',
-            '--no-header-files', '--no-man-pages'
+            '--no-header-files', '--no-man-pages', '--ignore-signing-information'
     mergedModule {
         requires "java.xml"
     }


build.gradle
```diff
@@ -106,7 +106,7 @@ jlink {
     imageDir = file(JLINK_DIR)
     imageZip = file(JLINK_DIR + ".zip")
     addOptions '--add-modules', MODULES_TO_ADD, '--strip-debug', '--compress', '2',
-            '--no-header-files', '--no-man-pages'
+            '--no-header-files', '--no-man-pages', '--ignore-signing-information'
```
HEdingfield, line 109: Is there a downside / danger to adding this?

If not, should we merge into 1.4.0 as well?

If so, we should flag this PR in #767 as another thing we'd need to eventually get into 1.4.0.

HEdingfield (approved):

build.gradle
yezr: I will definitely flag this in #767 to address before merging to 1.4.0

artoonie: I think we do need to. I created a draft PR for discussion here: https://github.com/BrightSpots/rcv/pull/770

Issue #766, opened 2023-11-03
[2.0-RC] Auto-Loading candidates fails when 'Maximum Number of Candidates That Can Be Ranked' setting is set to 'Maximum'


yezr:
RCTab v 1.4.0
Using the test_data `clear_ballot_kansas_primary`
Config: [clear ballot kansas primary test.json](https://github.com/BrightSpots/rcv/files/13252204/clear.ballot.kansas.primary.test.json)

**What happened:** RCTab failed to auto-load candidates until I unchecked 'Max' checkbox for 'Maximum Number of Candidates that can be ranked' Winning Rules setting. 
**What I would expect to happen:** RCTab would auto-load candidates successfully as long as the CVR was properly configured.

### First Attempt - Not Fully Configured
I configured my contest tab by tab. 
- Set a contest name
- Selected the `clear_ballot_kansas_primary.cvr.csv` file, Contest Id: `Democratic Primary` and added that cvr
- Click Auto-Load and got the following error 
<img width="868" alt="image" src="https://github.com/BrightSpots/rcv/assets/8996546/dfcbefd5-8747-47e9-bc4a-4d5edeccea6f">

### Second Attempt - Fully Configured
I thought maybe I needed to fully configure my contest so I tried again. I configured my contest tab by tab. 
- Set a contest name
- Selected the `clear_ballot_kansas_primary.cvr.csv` file, Contest Id: `Democratic Primary` and added that cvr
- Selected bottoms up with 15% threshold. **Maximum checkbox is checked by default**
<img width="870" alt="image" src="https://github.com/BrightSpots/rcv/assets/8996546/7eef45d7-4968-4a9c-bcbf-d8f584687d39">

- always skip to next rank overvote rule
- clicked 'Auto-Load Candidates' and got the same error
```
2023-11-03 12:04:04 EDT INFO: Auto-loading candidates from CVR files...
2023-11-03 12:04:11 EDT SEVERE: Rank: 1 exceeds max rankings allowed in config: 0
2023-11-03 12:04:11 EDT SEVERE: Failed to read source file "/Users/yezr/Downloads/clear ballot kansas primary cvr box id and precinct id swapped.csv": 
2023-11-03 12:04:11 EDT INFO: Auto-loaded 0 candidates.
```

### Third Attempt - Unchecked 'Maximum'
[...]

PR #776, opened 2023-12-12
fix Auto-Loading candidates fails when 'Maximum Number of Candidate... (#766)

mesutgk15:
…That Can Be Ranked' setting is set to 'Maximum' #766 

- ContestConfig.getMaxRankingsAllowed() returns '0' when Maximum option is checked, so i checked it from RawConfig
- A null check has been added to inner condition because the same method returns null when maximum option was checked but the input left blank

### DIFF
--- a/src/main/java/network/brightspots/rcv/ClearBallotCvrReader.java
+++ b/src/main/java/network/brightspots/rcv/ClearBallotCvrReader.java
@@ -82,11 +82,14 @@ class ClearBallotCvrReader extends BaseCvrReader {
           choiceName = Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL;
         }
         Integer rank = Integer.parseInt(choiceFields[RcvChoiceHeaderField.RANK.ordinal()]);
-        if (rank > this.config.getMaxRankingsAllowed()) {
-          Logger.severe(
-              "Rank: %d exceeds max rankings allowed in config: %d",
-              rank, this.config.getMaxRankingsAllowed());
-          throw new CvrParseException();
+        if (!this.config.getRawConfig().rules.maxRankingsAllowed.equals("max")) {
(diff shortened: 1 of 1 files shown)


tarheel (commented):
Interesting. So... it's a little weird that we'd throw this error here at all, given that the user is simply trying to load the candidate names from the CVR, not actually run the tabulation. In a typical case, they'd only be running into the error because the candidates haven't been populated yet (which is what they're trying to do). @artoonie do you think it would be better to just avoid throwing this error in this context? In the unlikely event that there's actually a problem with the rankings, it will still be detected when they run the tabulation.

artoonie:
I agree. Given that this check only happens in one reader and not the others, I think it makes sense to just remove this entirely -- especially knowing that this would be caught in tabulation (which perhaps we should double-check?).

yezr:
i think a problem is that it defaults to `getNumDeclaredCandidates()` when parsing out an actual int for the `getMaxRankingsAllowed()`. `getNumDeclaredCandidates()` is 0 before you have loaded any candidates.  
```Integer getMaxRankingsAllowed() {
    return stringToIntWithOption(
        getMaxRankingsAllowedRaw(),
        MAX_RANKINGS_ALLOWED_NUM_CANDIDATES_OPTION,
        getNumDeclaredCandidates());
  }
```

I like the exception here if there _really_ are marked ballots with ranks that exceed the configured limit. Because that is a failure reading CVRs - it is an invalid CVR given the configuration.

getMaxRankingsAllowed() has some other uses though, and I don't know if actually returning something like int.max() when the 'Maximum' checkbox is set would change other behavior...

yezr:
I did confirm that this code change allows the auto-load to work successfully. It just seems clunky that you need to explicitly check for the `max` string in the additional check instead of relying on `config.getMaxRankings()` to return int.max when you are configured in the UI to 'Maximum'

Issue #777, opened 2023-12-13
[2.0-RC] CVR Overvotes show up as blank in RCTab_CVR


yezr:
RCTab Version: 1.4.0
Config file: [overvote test 1.2.json](https://github.com/BrightSpots/rcv/files/13663406/overvote.test.1.2.json)
CVR file: [Book2.xlsx](https://github.com/BrightSpots/rcv/files/13663411/Book2.xlsx)

## What happened
Testing how RCTab handles overvotes I loaded the CVR that @chughes297 whipped up. I compared the [simple_cvr](https://github.com/BrightSpots/rcv/issues/753) to the original CVR and noticed that what was marked explicitly as an overvote in the CVR was blank in the simple_cvr
![image](https://github.com/BrightSpots/rcv/assets/8996546/212a681c-716a-4ccd-8131-ea015d882a98)

## What I Expected
I would expect that the overvote be marked in the simple_cvr. In this example, basically for it to match exactly the text of the CVR itself. For example, I would expect 'overvote' in 

- Round 1 for CVRs 1-4
- Round 2 for CVRs 11 & 12

yezr:
I looked into the code because I was curious and found some stuff that might point the future fixer of this issue in the right direction! Once I knock the rust off my IDE in the future maybe I could just fix it while I'm in there...
`selection = config.getNameForCandidate(selection);` returns null for 'overvote' in ResultsWriter line 643
`contestconfig processCandidateData` probably needs to add an 'overvote' -> 'overvote' like for UWI

yezr:
Now that 2.0 creates this file everytime once #753 is complete, I think we should ensure that the output is correct. Tagged w/ `needs-rcvrc-clarification` to confirm

artoonie:
#753 didn't fix this before, but has now been updated to fix this

PR #778, opened 2023-12-14
fix CVR Overvotes show up as blank in simple_cvr (#777)

mesutgk15:
CVR Overvotes was blank in the simple_cvr. With the fix, they are marked now in the generated file as in below (#777): 

Screenshot 2023-12-14 at 17 26 52

### DIFF
--- a/src/main/java/network/brightspots/rcv/ContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/ContestConfig.java
@@ -1129,6 +1129,11 @@ class ContestConfig {
       candidateAliasesToNameMap.put(
           Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL, Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL);
     }
+
+    if (isOverVoteLabelEnabled()) {
+      candidateAliasesToNameMap.put(
+              Tabulator.EXPLICIT_OVERVOTE_LABEL, Tabulator.EXPLICIT_OVERVOTE_LABEL);
+    }
   }
 
   private boolean undeclaredWriteInsEnabled() {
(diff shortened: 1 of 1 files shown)


tarheel:
@mesutgk15 thanks for all of the PRs! Just curious: how did you come across this project? Are you using this tool for real elections?

mesutgk15:
Hi @tarheel, you are welcome. No I just came across to your project while I was looking for the repositories that I can contribute, trying to improve myself with open source contributions.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1143,6 +1148,17 @@ private boolean undeclaredWriteInsEnabled() {
     return includeUwi;
   }
 
+  private boolean isOverVoteLabelEnabled() {
+    boolean overVote = false;
```
tarheel, line 1152: super nit, but elsewhere in the code we refer to it using a single word ("overvote"), so these should be `isOvervoteLabelEnabled` and `overvote`

tarheel (commented):
Looks right to me, but I'll admit that I don't remember this part of the code very well.

@artoonie we don't have an existing way to create a unit test to confirm the output in the generated generic CVR CSV, do we?

artoonie:
> @artoonie we don't have an existing way to create a unit test to confirm the output in the generated generic CVR CSV, do we?

I believe we do here: https://github.com/BrightSpots/rcv/blob/develop/src/test/java/network/brightspots/rcv/TabulatorTests.java#L240 (also, while we're looking: I'm noticing the function's comment is wrong. This function is not run before every test.)

tarheel:
> > @artoonie we don't have an existing way to create a unit test to confirm the output in the generated generic CVR CSV, do we?
> 
> I believe we do here: https://github.com/BrightSpots/rcv/blob/develop/src/test/java/network/brightspots/rcv/TabulatorTests.java#L240 (also, while we're looking: I'm noticing the function's comment is wrong. This function is not run before every test.)

Oh, great. @mesutgk15 could you add a unit test that would fail without your fix here?

@artoonie I think that comment is saying that we generate the simple CVR(s) before every tabulation. Maybe it just needs to be clarified?

Issue #782, opened 2024-01-08
[2.0-RC] Multi-Winner summary file miscalculates active ballots after first winner


yezr:
RCTab Version: 1.4.0-alpha
CVR: [vmwg cvr.xlsx](https://github.com/BrightSpots/rcv/files/13862763/vmwg.cvr.xlsx)
Config: [vmsg surplus transfer test.json](https://github.com/BrightSpots/rcv/files/13862765/vmsg.surplus.transfer.test.json)
summary.csv: [2023-12-08_21-31-54_summary.csv](https://github.com/BrightSpots/rcv/files/13862776/2023-12-08_21-31-54_summary.csv)

[This google sheet](https://docs.google.com/spreadsheets/d/1CQIzA-xmL2UQLSlccKB1YVAedrnJUdy2/edit#gid=1136353935) has the summary csv loaded into the `[1.4.0] summary.csv` sheet with some comments related to this issue. 

## What Happened
I was checking some VMSG logic in [this spreadsheet](https://docs.google.com/spreadsheets/d/1CQIzA-xmL2UQLSlccKB1YVAedrnJUdy2/edit?usp=sharing&ouid=101762621909074220355&rtpof=true&sd=true). It was testing exactly how surplus fraction math works. 

When reviewing the RCTab summary.csv for that example I noticed that the % of vote column for Round 2 (F21:26) was adding up to above 100%. Doing the math explicitly I found a discrepancy in the 'Active Ballots' cell for Round 2 (E27). That cell shows 20.9999 Active ballots. But, if I sum up the Round 2 votes by hand (E21:E26) I get 28.9999. I think the % of vote Column (F) is using this as an incorrect divisor. 

<img width="1416" alt="image" src="https://github.com/BrightSpots/rcv/assets/8996546/8cf5a0f7-ad19-4001-908e-8f282fb0bc7a">

## What I Would Expect
I would expect the Round 2 'Active Ballots' cell (E27) to be 28.9999, always the sum of the individual candidate vote totals for that round (rows 21 to 26). I would expect the '% of vote' calculations in F21:26 to use that correct number as the divisor. 

I think that it is excluding votes that count towards an already elected candidate. You can see in Round 6, after all three candidates are elected and any votes over the threshold are inactivated that the Active Ballots = 0 in cell Q27.

PR #784, opened 2024-01-15
track "locked in" ballots -- not active, nor inactive

### LINKED ISSUE #783: [2.0-RC] Multi-Winner summary file doesn't display "% of Vote" for final round
RCTab Version: 1.4.0-alpha
CVR: [vmwg cvr.xlsx](https://github.com/BrightSpots/rcv/files/13862763/vmwg.cvr.xlsx)
Config: [vmsg surplus transfer test.json](https://github.com/BrightSpots/rcv/files/13862765/vmsg.surplus.transfer.test.json)
summary.csv: [2023-12-08_21-31-54_summary.csv](https://github.com/BrightSpots/rcv/files/13862776/2023-12-08_21-31-54_summary.csv)

[This google sheet](https://docs.google.com/spreadsheets/d/1CQIzA-xmL2UQLSlccKB1YVAedrnJUdy2/edit#gid=1136353935) has the summary csv loaded into the [1.4.0] summary.csv sheet with some comments related to this issue.

## What happened
Reviewing the `summary.csv` I noticed that the final Round doesn't have any numbers for the "% of Vote" column (R21:26). 

image

## What I would expect
R21:26 would have "% of Vote" numbers just like each Round before it. I would expect the summary file to show 33.33% (or something super close to that with some rounding) for each of the winning candidates A, B & C.

artoonie:
closes #783 and #782

The "Active Ballots" in STV does not include ballots that are "locked in" with a previous candidate -- only the ballots "still in play".

That led to incorrect vote percentages reported out, and on the final round, the number of active ballots was always zero and therefore all vote percentages were 0 (or, inf, but reported as zero).

This tracks those ballots as a "LOCKED_IN" ballot. Would love to gather feedback on this terminology. I've opted to only track "locked in" ballots internally, and never externally, since the distinction is not useful to a third party.


yezr:
During our conversation working through the VMWG contest that these issues came from we used a couple different terms so this is a good conversation to have :) We started with "non-transfer value." Then, I started using "remains to elect" to be more explicit _why_ it is not transferred. By the end we began using the term "held" to describe the portion of someone's ballot that remains allocated towards electing an already elected candidate. 

I do like some part of the descriptor explicitly stating that it is allocated towards electing someone, despite it usually requiring a longer descriptor. Maybe `locked_in_to_elect` or `held_to_elect`? locked_in and held work equally well for me. 

@chughes297 do you have any preferences here

artoonie:
Do you want this reported externally too in the "simple CSV"?

yezr:
> Do you want this reported externally too in the "simple CSV"?

Do you mean separately reporting the locked_in ballot portions in the summary? Don't want that. What you've implemented - including it as part of the "Active Ballots" as it is currently reported out in the summary file - is the correct way to go.

artoonie:
Great, thanks for confirming!

yezr:
I built locally and ran the VMWG contest. The vote percentages in each round are correct and the final round now also shows a vote percent! 

However, I noticed that the held_to_elect ballots look like they are now showing up in the 'Inactive Ballots Total' row. In D33 the transfer value for 'Inactive Ballots Total' going from Round 1 to Round 2 shows 8 - the same number of ballots that now are officially held to elect the Round 1 elected winner Candidate C. 

The 'Inactive Ballots Total' row was correct in the original `summary.csv`. [This google sheet](https://docs.google.com/spreadsheets/d/1CQIzA-xmL2UQLSlccKB1YVAedrnJUdy2/edit#gid=1136353935) has the summary csv loaded into the [1.4.0] summary.csv sheet. This PR seems to put the held_to_elect values incorrectly into that row.

artoonie:
Agree that makes more sense to have inactive ballots not include "locked in" ballots. Updated to match.

yezr:
Looks good!

src/main/java/network/brightspots/rcv/CastVoteRecord.java
```diff
@@ -213,6 +213,7 @@ Map<String, BigDecimal> getWinnerToFractionalValue() {
   // it is active or not.
   enum StatusForRound {
     ACTIVE,
+    LOCKED_IN,
```
tarheel, line 216: I would update the comment above to be more explicit about what this new status value means. In particular, the comment above ("StatusForRound represents the ballot's status on a given round") is confusing because this new status value might apply to only a portion of a given ballot.

More generally: if I'm not mistaken, all of the existing `StatusForRound` values have been used to refer to an entire CVR, so this new status is both introducing a new type of status and changing the meaning of the other statuses. That's potentially fine, but I'm wondering whether there are dependencies that we need to update accordingly. E.g. does the behavior in `recordSelectionForCastVoteRecord` still make complete sense? At a minimum, I imagine that we need to add some clarifying comments in various places and/or rename some things.

src/main/java/network/brightspots/rcv/RoundTally.java
```diff
@@ -135,13 +142,19 @@ BigDecimal getBallotStatusTally(StatusForRound statusForRound) {
...
 
-  // Get the number af active ballots in this round
+  // Get the number of active ballots in this round
   BigDecimal numActiveBallots() {
     ensureFinalized();
     return numActiveBallots;
   }
 
-  // Get the number af active ballots in this round
+  // Get the number of "locked-in" ballots in this round
```
tarheel, line 151: clarify that this can be fractional? also, the comment is inaccurate

src/main/java/network/brightspots/rcv/RoundTally.java
```diff
@@ -135,13 +142,19 @@ BigDecimal getBallotStatusTally(StatusForRound statusForRound) {
...
+  BigDecimal numActiveOrLockedInBallots() {
+    ensureFinalized();
+    return numActiveBallots.add(ballotStatusTallies.get(StatusForRound.LOCKED_IN));
+  }
+
+  // Get the number of inactive ballots in this round
```
tarheel, line 157: so this can now also be fractional

tarheel (commented):
The change in app behavior makes sense to me. Just some questions about semantics...

artoonie:
@tarheel agree that this was changing the meaning of StatusForRound too much. I opted to revert that and just track "locked in" ballots separately. Does this feel cleaner to you now?

src/main/java/network/brightspots/rcv/RoundTally.java
```diff
@@ -135,13 +142,19 @@ BigDecimal getBallotStatusTally(StatusForRound statusForRound) {
...
 
-  // Get the number af active ballots in this round
+  // Get the number of active ballots in this round
   BigDecimal numActiveBallots() {
     ensureFinalized();
     return numActiveBallots;
   }
 
-  // Get the number af active ballots in this round
+  // Get the number of active ballots plus the fractional amount of "locked in" ballots.
```
tarheel, line 151: Confirming my understanding: a single CVR could contribute to both the active count and the locked-in count, correct? Like, 80% of it could be locked in while the remainder is still active? In other words, `numActiveBallots` is actually the sum of the fractional transfer values of all of the active ballots... right? Because if we were just counting the discrete number of active ballots and then adding the locked-in sum, we could end up double-counting.

tarheel (approved):

src/main/java/network/brightspots/rcv/RoundTally.java
artoonie: Yes, I believe your understanding is accurate. `numActiveBallots` might be better named `activeBallotSum` to avoid the confusion here? (and similar for `numInactiveBallots`)

tarheel: Yeah, I agree.

PR #785, opened 2024-01-17
Combine all Generic CSV exported files into one large CSV

### LINKED ISSUE #753: Reimagining Dominon Conversion file as rctab_cvr
Further reflection of the bug #734 and its corresponding PR #751 forced the conversation of what exactly we want out of this file, currently labeled `dominion_conversion`. The result changes a little how we want the file to work. It will make #734 moot and fix #735 

I removed all of these related issues from 1.3.2. This file isn't generated for Hart in 1.3.x, which is the only vendor that 1.3.x will be used for. 

### Goal
So we do we want from this file? We want to create a list of all input CVRs stripped only to their essential information. Those essentials will provide two things. 
- **Reproducibility** anyone should be able to take the information in there and use it to recreate RCTab's results. 
- **Transparency** Those essentials will also provide more transparency into how exactly we interpreted the raw CVRs. This will aide troubleshooting and auditability. 

### What needs to change from the current implementation
- Let's refer to this file as the `simple_cvr` format. (Open to other ideas. Also discussed was `basic_cvr` or `essential_cvr`) To fix #735  the file should be named `[dateTime]_simple_cvr_conversion_[contestId|contestName]`. If we don't have `contestId` can we use the `contestName` from the Contest Info tab. 
- There should be a single file that contains all of the input CVRs, regardless of the number of input files that they come from. This will make #734 moot since all input CVRs will be appended to the same file.
- To help with auditability now that all input CVRs are in one file, let's add a new column to the `simple_cvr` file: `input_file`. This will contain the full filename and extension of the input file that the CVR was converted from.

### LINKED ISSUE #777: [2.0-RC] CVR Overvotes show up as blank in RCTab_CVR

artoonie:
Closes #753 

1. Adds the source name to the "RCTab CVR CSV"
2. Removes the term "overvotes" and instead includes newline-separated values in case of an overvote
3. Resolve #777

### DIFF
--- a/src/main/java/network/brightspots/rcv/ResultsWriter.java
+++ b/src/main/java/network/brightspots/rcv/ResultsWriter.java
@@ -91,13 +91,10 @@ class ResultsWriter {
       String timestampString,
       String sequentialTabulationId) {
     outputType = sanitizeStringForOutput(outputType);
-    String sequence = sequentialTabulationId == null
-            ? ""
-            : sequentialSuffixForOutputPath(sequentialTabulationId);
     String fileName =
         String.format(
             "%s_%s%s",
-            timestampString, outputType, sequence);
+            timestampString, outputType, sequentialSuffixForOutputPath(sequentialTabulationId));
     return Paths.get(outputDirectory, fileName).toAbsolutePath().toString();
   }
 
@@ -600,6 +597,10 @@ class ResultsWriter {
           // we've moved on to a new contest, so we need to switch to the next source
           currentSourceIndex++;
           currentSourceData = perSourceDataForCsv.get(currentSourceIndex);
+
+          if (currentSourceData.sourceIndex != currentSourceIndex) {
+            throw new RuntimeException("Source list must be sorted by sourceIndex!");
+          }
         }
 
         CastVoteRecord castVoteRecord = castVoteRecords.get(i);
@@ -648,25 +649,23 @@ class ResultsWriter {
     for (int rank = 1; rank <= maxRanks; rank++) {
       if (castVoteRecord.candidateRankings.hasRankingAt(rank)) {
         CandidatesAtRanking candidates = castVoteRecord.candidateRankings.get(rank);
-        if (candidates.count() >= 1) {
-          // We list all candidates at a given ranking on separate lines
-          // This allows algorithms which accept overvotes.
-          List<String> allCandidatesAtRanking = new ArrayList<>(candidates.count());
-          for (String candidate : candidates) {
-            String selection = candidate;
-            // We map all undeclared write-ins to our constant string when we read them in,
-            // so we need to translate it back to the original candidate ID here.
-            if (selection.equals(Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL)) {
-              selection = undeclaredWriteInLabel;
-            } else if (selection.equals(Tabulator.EXPLICIT_OVERVOTE_LABEL)) {
-              selection = Tabulator.EXPLICIT_OVERVOTE_LABEL;
-            } else {
-              selection = config.getNameForCandidate(selection);
-            }
-            allCandidatesAtRanking.add(selection);
+        // We list all candidates at a given ranking on separate lines
+        // This allows algorithms which accept overvotes.
+        List<String> allCandidatesAtRanking = new ArrayList<>(candidates.count());
+        for (String candidate : candidates) {
+          String selection = candidate;
+          // We map all undeclared write-ins to our constant string when we read them in,
+          // so we need to translate it back to the original candidate ID here.
+          if (selection.equals(Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL)) {
+            selection = undeclaredWriteInLabel;
+          } else if (selection.equals(Tabulator.EXPLICIT_OVERVOTE_LABEL)) {
+            selection = Tabulator.EXPLICIT_OVERVOTE_LABEL;
+          } else {
+            selection = config.getNameForCandidate(selection);
           }
-          csvPrinter.print(String.join("\n", allCandidatesAtRanking));
+          allCandidatesAtRanking.add(selection);
         }
+        csvPrinter.print(String.join("\n", allCandidatesAtRanking));
       } else {
         csvPrinter.print("undervote");
       }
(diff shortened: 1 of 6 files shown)


src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -91,10 +91,13 @@ static String getOutputFilePath(
...
       String sequentialTabulationId) {
     outputType = sanitizeStringForOutput(outputType);
+    String sequence = sequentialTabulationId == null
+            ? ""
+            : sequentialSuffixForOutputPath(sequentialTabulationId);
     String fileName =
         String.format(
             "%s_%s%s",
-            timestampString, outputType, sequentialSuffixForOutputPath(sequentialTabulationId));
+            timestampString, outputType, sequence);
```
tarheel, line 100: isn't this equivalent?

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -538,33 +541,24 @@ void generateOverallSummaryFiles(
     generateSummaryJson(roundTallies, tallyTransfers, null, outputPath);
   }
 
-  // write CastVoteRecords for the specified contest to the provided folder
+  // write CastVoteRecords for the specified contest to the provided folder,
```
tarheel, line 544: So now it's to a single file in the provided folder. In this case, should the user just specify a full file path instead of a folder?

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -575,19 +569,41 @@ String writeGenericCvrCsv(
...
+      // we can use the PerSourceDataForCsv to determine which source each record came from.
+      int currentSourceIndex = 0;
+      PerSourceDataForCsv currentSourceData = perSourceDataForCsv.get(currentSourceIndex);
+
       // print rows:
-      for (CastVoteRecord castVoteRecord : castVoteRecords) {
+      for (int i = 0; i < castVoteRecords.size(); i++) {
+        if (i > currentSourceData.lastIndexInCvrList) {
+          // we've moved on to a new contest, so we need to switch to the next source
+          currentSourceIndex++;
```
tarheel, line 601: This feels a tiny bit fragile even if it's working right now because it assumes that the the ordering of the sources and the CVRs is stable.
 
Would it be more robust to store the source ID in the CVR object so we don't have to infer it from the CVR's index?

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -631,18 +648,24 @@ private void printRankings(
...
         CandidatesAtRanking candidates = castVoteRecord.candidateRankings.get(rank);
-        if (candidates.count() == 1) {
-          String selection = candidates.get(0);
-          // We map all undeclared write-ins to our constant string when we read them in,
-          // so we need to translate it back to the original candidate ID here.
-          if (selection.equals(Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL)) {
-            selection = undeclaredWriteInLabel;
-          } else {
-            selection = config.getNameForCandidate(selection);
+        if (candidates.count() >= 1) {
```
tarheel, line 651: why is this condition necessary if you've already checked `hasRankingAt`?

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -348,6 +341,19 @@ private List<CastVoteRecord> parseCastVoteRecords(ContestConfig config) {
       }
     }
 
+    // Output the RCTab-CSV CVR
```
tarheel, line 344: Is this just code reorganization or is there a more substantive reason that you moved this down a bit? A little hard to tell just from viewing this here.

src/main/java/network/brightspots/rcv/ResultsWriter.java
artoonie: This is run automatically on tabulation, where the user specifies a folder for all output files.

src/main/java/network/brightspots/rcv/ResultsWriter.java
artoonie: Yes, though it would add an unnecessary sort -- how do you feel about this update: assume it's stable, but verify and throw a halting exception if it's not?

src/main/java/network/brightspots/rcv/TabulatorSession.java
artoonie: Before it was inside the for-loop (i.e. per-source); now it's once per tabulator session

tarheel (approved):

src/main/java/network/brightspots/rcv/ResultsWriter.java
artoonie: I guess there's two things we need to verify:
1. `List<PerSourceDataForCsv> perSourceDataForCsv` hasn't had its order changed, since the field `lastIndexInCvrList` assumes its order is constant
2. `List<PerSourceDataForCsv>` and `List<CastVoteRecord>` are "in sync" with each other

My change addresses (1) but not (2).

I'm hesitant to add the source index to the CastVoteRecord because it would require being passed from the TabulatorSession down to the Reader and be implemented in each Reader individually. I can see it perhaps being useful to actually include a reference to the Source itself (but not its index?) in `CastVoteRecord`, which could even be used in lieu of `computedId` if we also included a line number. That feels broader than needed for the scope of this PR, though.

My justification is this: `perSourceDataForCsv` is explicitly and exclusively a structure intended to help index into a `List<CastVoteRecord>`. I've added a comment to specify this above writeRctabCvrCsv.

tarheel: OK, this all sounds fine for now.

PR #786, opened 2024-01-23
check if ranking is allowed rather than using unacceptable default

### LINKED ISSUE #766: [2.0-RC] Auto-Loading candidates fails when 'Maximum Number of Candidates That Can Be Ranked' setting is set to 'Maximum'

artoonie:
Resolves #766 

Auto-loading candidates failed when Max Rankings was set to "Max," because rather than returning a value like INT_MAX, we'd return NumCandidates (which was likely 0).

I think it's best to not return this value at all, but to allow the config to check it directly. In this PR, I replace `getMaxRankingsAllowed()` with `isRankingAllowed(int rank)` throughout.

The only change in functionality is when creating the CVR CSV, and only in the case that the Dominion max ranking is different than the one entered in the config. One test has been updated to match this. (Note: this would have changed anyway when we created the combined CVR CSV in #785.)

For completeness, I also removed the check that ClearBallot was doing that no other reader did (as discussed in #776).

### DIFF
--- a/src/main/java/network/brightspots/rcv/BaseCvrReader.java
+++ b/src/main/java/network/brightspots/rcv/BaseCvrReader.java
@@ -46,7 +46,15 @@ abstract class BaseCvrReader {
 
   // Any reader-specific validations can override this function.
   public void runAdditionalValidations(List<CastVoteRecord> castVoteRecords)
-      throws CastVoteRecord.CvrParseException {}
+      throws CastVoteRecord.CvrParseException {
+    for (CastVoteRecord cvr : castVoteRecords) {
+      for (Pair<Integer, CandidatesAtRanking> ranking : cvr.candidateRankings) {
+        if (!this.config.isRankingAllowed(ranking.getKey())) {
+          throw new CastVoteRecord.CvrParseException();
+        }
+      }
+    }
+  }
 
   // Gather candidate names from the CVR that are not in the config.
   Map<String, Integer> gatherUnknownCandidates(List<CastVoteRecord> castVoteRecords) {
--- a/src/main/java/network/brightspots/rcv/BaseCvrReader.java
+++ b/src/main/java/network/brightspots/rcv/BaseCvrReader.java
@@ -40,8 +40,8 @@ abstract class BaseCvrReader {
       throws CastVoteRecord.CvrParseException, IOException;
 
   // Individual contests may have a different value than what the config allows.
-  public Integer getMaxRankingsAllowed(String contestId) {
-    return config.getMaxRankingsAllowed();
+  public boolean isRankingAllowed(int rank, String contestId) {
+    return config.isRankingAllowed(rank);
   }
 
   // Any reader-specific validations can override this function.
--- a/src/main/java/network/brightspots/rcv/ResultsWriter.java
+++ b/src/main/java/network/brightspots/rcv/ResultsWriter.java
@@ -542,7 +542,8 @@ class ResultsWriter {
   // returns the filepath written
   String writeGenericCvrCsv(
       List<CastVoteRecord> castVoteRecords,
-      Integer numRanks,
+      CvrSource source,
+      BaseCvrReader reader,
       String csvOutputFolder,
       String inputFilepath,
       String contestId,
@@ -581,7 +582,11 @@ class ResultsWriter {
       csvPrinter.print("Record Id");
       csvPrinter.print("Precinct");
       csvPrinter.print("Precinct Portion");
-      for (int rank = 1; rank <= numRanks; rank++) {
+      int maxRank = 0;
+      for (CastVoteRecord castVoteRecord : castVoteRecords) {
+        maxRank = Math.max(maxRank, castVoteRecord.candidateRankings.maxRankingNumber());
+      }
+      for (int rank = 1; rank <= maxRank; rank++) {
         String label = String.format("Rank %d", rank);
         csvPrinter.print(label);
       }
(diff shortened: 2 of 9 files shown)


yezr:
> For completeness, I also removed the check that ClearBallot was doing that no other reader did (as discussed in https://github.com/BrightSpots/rcv/pull/776).

Was this put there because we had some ClearBallot CVRs that showed the invalid behavior this was checking for? It looks like nothing in the tests have that behavior (or else it would trigger an exception with this check in there). But maybe at one time we had a CVR exhibiting that behavior? 

We discussed in #776 we should make sure that if the ClearBallot CVR reader doesn't catch it, that the tabulation does. That almost would be preferred, because then we have the check in the path that all vendors must successfully go through. I can take on the task of trying to create a ClearBallot CVR that does that to confirm that is the case.

yezr:
And #776 was already closed manually, but just confirming that this PR fixes the underlying issue (#766) that #776 was trying to fix

artoonie:
> Was this put there because we had some ClearBallot CVRs that showed the invalid behavior this was checking for? It looks like nothing in the tests have that behavior (or else it would trigger an exception with this check in there). But maybe at one time we had a CVR exhibiting that behavior?

I'm not sure. We only have one clearballot example and it doesn't hit this error.

> We discussed in #776 we should make sure that if the ClearBallot CVR reader doesn't catch it, that the tabulation does. That almost would be preferred, because then we have the check in the path that all vendors must successfully go through. I can take on the task of trying to create a ClearBallot CVR that does that to confirm that is the case.

Perhaps this warrants more discussion -- no other reader throws an error when there are more than the allowed # of votes. A cursory glance indicates that other readers simply ignore any candidate beyond the allowed number of ranks.

I've opted to (1) copy that paradigm, and ignore instead of throwing an error, and (2) check this explicitly in the tabulator in case any reader fails to follow the same paradigm.

Other options:
1. Ensure all readers throw a fatal error when an out-of-bounds rank is provided (instead of silently ignoring)
2. Leave as-is, where only clearballot throws such an error. (Maybe there's a reason for this I don't see?)



> And #776 was already closed manually, but just confirming that this PR fixes the underlying issue (#766) that #776 was trying to fix

Confirmed

yezr:
I don't imagine that a ballot will ever be made that allows a voter to vote for more than what the administrators have determined as their limit. However, I will still add a ticket for us to decide what RCTab should do in that case: automatically limit ballots to the configured maximum or throw a halting exception. 

This is enough to resolve #766 and we can address the above question according to priority.

src/main/java/network/brightspots/rcv/BaseCvrReader.java
```diff
@@ -40,13 +40,21 @@ abstract void readCastVoteRecords(List<CastVoteRecord> castVoteRecords)
...
   public void runAdditionalValidations(List<CastVoteRecord> castVoteRecords)
-      throws CastVoteRecord.CvrParseException {}
+      throws CastVoteRecord.CvrParseException {
+    for (CastVoteRecord cvr : castVoteRecords) {
+      for (Pair<Integer, CandidatesAtRanking> ranking : cvr.candidateRankings) {
+        if (!this.config.isRankingAllowed(ranking.getKey())) {
+          throw new CastVoteRecord.CvrParseException();
+        }
+      }
+    }
```
tarheel, line 56: is this slow for a large source?

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -624,11 +631,16 @@ String writeGenericCvrCsv(
...
+      BaseCvrReader reader,
+      CvrSource source,
       CSVPrinter csvPrinter,
       CastVoteRecord castVoteRecord)
       throws IOException {
     // for each rank determine what candidate id, overvote, or undervote occurred and print it
     for (int rank = 1; rank <= maxRanks; rank++) {
+      if (!reader.isRankingAllowed(rank, source.getContestId())) {
+        break;
+      }
```
tarheel, line 643: could you clarify which case this is handling?

tarheel (approved):

artoonie (commented):
(I wrote this comment weeks ago but forgot to Submit Review -- submitting now for posterity)

PR #788, opened 2024-01-23
UWIs must always be eliminated first

### LINKED ISSUE #397: eliminate UWIs first even if a candidate wins in the first round

artoonie:
Closes #397 

No winner can be selected while UWIs are still active


yezr:
confirmed that running the `multi_seat_UWI_test` that this fixes the issue: UWI is eliminated first in STV regardless of how many votes they have

tarheel (approved):

PR #790, opened 2024-01-24
Build Security: airgapped caching and golden hash generation

### LINKED ISSUE #756: Confirm we can verify dependencies during trusted build process.
Part of getting certified in CA is having a trusted build process. CASOS makes a fresh build from source and then distributes that artifact to jurisdictions. 

Part of that process has been verifying the third party dependencies we use. During the 1.3.1 trusted build process each of the dependencies were verified manually. Can we brainstorm a way to make this dependency verification programmatic? 

One option is [verifying dependencies with gradle](https://docs.gradle.org/current/userguide/dependency_verification.html#:~:text=The%20verification%20metadata%20file&text=Doing%20so%2C%20Gradle%20will%20verify,%E2%80%8B). It looks like you can '[bootstrap](https://docs.gradle.org/current/userguide/dependency_verification.html#sec:bootstrapping-verification)' a set of hash checks for each dependency. After the initial bootstrapping, hash checking happens automatically when building with gradle.

### LINKED ISSUE #759: Include hash of 'Trusted Build' from source in release artifacts
The RCTab TDP includes these instructions for creating and verifying a build from from source

> Note, however, that two independent runs of the jlinkZip task will result in ZIP files with different hashes! Although the file contents in the two ZIP files will be identical, the timestamps of those files will not be. Those timestamps are included in the hash and thus, the hashes will not match. As a result, generating a hash for the ZIP file itself is only suitable for verifying that this build has not been modified.

And continues

> It is, however, possible to demonstrate that two builds are equivalent by stripping the timestamps.
> 
> The process for generating a single timestamp-independent hash signature for a build ZIP requires a few steps: not only does the ZIP file store timestamps for the files it contains, but one of the files within the ZIP file, lib/modules, is, in turn, a collection of time stamped files. Fully decomposing the build to its individual constituent files thus requires using the jimage utility (part of the Java 17 development kit) to extract the files contained within lib/modules.
> 
> - First, unzip the ZIP you built locally (by following the precise instructions in the “Full build” section above). For the rest of this example we’ll assume that you extracted the zip files to `c:\rctab_v1.3.2_windows\` 
> - Navigate to the top-level rcv directory (for our example `c:\rctab_v1.3.2_windows\rcv\`). 
>   - *Make sure not to execute any other commands that create, modify, or delete files within this directory (like opening the application). Any modifications to these files will change the result of the hashing process.
> - Create a hash.bat file in the `c:\rctab_v1.3.2_windows\rcv\` folder with the following command `echo.>hash.bat`
> - Open `hash.bat` in notepad and paste in the text below these bullets (starting with ::hash.bat)
> - Run the batch file in the cmd prompt by typing `hash.bat` then enter. **It can take 30-45 minutes to complete as it is computing a hash for every single file**
> ```
> :: hash.bat
> @echo off
> 
> :: NOTE: This script must be placed one level up from the rcv directory
> 
> echo Initiating batch hash procedure...
> echo %date% %time%
> 
> setlocal EnableExtensions EnableDelayedExpansion
> 
> set "HASHFILE=all_hashes.txt"
> set "TEMPHASHFILE=all_hashes_temp.txt"
> set "EXTRACTIONDIR=.\rcv\modules_extracted"
> set "MODULESFILE=.\rcv\lib\modules"
> 
> if exist %HASHFILE% (
> 	echo Deleting existing hash file, %HASHFILE% ...
>     del %HASHFILE%
> )
> 
> if exist %EXTRACTIONDIR% (
> 	echo Deleting existing extracted modules directory, %EXTRACTIONDIR% ...
>     rmdir /s /q %EXTRACTIONDIR%
> )
> 
> echo Extracting contents of modules file...
> mkdir %EXTRACTIONDIR%
> cd %EXTRACTIONDIR%
> jimage extract ..\..\%MODULESFILE%
> 
> echo Temporarily relocating modules file...
> cd ..\..
> copy %MODULESFILE% .
> del %MODULESFILE%
> 
> :: Calculate the hash for every file here and in all subdirectories, appending to the file (format "(filename) = (hash)")
> echo Calculating hashes...
> for /r .\rcv %%f in (*) do (
>     <NUL set /p ="%%f = " >> %HASHFILE%
>     C:\Windows\System32\certutil.exe -hashfile "%%f" SHA512 | findstr /v ":" >> %HASHFILE%
>     echo hashing %%f
> )
> 
> echo Restoring modules file...
> move .\modules %MODULESFILE%
> 
> :: Replace the absolute paths to each file with relative paths (e.g. C:\temp\rcv => .\rcv)
> 
> echo Replacing absolute paths with relative paths in hash file...
> set "SEARCHTEXT=%cd%"
> set "REPLACETEXT=."
> for /f "delims=" %%A in ('type "%HASHFILE%"') do (
>     set "string=%%A"
>     set "modified=!string:%SEARCHTEXT%=%REPLACETEXT%!"
>     echo !modified!>>"%TEMPHASHFILE%"
> )
> del "%HASHFILE%"
> rename "%TEMPHASHFILE%" "%HASHFILE%"
> 
> echo Sorting the hash file...
> sort "%HASHFILE%" > "%TEMPHASHFILE%"
> del "%HASHFILE%"
> rename "%TEMPHASHFILE%" "%HASHFILE%"
> 
> echo Calculating the hash of the entire sorted hash file...
> C:\Windows\System32\certutil.exe -hashfile %HASHFILE% SHA512
> 
> endlocal
> ```
> Example output (Windows):
> ```
> SHA512 hash of all_hashes.txt:
> a3bc3f93a213f3542ca64d085a701174c154044a37e5c33c39adcbbc325e38c609ef1ae6f678bd34b4a477a85201d48dfec8b8b341b4c62b40f127bbfdadbb81
> CertUtil: -hashfile command completed successfully.
> ```

Currently, a new release does not contain this source-built hash to verify against. Instead, it contains a hash of each of the six build artifacts: one installer and one `.zip` for each of the three OSes we build for. 

Can we add an additional step in the release process to calculate the source-built hash and include it as an additional release artifact.

artoonie:
Closes #756: airgapped cache
Closes #759: hash of trusted build

This pull request does four things:
1. It caches all dependencies and zips them up, one cache zip per OS, so that RCTab can be built on airgapped machines with minimal other dependencies. (You still need Java and Gradle.)
2. It creates a "golden hash" for the jlinkZip. This is a reproducible value per OS, meaning it contains no timestamps in the zip.
3. The cache zip's version of a "golden hash" is in a file `checksums.csv`, which lists a checksum per dependency. That file has an MD5, a SHA-1, and a SHA-256, because we want to correspond our values to those on maven.org, and we don't know what sort of checksum is uploaded to maven.org (it varies by dependency).
4. Use .gitattributes to enforce LF line endings for cross-platform, cross-machine consistency

After code review, I will:
1. Increase the .zip retention to 90 days (it's set to 1 during testing)
2. Revert the change in .github/workflows/release.yml that creates a package on every push to this branch
3. Update the wiki to reflect the changes in generating the golden hash


yezr:
#759 has the `hash.bat` contents that previously anyone could use to verify the golden hash of their local copy. Has that local verification process changed? If so, can you detail that in #759

artoonie:
Yes, the `hash.bat` process has changed. After this, it will be sufficient to run ` ./sha-of-zip.sh <zipFilepath> <OS: Windows, MacOS, or Linux> <sha version: 1, 256, or 512>` or `./sha-of-zip.bat <ZIP_FILEPATH> <sha version: 1, 256, or 512>` to get the golden hash. I can update the wiki to reflect that when this is merged.

README.md
```diff
@@ -53,6 +53,22 @@ On Linux, you may install the .deb file, then run `/opt/rcv/bin/RCTab` to launch
...
+3. Download the appropriate cache files for your OS: cache.[OS].zip
+4. Stop the gradle daemon with `gradle --stop`
+5. Delete the directory ~/.gradle/caches if it exists
+6. Extract the appropriate caches/[filename].zip to ~/.gradle/caches so that the "caches" directory is in ~/.gradle
+7. In the extracted directory, you may manually verify each dependency using checksums.csv in accordance with your own policies
+8. Run `gradle assemble --offline` and ensure you get no errors
+9. Run `gradle run --offline` to launch RCTab
+10. Run `gradle jpackage --offline` to generate a jpackage package
+
+Note: this is not currently available on Silicon-chip Apple Devices. Only Intel Mac is supported.
```
tarheel, line 70: What a time to be alive.

tarheel (commented):
I'll let @HEdingfield tackle this one!

yezr:
I've edited the [wiki](https://github.com/BrightSpots/rcv/wiki/Golden-Hash-Generation) with steps on golden hash generation to also get a review!

artoonie:
Reviewed! Looks good. I made a couple cosmetic updates and added some details clarifying where instructions were Windows-only.

HEdingfield:
@artoonie I'm confused about why there are so many test files updated in this PR. It seems like there are no changes happening in the diff. Any idea what's up, and can we revert these?

artoonie:
@HEdingfield That was the line-ending fix (the fourth bullet point in the PR description)

README.md
artoonie: The future was only three months away!

build.gradle
```diff
@@ -23,6 +23,7 @@ dependencies {
     implementation 'com.fasterxml.jackson.core:jackson-annotations:2.15.2'
     implementation 'com.fasterxml.jackson.core:jackson-databind:2.15.2'
     implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-xml:2.15.2'
+    implementation "com.github.spotbugs:spotbugs:4.8.3"
```
HEdingfield, line 26: Line 67 refers to 4.7.3, and `exclude.yaml` refers to 4.6.0. Can we address these disparities?

Also, the line for the plugin (`id "com.github.spotbugs" version "5.0.14"`) has yet another version. I'm not sure if this is specific to the plugin though, which might be ok.

Finally, does it really make sense to add this as a _dependency_ for our project? Seems much more like a development tool that's nice-to-have.

README.md
```diff
@@ -53,6 +53,20 @@ On Linux, you may install the .deb file, then run `/opt/rcv/bin/RCTab` to launch
     
     `$ chmod 777 gradlew`
 
+#### Method 3 (Least Easy): Building on an airgapped Machine
+
+
+1. Download gradle from https://gradle.org/releases/ and place it in your path
+2. Download and extract the source code from [releases page](https://github.com/BrightSpots/rcv/releases)
+3. Download the appropriate cache files for your OS: cache.[OS].zip
```
HEdingfield, line 61: Can you link to where these are found?

README.md
```diff
@@ -53,6 +53,20 @@ On Linux, you may install the .deb file, then run `/opt/rcv/bin/RCTab` to launch
...
+4. Stop the gradle daemon with `gradle --stop`
+5. Delete the directory ~/.gradle/caches if it exists
+6. Extract the appropriate caches/[filename].zip to ~/.gradle/caches so that the "caches" directory is in ~/.gradle
+7. In the extracted directory, you may manually verify each dependency using checksums.csv in accordance with your own policies
```
HEdingfield, line 65: Where is checksums.csv found?

README.md
```diff
@@ -53,6 +53,20 @@ On Linux, you may install the .deb file, then run `/opt/rcv/bin/RCTab` to launch
...
+8. Run `gradle assemble --offline` and ensure you get no errors
+9. Run `gradle run --offline` to launch RCTab
+10. Run `gradle jpackage --offline` to generate a jpackage package
```
HEdingfield, line 68: Are all these meant to be run in sequence? It seems like maybe you could at least combine the last two saying "Run X to XX or Y to YY".

Also, will people know what "a jpackage package" is?

.github/workflows/release.yml
```diff
@@ -5,24 +5,24 @@ on:
...
-## To test this workflow without creating a release, uncomment the following and add a branch name (making sure "push"
-##  is at the same indent level as "release":
-#  push:
-#    branches:
-#      - 'branch-name'
+# To test this workflow without creating a release, uncomment the following and add a branch name (making sure "push"
+#  is at the same indent level as "release":
+  push:
+    branches:
+      - 'feature/issue-756_airgap'
```
HEdingfield, line 12: Reminder to revert.

.github/workflows/release.yml
```diff
@@ -130,11 +171,16 @@ jobs:
...
+            ${{ github.workspace }}/${{    steps.zipfn.outputs.FILEPATH    }}
+            ${{ github.workspace }}/${{    steps.zipfn.outputs.FILEPATH    }}.sha512
+            ${{ github.workspace }}/${{    steps.zipfn.outputs.FILEPATH    }}.golden.sha512
+            ${{ github.workspace }}/${{    steps.exefn.outputs.FILEPATH    }}
+            ${{ github.workspace }}/${{    steps.exefn.outputs.FILEPATH    }}.sha512
+            ${{ github.workspace }}/${{   steps.cachefn.outputs.FILEPATH   }}
+            ${{ github.workspace }}/${{   steps.cachefn.outputs.FILEPATH   }}.sha512
+            ${{ github.workspace }}/${{ steps.checksumsfn.outputs.FILEPATH }}
+            ${{ github.workspace }}/${{ steps.checksumsfn.outputs.FILEPATH }}.sha512
+          retention-days: 1
```
HEdingfield, line 183: Reminder to revert to 90.

build.gradle
artoonie: 1. Fixed the version discrepancy
2. The version 5.0.14 is the plugin version, which contains a different spotbugs version
3. I can't recall exactly why this was required, but it would fail if I marked it as `testImplementation` instead of `implementation`. iirc, since it's used in the build process, it has to be included in the jpackage. I can try to re-investigate what happens if I change it back and better document it if needed.

HEdingfield: Nice, thanks. What about this one?

> exclude.yaml refers to 4.6.0. Can we address these disparities?

Re: 3. maybe yezr@ or you could file an issue for that?

artoonie: Ah my bad, glossed over that! Fixed

config/spotbugs/exclude.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <FindBugsFilter
-    xmlns="https://github.com/spotbugs/filter/4.6.0"
+    xmlns="https://github.com/spotbugs/filter/4.8.3"
```
HEdingfield, line 3: There's another one in line 5, but approved other than that!

HEdingfield (approved):

PR #805, opened 2024-03-04
coalesce(computedId, suppliedId) in audit.log + fixes

### LINKED ISSUE #804: Identifier in audit.log is not unique
When troubleshooting tabulations you can look in the `audit.log` to see how ballots were tallied in each round. Currently, `CastVoteRecord.logRoundOutcome` is used to write to the `audit.log`. That method uses `CastVoteRecord.suppliedId` as the ballot identifier.

#### What Happened
However, sometimes that `suppliedId` is not unique - for example with Dominion. The `DominionCvrReader` uses `Record Id` as the `suppliedId`. But you need to concat both `Tabulator Id` and `Batch Id` in order to make it unique.

#### What I Would Expect
 `CastVoteRecord` also has a `.computedId` which can be used to build a unique identifier for each ballot. We should use this `computedId` in the `audit.log`. 

This ticket should fix Dominion. I'll add another ticket to look through the other Vendors and ensure that either 1) `suppliedId` is unique or 2) we are building a unique identifier in `computedId`

yezr:
resolves #804 by coalescing computedId then suppliedId when writing to audit log. Also fixes Dominion and Csv cvr readers to properly populate ComputedId

### DIFF
--- a/src/main/java/network/brightspots/rcv/CastVoteRecord.java
+++ b/src/main/java/network/brightspots/rcv/CastVoteRecord.java
@@ -66,14 +66,7 @@ class CastVoteRecord {
       String precinct,
       String precinctPortion,
       List<Pair<Integer, String>> rankings) {
-    this.contestId = contestId;
-    this.tabulatorId = tabulatorId;
-    this.batchId = batchId;
-    this.computedId = null;
-    this.suppliedId = suppliedId;
-    this.precinct = precinct;
-    this.precinctPortion = precinctPortion;
-    this.candidateRankings = new CandidateRankingsList(rankings);
+    this(contestId, tabulatorId, batchId, suppliedId, null, precinct, precinctPortion, rankings);
   }
 
   CastVoteRecord(
@@ -88,8 +81,8 @@ class CastVoteRecord {
     this.contestId = contestId;
     this.tabulatorId = tabulatorId;
     this.batchId = batchId;
-    this.computedId = computedId;
     this.suppliedId = suppliedId;
+    this.computedId = computedId;
     this.precinct = precinct;
     this.precinctPortion = precinctPortion;
     this.candidateRankings = new CandidateRankingsList(rankings);
@@ -97,11 +90,7 @@ class CastVoteRecord {
 
   CastVoteRecord(
       String computedId, String suppliedId, String precinct, List<Pair<Integer, String>> rankings) {
-    this.computedId = computedId;
-    this.suppliedId = suppliedId;
-    this.precinct = precinct;
-    this.precinctPortion = null;
-    this.candidateRankings = new CandidateRankingsList(rankings);
+    this(null, null, null, suppliedId, computedId, precinct, null, rankings);
   }
 
   String getContestId() {
--- a/src/main/java/network/brightspots/rcv/CastVoteRecord.java
+++ b/src/main/java/network/brightspots/rcv/CastVoteRecord.java
@@ -77,6 +77,25 @@ class CastVoteRecord {
   }
 
   CastVoteRecord(
+          String contestId,
+          String tabulatorId,
+          String batchId,
+          String suppliedId,
+          String computedId,
+          String precinct,
+          String precinctPortion,
+          List<Pair<Integer, String>> rankings) {
+    this.contestId = contestId;
+    this.tabulatorId = tabulatorId;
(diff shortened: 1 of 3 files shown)


src/main/java/network/brightspots/rcv/CastVoteRecord.java
```diff
@@ -76,6 +76,25 @@ class CastVoteRecord {
...
+          List<Pair<Integer, String>> rankings) {
+    this.contestId = contestId;
+    this.tabulatorId = tabulatorId;
+    this.batchId = batchId;
+    this.computedId = computedId;
+    this.suppliedId = suppliedId;
+    this.precinct = precinct;
+    this.precinctPortion = precinctPortion;
+    this.candidateRankings = new CandidateRankingsList(rankings);
+  }
```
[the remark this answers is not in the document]
tarheel: Even if you keep `computedId` optional, you can consolidate your logic a bit by having the constructors with fewer arguments call `this` with the full arg list instead of having all the redundant assignments across the different constructors.

src/main/java/network/brightspots/rcv/CsvCvrReader.java
```diff
@@ -86,7 +88,7 @@ void readCastVoteRecords(List<CastVoteRecord> castVoteRecords)
 
         // create the new CastVoteRecord
         CastVoteRecord newCvr =
-            new CastVoteRecord(source.getContestId(), "no supplied ID", "no precinct", rankings);
```
tarheel, line 89: was this just a bug?

src/main/java/network/brightspots/rcv/CastVoteRecord.java
```diff
@@ -115,10 +134,10 @@ void logRoundOutcome(
...
     StringBuilder logStringBuilder = new StringBuilder();
     logStringBuilder.append("[Round] ").append(round).append(" [CVR] ");
-    if (!isNullOrBlank(suppliedId)) {
-      logStringBuilder.append(suppliedId);
-    } else {
+    if (!isNullOrBlank(computedId)) {
       logStringBuilder.append(computedId);
+    } else {
+      logStringBuilder.append(suppliedId);
     }
```
tarheel, line 141: should we show both if they're both present? isn't it plausible that someone reviewing the audit log would want to search by the supplied ID?

tarheel: I wonder if non-unique `suppliedId` values should be a halting exception, or at least a warning. When they're not, presumably either we're not ingesting those IDs properly (e.g. the case you mentioned in which you need to concatenate multiple fields but we were only taking one of those fields) or there's a problem with the source data?

yezr: This example is from Dominion. In their CVR they really have three separate "ids": TabulatorId, BatchId, and RecordId. Though that data resides in three distinct places in the CVR file, I would bet that Dominion assumes that only the three together can uniquely identify a record.

Our code chose to put RecordId into CastVoteRecord.suppliedId, probably because it looks the most unique: it increments with every record until a new batch or tabulator Id show up (in the data I looked at, often thousands of records). Maybe we could have just left `suppliedId` `null`? Any of the three could make the case that they should be the `suppliedId`. 

I think of the two as 
`suppliedId` = The provider's attempt at a unique identifier, not reliably unique but _can_ be. Helpful sometimes even if it's not unique when navigating vendor files or in their UI and this value is sometimes more visible (despite the data necessary for a unique identifier also potentially being there).
`computedId` = Our attempt at a unique identifier, always reliably unique. Can be used during troubleshooting to identify a single record for troubleshooting.

In your example of us not ingesting properly or there is a problem with the source data it is the non-unique `computedId` that should cause a halting exception. It wouldn't be a surprise when `suppliedId` is not unique.

src/main/java/network/brightspots/rcv/CastVoteRecord.java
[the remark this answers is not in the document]
yezr: I refactored this with the newest commit [b281aa2](https://github.com/BrightSpots/rcv/pull/805/commits/b281aa2eeba86d3031dda8454fcdb13ce139cacb)

artoonie (approved):
This seems good to me -- @tarheel feel free to take another look

PR #810, opened 2024-03-15
include a "#" column on candidate and CVR list

### LINKED ISSUE #674: List total number of CVR files added on CVR Files tab
so admins have another way to check that they've uploaded their full set of CVRs

artoonie:
closes #674 

Done for both the candidate and CVR list
Adapted from this code: https://stackoverflow.com/a/41282740/1057105 (do we want to include a reference to s/o in the code?)

### DIFF
--- a/src/main/java/network/brightspots/rcv/GuiConfigController.java
+++ b/src/main/java/network/brightspots/rcv/GuiConfigController.java
@@ -1789,6 +1789,9 @@ public class GuiConfigController implements Initializable {
     }
   }
 
+  /**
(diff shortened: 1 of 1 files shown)


yezr:
Yes, let's also add to the candidate list.

tarheel (approved):
I'm a bit confused by `startNumber`. It looks like you always pass 1 for this arg and you never change the value. But there must be something in the interface that modifies this value...?

> Adapted from this code: https://stackoverflow.com/a/41282740/1057105 (do we want to include a reference to s/o in the code?)

Might as well.

PR #811, opened 2024-03-15
fix (and add tests for) tabulate by precinct

### LINKED ISSUE #802: [2.0 RC] When using tabulate by precinct, summary.json thresholds are null
### What Happened
Tabulated `2013_minneapolis_mayor` test data and checked a couple of the `precinct_summary.json`s. In each of them, the `round.threshold` under a `round` was null and `summary.threshold` at the bottom was null.
[2024-02-23_16-48-10_MINNEAPOLIS_W-1_P-02_precinct_summary.json](https://github.com/BrightSpots/rcv/files/14390428/2024-02-23_16-48-10_MINNEAPOLIS_W-1_P-02_precinct_summary.json)

Looking at the full contest `summary.json` it has `threshold` filled out correctly. 
[2024-02-23_16-48-10_summary.json](https://github.com/BrightSpots/rcv/files/14390432/2024-02-23_16-48-10_summary.json)

artoonie:
Closes #802 -- the precinct files didn't include a threshold.

I also noticed the precinct files were never tested. However, we don't want to test every single precinct file as there can be hundreds of precincts in a single contest.

I've opted to allow you to include an arbitrary set of precinct files in the test directory, and it will only test those. It's a little dangerous, I guess, because it'll be easy for a bug to slip in and have it never test any precinct file. I have two ideas for resolving:
1. `private static void compareFiles` returns a boolean indicating whether the check was skipped, and if tabulate by precinct is checked, we ensure at least N files have not been skipped
2. similarly, that function could `throw IOException` if the expected file isn't found, instead of accepting a boolean argument, and we could perform the same logic

(2) feels cleaner but I'm hesitant to use exceptions for any expected behavior in Java, so I lean towards (1). Thoughts?


yezr:
If I'm reading the code correctly this is adding the precinct files to be compared just like `.summary` files currently are if tabulate by precinct is enabled? 

> I've opted to allow you to include an arbitrary set of precinct files in the test directory
How does this work? Do you just check in to git only the precinct files that you want checked?

artoonie:
Not every time it's enabled, since it adds a lot of files for little value. It's optional, and is only tested when the file is included in the test folder.

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -441,7 +442,21 @@ private void setWinningThreshold(RoundTally currentRoundTally, BigDecimal minimu
       winningThreshold = minimumVoteThreshold;
     }
 
+    if (currentRoundTally != roundTallies.get(currentRoundTally.getRoundNumber())) {
+      throw new RuntimeException("RoundTally object is not the same as the one in the map.");
+    }
```
tarheel, line 447: Why do we pass `currentRoundTally` to this method at all?

tarheel (commented):
I agree that we should enforce some kind of check here; I wouldn't want the test to pass just because there weren't any expected precinct summary files included. I'm indifferent between your two proposed implementations, but, either way, it seems reasonable to set some minimum number that must be present. Honestly, maybe just set it to 1? Ideally we'd be able to choose a custom value for this number for a given test, but it's actually not obvious to me that there's a simple way to do that.

src/main/java/network/brightspots/rcv/Tabulator.java
artoonie: Because we now support a per-round threshold, so currentRoundTally is used to (1) get the total number of ballots in this round, and (2) get the round number. We could replace it with two arguments (roundNum and currentRoundNumBallots) instead.

artoonie:
> I agree that we should enforce some kind of check here; I wouldn't want the test to pass just because there weren't any expected precinct summary files included. I'm indifferent between your two proposed implementations, but, either way, it seems reasonable to set some minimum number that must be present. Honestly, maybe just set it to 1? Ideally we'd be able to choose a custom value for this number for a given test, but it's actually not obvious to me that there's a simple way to do that.

I've added a check per test that specifies the number of expected precinct files to check, with it defaulting to zero

src/main/java/network/brightspots/rcv/Tabulator.java
tarheel: Sorry, to clarify my question: since `roundTallies` and `roundNumber` are both instance variables, why do we need to pass args to this method to get that information? Actually, same thing for `minimumVoteThreshold` (which comes from `config`) and the `roundNumber` arg in the new `setWinningThreshold` method.

src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -193,11 +194,20 @@ private static String getTestFilePath(String stem, String suffix) {
...
+  }
+
+  private static void runTabulationTest(String testStem, String expectedException) {
+    runTabulationTest(testStem, expectedException, 0);
   }
 
   // helper function to support running various tabulation tests
-  private static void runTabulationTest(String stem, String expectedException) {
+  private static void runTabulationTest(String stem, String expectedException,
+                                        int expectedNumPrecinctFileToCheck) {
```
tarheel, line 210: `expectedNumPrecinctFilesToCheck`? (add 's'?)

tarheel (approved):

src/main/java/network/brightspots/rcv/Tabulator.java
artoonie: Ah, got it -- it's not a strong preference, just a stylistic choice. The function as it stands is not required to be run on the current round.

I don't recall why I added this exception, but I assume I at some point had a bug that would have been caught by this check. As for the method signature, that is unchanged from the existing code.

I don't mind removing them, but do find it a little more readable as-is.

PR #816, opened 2024-03-29
Tabulate by Batch, and readiness for Tabulate By X

### LINKED ISSUE #807: Tabulating by "sub-level" Enhancement 1: Tabulate by Batch
## Background
When tabulating a contest, RCTab currently reports the results at two levels

- By default, the `summary` files are reported at the full contest level: all configured CVRs are included
- Optionally, users can select "Tabulate By Precinct" in the Output tab. When selected, an additional `summary` file is created for each precinct. These files are tabulated at the precinct level. They are similar to the full contest `summary` files except for
  - They include a `Precinct` row in the contest info with the name of the precinct. 
  - Ballots from outside of that precinct are not included in the round by round tallies. So in each precinct `summary` file the Total Number of Ballots reported and vote totals, transfer numbers, % of vote, and inactive ballot counts at each round can be different.
image

These precinct-level `summary` files follow the same elect/eliminate steps as the full contest: candidates are elected/eliminated in the same round as the full contest **no matter the vote total at the precinct-level**. 

## Proposed enhancements to sub-jurisdiction reporting
We've got feedback from election administrators that it would be beneficial to make more configurable the different levels at which results are reported. Some examples are

- **Reporting at the batch level for hand counts** this allows hand counts to easily select a physically separate stack of ballots to do a hand count on. Results of the hand count could easily be compared to a batch-level `summary` file
- **Reporting at the Congressional District level for Presidential Primaries** Some states award Electoral College electors based on winners at the Congressional District-level. [[0]](#footnotes)
- **If we're getting real fancy**, we could make it configurable in the Output tab, in conjunction with the CVR tab, any column or property of the CVR that we could group by for sub-level reporting [[1]](#footnotes)

## Ok yeah, but what about this ticket
The scope of this ticket is adding the first step in enhanced sub-level reporting: tabulate by Batch Id. All the context above ☝️is for whoever takes this on to consider while developing it that there is a high potential that this sub-level reporting could be extended in the future. Batch level reporting would
- Similar to Tabulate by Precinct, this would be a toggleable box in the Output tab
- When toggled, it would use the Batch Id supplied to group ballots. One additional `summary` file per Batch.
- Similar to Tabulate by Precinct, this would use the elect/eliminate decisions by round from the contest-level tabulation.
---
#### Footnotes
[0] This requires also changing the way sub-level winners work. As noted above, Tabulate by Precinct uses full contest elect/eliminate. Implementing Congressional District-level winners would need to elect/eliminate at _only_ the CD level. That's a whole 'nother ticket. Just wanted to provide example of different sub-level reporting requests for background. 
[1] Like a researcher or troubleshooter could supplement the original CVR with other data and group by whatever that is. This has not been requested yet, but wanted to flag it for this implementation.

artoonie:
Request for comment on language.
As a noun, I call it a "a tabulate-by field"
As a verb, I call it "Tabulate By Field"

I can imagine preferences for other names, by "Tabulate By Pivot" or "Tabulate By Metadata Field" or something else, but "Tabulate By Field" seems sufficiently clear to me. Thoughts? Any other words that may reduce ambiguity?

Closes #807


artoonie:
@yezr this has been lightly tested but I could use some help gathering test data, test configurations, and checking edge cases.

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -63,8 +66,8 @@ class ResultsWriter {
 
   // number of rounds needed to elect winner(s)
   private int numRounds;
-  // all precinct Ids which may appear in the output cvrs
-  private Set<String> precinctIds;
+  // all Field Ids which may appear in the output cvrs
```
tarheel, line 69: s/which/that

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -565,7 +573,8 @@ String writeRctabCvrCsv(
               outputFile.getAbsolutePath());
       CSVPrinter csvPrinter;
       BufferedWriter writer = Files.newBufferedWriter(outputFile.toPath());
-      csvPrinter = new CSVPrinter(writer, CSVFormat.DEFAULT);
+      CSVFormat format = CSVFormat.DEFAULT.builder().setNullString("").build();
```
tarheel, line 576: convenient

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -59,20 +60,20 @@ class Tabulator {
   // roundTallies is a map from round number to a map from candidate ID to vote total for the round
   // e.g. roundTallies[1] contains a map of all candidate ID -> votes for each candidate in round 1
   // this structure is computed over the course of tabulation
-  private final Map<Integer, RoundTally> roundTallies = new HashMap<>();
-  // precinctRoundTallies is a map from precinct to roundTallies for that precinct
-  private final Map<String, Map<Integer, RoundTally>> precinctRoundTallies = new HashMap<>();
+  private final RoundTallies roundTallies = new RoundTallies();
+  // roundTalliesByField is a map from a tabulate-by field to roundTallies for that field
+  private final BreakdownByField<RoundTallies> roundTalliesByField = new BreakdownByField<>();
```
tarheel, line 65: Am I misreading the code, or does this mean that if you're using more than one breakdown (with the current code, that would be both batch and precinct), there will be a collision if there's any overlap in the IDs? E.g. if there's a batch 1 and also a precinct 1, they'll be writing to and reading from the same tallies?

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -82,20 +83,27 @@ class Tabulator {
...
     this.config = config;
 
+    fieldIds.initialize(TabulateByField.BATCH);
+    fieldIds.initialize(TabulateByField.PRECINCT);
+
     for (CastVoteRecord cvr : castVoteRecords) {
-      String precinctId = cvr.getPrecinct();
-      if (precinctId != null) {
-        precinctIds.add(precinctId);
+      for (TabulateByField field : config.enabledFields()) {
```
tarheel, line 90: good example of the ambiguity here: without context, I would never guess that `config.enabledFields()` is giving me a list of enabled breakdown fields

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -789,8 +798,8 @@ void generateSummaryFiles(String timestamp) throws IOException {
     }
   }
 
-  Set<String> getPrecinctIds() throws IOException {
-    return precinctIds;
+  FieldIdSet getFieldIds() {
```
tarheel, line 801: Should `fieldIds`/`getFieldIds` clarify that it's referring to the _enabled_ ones? And why do we need the `fieldIds` instance variable instead of just using `config.enabledFields()`? Aren't we throwing an exception if there's a discrepancy between the config and the CVRs in this regard anyway?

src/main/java/network/brightspots/rcv/Tabulator.java
tarheel: Ah, no, now I see that `roundTalliesByField` is a map from field to fieldId to round to vote count. Maybe `roundTalliesByFields` (plural) would be more clear?

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -1334,6 +1338,72 @@ enum CandidateStatus {
...
+      return roundTallies.get(roundNumber);
+    }
+  }
+
+  /**
+   * A wrapper around Map&lt;TabulateByField, Map&lt;String, T&gt;&gt;
+   * Breaks down the templated type T by a specific field.
+   * The field type is the outer map (TabulateByField), and the field ID is the inner map (String).
+   */
+  static class BreakdownByField<T> {
```
tarheel, line 1362: OK, so you made this generic so that you could reuse it for both tallies and tally transfers.

tarheel (commented):
This seems great. My feedback is mainly around the language. The verb phrase "tabulate by field" is OK, but could be clearer. There and elsewhere, my concern is mostly about how generic the word "field" is.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1291,5 +1332,21 @@ public String getInternalLabel() {
     }
   }
 
+  enum TabulateByField {
```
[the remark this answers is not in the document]
artoonie: I like `slice`, and think it will make sense to s/field/slice throughout this PR

artoonie:
@tarheel all comments are addressed!

tarheel (approved):

PR #818, opened 2024-03-31
allow missing CVR headers

### LINKED ISSUE #813: CSV Candidate Auto-Load fails if no text in upper left corner
The following CSV will fail auto-load candidates:

```
,A,B,C,D
Person 1,1,2,3,4
```

The following will succeed:
```
Name,A,B,C,D
Person 1,1,2,3,4
```

artoonie:
closes #813 

This type of CSV is allowed and I don't see a reason to not support it. Added a test to account for it.

Because `getHeaderNames` skips all null/empty columns, I had to move away from using the built-in header parser.

I also put the `Generic CSV Writer` behind a `encounteredSourceProblem` flag since it will lead to more exceptions on failure.


yezr:
Do we need to then add some basic validation checks to make sure that each candidate column has characters? I created a .csv [generic_csv_test_cvr_2.csv](https://github.com/BrightSpots/rcv/files/14960483/generic_csv_test_cvr_2.csv) that has a blank candidate name and I get the following error. 

image

artoonie:
Good catch -- yes, will need to fix that.

artoonie:
Fixed -- it took a bit more work than expected, the built-in header parser really doesn't like null header names, even though it's fairly common practice in CVRs/Spreadsheets to have empty headers...go ahead and give this a whirl?

yezr:
image
success!

src/test/resources/network/brightspots/rcv/test_data/generic_csv_test/generic_csv_test_cvr.csv
```diff
@@ -1,4 +1,4 @@
-Response ID,Lettuce,Broccoli,Cucumber,Cauliflower,Write-In
+,Lettuce,Broccoli,Cucumber,Cauliflower,Write-In
```
HEdingfield, line 1: Can we please keep this as-is, and make a new file and test for this case instead?

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -344,16 +344,18 @@ private List<CastVoteRecord> parseCastVoteRecords(ContestConfig config) {
...
+                new ResultsWriter().setContestConfig(config).setTimestampString(timestampString);
+        this.convertedFilePath =
+                writer.writeRctabCvrCsv(
+                        castVoteRecords,
+                        perSourceDataForCsv,
+                        config.getOutputDirectory());
+      } catch (IOException exception) {
+        // error already logged in ResultsWriter
+      }
     }
```
HEdingfield, line 359: Nit: `else` this with line 361.

src/main/java/network/brightspots/rcv/CsvCvrReader.java
HEdingfield, line 1: @tarheel would you mind taking a look at this too?

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -392,8 +392,18 @@ private LoadedCvrData parseCastVoteRecords(ContestConfig config, Progress progre
       progress.markFileRead();
     }
 
-    // Output the RCTab-CSV CVR
-    if (!encounteredSourceProblem) {
+    if (encounteredSourceProblem) {
```
HEdingfield, line 399: Looks like you've got duplicate code between here and line 420.

artoonie: Bad merge! Fixed, thanks for catching.

src/main/java/network/brightspots/rcv/CsvCvrReader.java
HEdingfield: I feel confident enough to approve this without @tarheel's input now. 👍

HEdingfield (approved):

PR #819, opened 2024-03-31
going over the max ranking throws a halting exception

### LINKED ISSUE #808: CVRs with rankings > configured Maximum Number of Candidates That Can Be Ranked should throw a halting exception
CVR: [CSV CVRs - more ranks than configured.csv](https://github.com/BrightSpots/rcv/files/14501661/CSV.CVRs.-.more.ranks.than.configured.csv)
Config: [undefined_more-rankings-than-configured.json](https://github.com/BrightSpots/rcv/files/14501662/undefined_more-rankings-than-configured.json)
The CVR above has 4 rankings, but is configured for a maximum of 3. 
image

In the Winning Rules tab there is a setting for `Maximum Number of Candidates That Can Be Ranked`. If any CVR has a ranking that exceeds this limit it should fail validation as a halting exception. 

This ticket is following up with the conversation [here](https://github.com/BrightSpots/rcv/pull/786#issuecomment-1908816023) in #786.

artoonie:
except in ES&S XLSX files, as discussed in the issue.

Closes #808


src/main/java/network/brightspots/rcv/CandidateRankingsList.java
```diff
@@ -81,6 +81,9 @@ CandidatesAtRanking get(int i) {
   }
 
   int maxRankingNumber() {
+    if (numRankings == 0) {
+      throw new IllegalArgumentException("Max ranking may only be called on non-empty rankings");
```
HEdingfield, line 85: Is adding this actually necessary? Just concerned that it might cause an exception when one isn't warranted (e.g. in cases when `maxRanking` isn't defined and therefore defaults to 0).

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -652,11 +652,6 @@ private void printRankings(
       throws IOException {
     // for each rank determine what candidate id, overvote, or undervote occurred and print it
     for (int rank = 1; rank <= maxRanks; rank++) {
-      // If the configuration did not allow this ranking, we want to exclude it
-      // from the RCTab CVR -- even if it was present in the source CVRs.
-      if (!reader.isRankingAllowed(rank, source.getContestId())) {
-        break;
-      }
```
HEdingfield, line 659: Is the idea that this is no longer needed because it'll hit a halting exception before it has a chance to get to this state?

src/test/resources/network/brightspots/rcv/test_data/convert_to_cdf_from_dominion/convert_to_cdf_from_dominion_config.json
```diff
@@ -68,7 +68,7 @@
     "decimalPlacesForVoteArithmetic": "4",
     "minimumVoteThreshold": "0",
     "maxSkippedRanksAllowed": "1",
-    "maxRankingsAllowed": "8",
+    "maxRankingsAllowed": "5",
```
HEdingfield, line 71: Curious why this was necessary?

HEdingfield (commented):
LGTM, but hoping @tarheel can give this a final approval.

src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -176,8 +176,10 @@ public void runAdditionalValidations(List<CastVoteRecord> castVoteRecords)
   }
 
   @Override
-  public boolean isRankingAllowed(int rank, String contestId) {
-    return rank > 0 && rank <= contests.get(contestId).getMaxRanks();
+  protected boolean isRankingAllowed(int rank, String contestId) {
+    return rank > 0
+      && rank <= contests.get(contestId).getMaxRanks()
+      && config.isRankingAllowed(rank);
```
tarheel, line 182: I'm having trouble understanding/remembering why the last condition isn't a sufficient check on its own here. At a minimum, I think the `rank > 0` part is redundant?

src/main/java/network/brightspots/rcv/ResultsWriter.java
tarheel: Yeah, I think this is now supposed to get caught when we first read in the CVRs.

src/main/java/network/brightspots/rcv/StreamingCvrReader.java
```diff
@@ -213,11 +213,16 @@ private void cvrCell(int col, String cellData) {
       currentPrecinct = cellData;
     } else if (idColumnIndex != null && col == idColumnIndex) {
       currentSuppliedCvrId = cellData;
-    }
+    } else if (col >= firstVoteColumnIndex
+        && (config.isMaxRankingsSetToMaximum()
+            || col < firstVoteColumnIndex + config.getMaxRankingsAllowedWhenNotSetToMaximum())) {
```
tarheel, line 218: We don't throw here if this check fails because we rely on `runAdditionalValidations` to catch it later?

src/main/java/network/brightspots/rcv/StreamingCvrReader.java
```diff
@@ -227,15 +232,15 @@ private void cvrCell(int col, String cellData) {
...
         candidates = cellString.split(Pattern.quote(overvoteDelimiter));
       } else {
-        candidates = new String[] {cellString};
+        candidates = new String[]{cellString};
       }
 
       for (String candidate : candidates) {
         candidate = candidate.trim();
-        if (candidates.length > 1 && (candidate.equals("") || candidate.equals(skippedRankLabel))) {
+        if (candidates.length > 1 && (candidate.isBlank() || candidate.equals(skippedRankLabel))) {
```
tarheel, line 240: does this change have any practical consequence?

src/main/java/network/brightspots/rcv/DominionCvrReader.java
artoonie: Yes, `rank > 0` is redundant, though perhaps not obviously so? This checks: "Is this valid for this contest" (0 < rank < maxRanks), and separately, "is this valid in the config".

I don't mind removing `rank > 0`, but I can see an argument for leaving it in: it allows the config to define its own validity rules independent of what's valid for this dominion contest.

src/main/java/network/brightspots/rcv/StreamingCvrReader.java
artoonie: This is based on the following discussion: https://github.com/BrightSpots/rcv/issues/808#issuecomment-2028944594

Essentially, it's valid to have data after the last max ranking.

src/main/java/network/brightspots/rcv/StreamingCvrReader.java
artoonie: No, because of the `trim` above, but this is different in that a series of whitespace `isBlank` but is not `equals("")`

src/test/resources/network/brightspots/rcv/test_data/convert_to_cdf_from_dominion/convert_to_cdf_from_dominion_config.json
artoonie: Just minimizes the number of changes to the expected CSV: without it, every line would have three additional commas for the three additional rankings.

tarheel (approved):

PR #823, opened 2024-04-01
support having more winners than candidates

### LINKED ISSUE #797: WIP Configuring Number of Winners > Number of Declared Candidates results in confusing error message
**RCTab Version**: 1.4.0 and 1.3.1

## What Happened
If you configure Number of Winners > the number of candidates you have configured in the Candidates tab you get the following error 

`2024-02-15 08:55:30 EST SEVERE: numberOfWinners must be an integer from 0 to [numberOfConfiguredCandidates]!`
from line 700 in ContestConfig.java in 1.4.0

So for example I configured one candidate and set the Number of Winners to 3 and got
image

I imagine this check failed validation because when you have less configured candidates than the configured number of winners you don't need to run any tabulation rounds: each configured candidate is automatically elected. If I only have two candidates running for 3 City Council seats then I don't need to count any votes to know that those two will be elected.
## What I Expect To Happen
A more detailed message like 
> "The number of declared Candidates (`numDeclaredCandidates`) must be greater than or equal to the Number Of Winners (`numberOfWinners`) for Multi-Winner contests. 

OR

Another option we are discussing is to let this pass config validation. When it gets to the tabulation I expect it will get to the logic that says "Do we have less continuing candidates than the configured number of winners? If so, elect them"

I lean towards RCTab producing output with winners in this example, because that can be certified as an election result whereas this halting exception cannot.

artoonie:
Closes #797 

Adds test. Doesn't block tabulation or pop up any sort of prompt to proceed, but it does warn in the audit log.


HEdingfield (approved):
LGTM!

PR #832, opened 2024-05-08
make the log output faster, and add colors too

### LINKED ISSUE #830: UI hangs during tabulation
While reviewing #812 I noticed that the progress bar hangs during tabulation. Armin's [comment](https://github.com/BrightSpots/rcv/pull/812#issuecomment-2101250292) says this was always there, not new to #812. It is a little more noticeable, since now it is obvious that the progress bar hangs. 

It starts out moving, then at some point during tabulation hangs. Here's [a video](https://drive.google.com/file/d/1yLZKU1SLhkGhr4T4q0GecG8ek_iz8YGU/view?usp=drive_link).

### LINKED ISSUE #831: Tabulate takes ~5 seconds longer on consecutive runs
Testing #812 I noticed that consecutive runs of tabulate take ~5 seconds longer each run. 

 Here's some splits I did with the Minneapolis Mayor test data. It got about 5 seconds longer each time I ran it.
image

artoonie:
Closes #831 
Closes #830 

The issue was that appending to the textarea used string concatenation, which required reading the entire log string and recreating it each time a new log message was added. That made each subsequent GUI run slower, and also explains why we didn't see this in the unit tests: it's only a GUI issue. It also caused the GUI to hang for a short bit of time each time a log message was added.

I replaced the TextArea with a ListView. This makes it trivial to do something that I've wanted for a while: colors for log levels. I can change the foreground or background color. I would appreciate feedback on how this looks:

Screenshot:
image

### DIFF
--- a/src/main/java/network/brightspots/rcv/Logger.java
+++ b/src/main/java/network/brightspots/rcv/Logger.java
@@ -21,7 +21,7 @@
  *  Captures all INFO level logging for the execution of a session.
  *  "session" could span multiple tabulations in GUI mode.
  *
- * GUI handler (INFO) -> textArea
+ * GUI handler (INFO) -> listView
  *  Displays INFO level logging in GUI for user feedback in GUI mode.
  *
  * Default handler -> console
@@ -44,7 +44,12 @@ import java.util.logging.Handler;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
 import javafx.application.Platform;
-import javafx.scene.control.TextArea;
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableList;
+import javafx.scene.control.Label;
+import javafx.scene.control.ListView;
+import javafx.scene.layout.Background;
+import javafx.scene.paint.Color;
 
 class Logger {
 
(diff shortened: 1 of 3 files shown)


yezr:
I love the colors. Side scroll totally works. My only nit is the line spacing. Is it possible to move the lines closer? 
image

artoonie:
https://github.com/BrightSpots/rcv/assets/537316/fa2aa0da-ee3c-49b4-8c8f-ca9c1597ff03

Sounds good. I spruced up the style a little more:
1. Reduced line spacing
2. Decided to go back to word-wrap, but this is an easy change either way. I realize there's important info at the end of the line, and may be better to not force a scroll
3. Made the highlight reach across the whole line
4. Added the ability to "select" each row and have that row maintain its log severity color (row = a single log string, which may span multiple lines)
5. Added the ability to copy lines via right click

tarheel:
The code changes seem pretty uncontroversial. Will let @yezr accept.

artoonie:
From https://docs.oracle.com/javase/8/javafx/api/javafx/application/Platform.html#runLater-java.lang.Runnable-

```NOTE: applications should avoid flooding JavaFX with too many pending Runnables. Otherwise, the application may become unresponsive. Applications are encouraged to batch up multiple operations into fewer runLater calls. Additionally, long-running operations should be done on a background thread where possible, freeing up the JavaFX Application Thread for GUI operations.```

That was the cause of the GUI lag that has existed outside of this PR, but wasn't as noticeable until the new GUI's loading bar.

@tarheel there is a small final commit on this PR that fixes it -- it's important to get this right to avoid race conditions where the log output never updates. It's the first and only use of a `synchronized` block in the codebase. Can you take a quick look at commit [21dc086](https://github.com/BrightSpots/rcv/pull/832/commits/21dc0863ce0fcbb996ee4ecbfca63cec88e5e59a)?

File: src/main/java/network/brightspots/rcv/Logger.java [hunk 1]
```diff
@@ -36,8 +36,10 @@
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
+import java.util.concurrent.BlockingDeque;
```
tarheel, line 42: unused? otherwise this seems fine to me!

yezr:
nit: can we make the highlight apply only to the text instead of the entire line? In that example video when all of the visible lines have a background there is no contrast with the regular background so it maybe looks like the background is usually that color?

artoonie:
Happy to make that change! Just want to flag something first:

That's how it [used to look here](https://github.com/BrightSpots/rcv/pull/832#issuecomment-2102882147), but in almost all cases with an error, the error is only one `severe` surrounded by many `info`s, unlike the demo video I posted. I changed it from only highlighting the text to highlighting the entire line because it makes it easier to see at-a-glance when scrolling through the logs which lines had errors. It's especially helpful if you have sidescroll, since you might be scrolled over to the right and scrolling up, and if the error line is short you won't notice the error line at all, whereas if it's the full line you'll see it regardless of horizontal scroll.

In light of this reasoning, can you let me know if you still prefer it with only the text background highlighted?

Happy to share side-by-side screenshots, it's only one line of code difference between the two options.

yezr:
Thanks for the context Armin. Let's keep the full line highlighting.

yezr (approved):
looks good to me!

PR #834, opened 2024-05-31
Granular progress bar

### LINKED ISSUE #825: Granular progress bar in tabulate modal
https://github.com/BrightSpots/rcv/issues/679 describes a new modal that provides one place for pre-tabulation checks. Part of that modal is a progress bar to indicate to users that work is being done. Initially, that progress bar will be simply bouncing back and forth to show progress while work is being done. It won't necessarily help gauge how much time is remaining.

This ticket builds on that. We know for each step how many sub-tasks are required to finish. We should use that to provide more granular progress on the progress bar.

### Reading CVRs
First step is Reading CVRs (X of Y) where Y=total number of configured CVRs and X=number of those CVRs that have been parsed through. Progress bar shows X/Y.

### Tabulating Round by Round Tallies
Second step is Tabulating Round by Round Tallies. Progress bar shows (Count of Candidates Elected/Eliminated)/(Total Count of Candidates)

For now, we don't call out explicitly if we are tabulating by precinct or batch. It is lumped under the Round by Round Tallies section. Do we need to separate this and call it out explicitly? Let's get started into this and revisit.

artoonie:
Closes #825 

Builds a granular progress bar in the simplest way I could think of:
* 50% of time is allocated to reading CVRs
* 50% of time is allocated to tabulation

Tabulation is measured in number of eliminated candidates. This makes Round 1 take longer than other rounds, since more happens in the first round before eliminations begin occurring, but it's still a pretty close approximation.

The benefit of this approach is that we can initialize the full state before reading any CVRs -- we know the maximum number of eliminations.

Of course, we can't know if we can stop early in the progress bar. For example, if a candidate gets over 50% of votes in the first round, there will be zero eliminations. This seems fine to me -- the progress bar will always be a worst-case, and finishing early is okay.

The progress bar also handles the odd case of Multi-Pass IRV, where the same files are read multiple times and the same candidate is eliminated multiple times. However, every test case I have of that is quite small, so while it seems to work, I haven't dug deeply to guarantee it works as intended. Maybe @nurse-the-code could help create a very large multi-pass IRV test case and make sure this works as intended?

Demo:

https://github.com/BrightSpots/rcv/assets/537316/55427302-3b0a-494a-bfd8-cc1f3937a02d

### DIFF
--- a/src/main/java/network/brightspots/rcv/GuiConfigController.java
+++ b/src/main/java/network/brightspots/rcv/GuiConfigController.java
@@ -1847,7 +1847,8 @@ public class GuiConfigController implements Initializable {
             @Override
             protected Boolean call() {
               TabulatorSession session = new TabulatorSession(configPath);
-              List<String> errors = session.tabulate(operatorName, expectedCvrStatistics);
+              List<String> errors = session.tabulate(
+                      operatorName, expectedCvrStatistics, this::updateProgress);
               if (errors.isEmpty()) {
                 succeeded();
               } else {
@@ -1876,7 +1877,7 @@ public class GuiConfigController implements Initializable {
             @Override
             protected Boolean call() {
               TabulatorSession session = new TabulatorSession(configPath);
-              return session.convertToCdf();
+              return session.convertToCdf(this::updateProgress);
             }
           };
       setUpTaskCompletionTriggers(task,
@@ -1901,7 +1902,7 @@ public class GuiConfigController implements Initializable {
             TabulatorSession session = new TabulatorSession(configPath);
             LoadedCvrData cvrStatics = null;
             try {
-              cvrStatics = session.parseAndCountCastVoteRecords();
+              cvrStatics = session.parseAndCountCastVoteRecords(this::updateProgress);
               succeeded();
             } catch (TabulatorSession.CastVoteRecordGenericParseException e) {
               Logger.severe("Failed to read CVRs: %s", e.getMessage());
--- a/src/main/java/network/brightspots/rcv/Main.java
+++ b/src/main/java/network/brightspots/rcv/Main.java
@@ -80,10 +80,10 @@ public class Main extends GuiApplication {
 
       TabulatorSession session = new TabulatorSession(path);
       if (convertToCdf) {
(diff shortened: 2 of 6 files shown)


src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -1876,7 +1877,7 @@ protected Task<Boolean> createTask() {
             @Override
             protected Boolean call() {
               TabulatorSession session = new TabulatorSession(configPath);
-              return session.convertToCdf();
+              return session.convertToCdf(this::updateProgress);
```
HEdingfield, line 1880: Where is this actually used? I don't see any place in the GUI that shows a progress bar for converting to CDF.

HEdingfield (approved):
Brilliant job figuring this out!

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: It's not used, happy to remove though I don't see any downsides leaving it in either

HEdingfield: No problem, fine to leave as-is. Merge away!

yezr:
from the original issue

>  is post-tabulation work like tabulate by precinct and tabulate by batch included

As we've discussed, I think both of these take so little time since it is really just dumping in-memory counts onto disc. But just to confirm: is the time it takes to do these included under the progress bar as it's written in this PR? Even if it's not a part of the X/Y calculations, can we confirm that work happens while a progress bar is visible.

artoonie:
Confirmed: the "tabulate by" calculations happen on a round-by-round basis. It does take slightly longer at the end to write out the tabulate-by files, but that's hardly noticeable.

Here's an example with 4400 ballots:

## With "Tabulate by Precinct"
https://github.com/BrightSpots/rcv/assets/537316/d264dc39-9f97-4134-aa2b-9692af046284

## Without "Tabulate by Precinct"
https://github.com/BrightSpots/rcv/assets/537316/df03645e-7788-4d33-b250-3a8df9911043

PR #835, opened 2024-05-31
issue 826: show per-source ballot counts

### LINKED ISSUE #826: Include per-CVR ballot count numbers in pre-tabulation modal
#679 adds a pre-tabulation modal that includes a count of total ballots across all configured CVRs. #675 adds logging of ballot counts on a per-CVR basis. This ticket is to add the per-CVR ballot counts to the pre-tabulation modal. 

This extra information on the pre-tabulation modal allows election admins to double-check their configuration. For example, if different types of ballots are in different CVRs (like UOCAVA vs. early voting vs. in person voting OR each sub-jurisdiction has their own CVR file) those counts can be reviewed for inconsistencies before running a tabulation.

artoonie:
closes #826

This UI feels quite cluttered now. The table only appears after the first button is hit, so on pop-up it's cleaner.
1. Is there any risk of the same file name in different folders? Should we show the full path? If we do, it looks even harder to read.

image

This auto-sizes the filename column to try to make sure the full string is visible

### DIFF
--- a/src/main/java/network/brightspots/rcv/GuiTabulateController.java
+++ b/src/main/java/network/brightspots/rcv/GuiTabulateController.java
@@ -218,26 +218,27 @@ public class GuiTabulateController {
   }
 
   private void watchParseCvrServiceProgress(Service<LoadedCvrData> service) {
-    EventHandler<WorkerStateEvent> onSucceededEvent = workerStateEvent -> {
-      enableButtonsUpTo(tabulateButton);
-      LoadedCvrData data = service.getValue();
-      tabulateButton.setText("Tabulate " + String.format("%,d", data.numCvrs()) + " ballots");
+    EventHandler<WorkerStateEvent> onSucceededEvent =
+        workerStateEvent -> {
+          enableButtonsUpTo(tabulateButton);
+          LoadedCvrData data = service.getValue();
+          tabulateButton.setText("Tabulate " + String.format("%,d", data.numCvrs()) + " ballots");
 
-      // Populate the per-source CVR count table, and calculate the width of the filename column
-      perSourceCvrCountTable.getItems().clear();
-      int maxFilenameLength = 0;
-      for (Pair<String, Integer> perSourceCount : data.getPerSourceCvrCounts()) {
-        String countString = String.format("%,d", perSourceCount.getValue());
-        String fileString = new File(perSourceCount.getKey()).getName();
-        perSourceCvrCountTable.getItems().add(new Pair<>(fileString, countString));
-        maxFilenameLength = Math.max(maxFilenameLength, fileString.length());
-      }
-      perSourceCvrColumnFilepath.setPrefWidth(getEstWidthForStringInPixels(maxFilenameLength));
-      perSourceCvrCountTable.setVisible(true);
+          // Populate the per-source CVR count table, and calculate the width of the filename column
+          perSourceCvrCountTable.getItems().clear();
+          int maxFilenameLength = 0;
+          for (Pair<String, Integer> perSourceCount : data.getPerSourceCvrCounts()) {
+            String countString = String.format("%,d", perSourceCount.getValue());
+            String fileString = new File(perSourceCount.getKey()).getName();
+            perSourceCvrCountTable.getItems().add(new Pair<>(fileString, countString));
+            maxFilenameLength = Math.max(maxFilenameLength, fileString.length());
+          }
+          perSourceCvrColumnFilepath.setPrefWidth(getEstWidthForStringInPixels(maxFilenameLength));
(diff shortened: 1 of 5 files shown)


src/main/java/network/brightspots/rcv/GuiTabulateController.java
```diff
@@ -208,13 +223,30 @@ private void watchParseCvrServiceProgress(Service<LoadedCvrData> service) {
...
+          // Populate the per-source CVR count table, and calculate the width of the filename column
+          perSourceCvrCountTable.getItems().clear();
+          int maxFilenameLength = 0;
+          for (ResultsWriter.CvrSourceData sourceData : data.getCvrSourcesData()) {
+            String countString = String.format("%,d", sourceData.getNumCvrs());
+            String fileString = new File(sourceData.source.getFilePath()).getName();
+            perSourceCvrCountTable.getItems().add(new Pair<>(fileString, countString));
+            maxFilenameLength = Math.max(maxFilenameLength, fileString.length());
+          }
+          perSourceCvrColumnFilepath.setPrefWidth(getEstWidthForStringInPixels(maxFilenameLength));
```
HEdingfield, line 236: > Is there any risk of the same file name in different folders? Should we show the full path? If we do, it looks even harder to read.

One way you could handle this is by making the columns resizable by the user, and having this column justify right and set to a fixed width (so it first shows them the file name, and if they expand it, it can show the path). Alternatively, you could split the filename into one column, and the path into a separate one (maybe to the right of it).

Defer to @yezr on whether or not this is worth it.

src/main/java/network/brightspots/rcv/GuiTabulateController.java
```diff
@@ -208,13 +223,30 @@ private void watchParseCvrServiceProgress(Service<LoadedCvrData> service) {
...
+
           data.discard();
           lastLoadedCvrData = data;
         };
 
     watchGenericService(service, onSucceededEvent);
   }
 
+  private float getEstWidthForStringInPixels(int stringLength) {
+    return stringLength * 7.5f;
```
HEdingfield, line 247: I wonder if this varies given OS, monitor specs, magnification accessibility settings, etc.? Is there no autosize option we could use for a column instead?

src/main/java/network/brightspots/rcv/TabulatorSession.java
```diff
@@ -250,6 +251,7 @@ List<String> tabulate(String operatorName, LoadedCvrData expectedCvrData,
             exceptionsEncountered.add(TabulationAbortedException.class.toString());
           } else {
             runTabulationForConfig(config, castVoteRecords.getCvrs(), progress);
+            castVoteRecords.printSummary();
```
HEdingfield, line 254: Curious why this is part of this PR? Just something you wanted to add opportunistically? Or is it some weird artifact from a merge?

HEdingfield (approved):

src/main/java/network/brightspots/rcv/TabulatorSession.java
HEdingfield: Nevermind, just saw your note in #821!

src/main/java/network/brightspots/rcv/GuiTabulateController.java
artoonie: It is currently resizeable by the user, and I tried that at first but I found that manually expanding a cell in a horizontally-scrolling window is actually kind of hard: you're expanding, but so is the container, so it scrolls -- so you're resizing controls three things: the cell size, the container size, and the scroll position. This felt clunkier than the current solution.

src/main/java/network/brightspots/rcv/GuiTabulateController.java
artoonie: There is a protected method that resizes, but doing that feels so much clunkier than this estimate: https://stackoverflow.com/a/38686402

This estimate is likely to make a cell that's too wide, but that's not really an issue. It should never make a cell that's too small. I think it's dependent on screen density but not any other factors, though I haven't tested on accessibility string scaling (and wonder if our UI works at all there?)

Note that the column is resizable so the estimate doesn't need to be perfect. LMK if you prefer the method in that stackoverflow link.

HEdingfield: Fine as-is, thanks!

src/main/java/network/brightspots/rcv/GuiTabulateController.java
HEdingfield: Ok, sounds good. 👍

HEdingfield (approved):

PR #843, opened 2024-06-11
Workaround for windows log issues

### LINKED ISSUE #840: Logger hanging in Windows when 'Open Log' button fails
> In my test run (both with `2013_minneapolis_park_config.json` and `dominion_alaska_config.json`), this button didn't work and I got a warning in the logs saying:

> WARNING: Failed to open file explorer: Cannot run program "open": CreateProcess error=2, The system cannot find the file specified

Then window then stayed open and I had to X out of it to get rid of it.

The file definitely exists... I'm guessing this has to do with Windows maybe behaving differently than a Mac? I'm on Windows 11.

Another issue arising from this: I noticed that no message appeared in the logger window the first couple times I clicked, and then the third time it appeared with three messages.

_Originally posted by @HEdingfield in https://github.com/BrightSpots/rcv/pull/812#discussion_r1625226102_

artoonie:
Closes #840 

I was able to reproduce the bug, and each time it happened, I got some log spam that said:
```
INFO: index exceeds maxCellCount. Check size calculations for class network.brightspots.rcv.Logger$1
```

I spent a time investigating, but the code is so deep in the javafx windowing code that I couldn't find any clean workarounds.

The one thing that worked was simply filtering out the log spam message, which seems to make the problem go away entirely. I don't love this solution but couldn't immediately come up with another.

### DIFF
--- a/src/main/java/network/brightspots/rcv/Logger.java
+++ b/src/main/java/network/brightspots/rcv/Logger.java
@@ -274,7 +274,8 @@ class Logger {
           }
 
           private boolean shouldIgnore(LogRecord record) {
-            // On Windows, scrollToBottom can trigger a log message in VirtualFlow.java that causes:
+            // On Windows, scrollToBottom can trigger a log message in JavaFX.
+            // We'll get log spam from VirtualFlow.java that causes:
             // 1. A log message to be added to the queue
             // 2. The scroll-to-bottom to fail
             // This can cause a cycle of repeated log spam and sporadic scroll failures.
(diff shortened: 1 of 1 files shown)


src/main/java/network/brightspots/rcv/Logger.java
```diff
@@ -273,6 +273,17 @@ public void publish(LogRecord record) {
             }
           }
 
+          private boolean shouldIgnore(LogRecord record) {
+            // On Windows, scrollToBottom can trigger a log message in VirtualFlow.java that causes:
```
HEdingfield, line 277: Maybe mention that this is specific to JavaFX? `VirtualFlow.java` on its own is confusing.

HEdingfield (approved):

HEdingfield (commented):
Nice!

PR #845, opened 2024-06-12
reduce audit log spam

### LINKED ISSUE #844: Noisy logging when processing generic .CSV files as CVRs
Since generic CSV files as CVRs were first introduced in https://github.com/BrightSpots/rcv/issues/549 (and fixed in https://github.com/BrightSpots/rcv/pull/670), the logging has been a little noisy with `FINE` statements at times when tabulating an election with a CSV CVR.

```
2024-06-11 12:38:29 EDT INFO: Tabulating 'Mock Election'...
2024-06-11 12:38:29 EDT INFO: Parsing cast vote records...
2024-06-11 12:38:29 EDT INFO: Reading generic CSV cast vote records from file: /Users/malachigruenhagen/rcvrc/undefined-behaviors/2024-06-05-previous-round-tiebreak/mock_election_CVR-00.csv...
2024-06-11 12:38:29 EDT FINE: ReadOnlyBooleanProperty [bean: TextArea[id=textStatus, styleClass=text-input text-area console-text-area], name: focused, value: false] focused=false
2024-06-11 12:38:29 EDT INFO: Reading CSV cast vote record file: /Users/malachigruenhagen/rcvrc/undefined-behaviors/2024-06-05-previous-round-tiebreak/mock_election_CVR-00.csv...
2024-06-11 12:38:29 EDT FINE: ReadOnlyBooleanProperty [bean: TextArea[id=textStatus, styleClass=text-input text-area console-text-area], name: focused, value: true] focused=true
2024-06-11 12:38:29 EDT INFO: Parsed 9 cast vote records successfully.
2024-06-11 12:38:29 EDT INFO: There are 5 declared candidates for this contest:
```

This appears to have gotten worse with https://github.com/BrightSpots/rcv/commit/bf032c75bf905449e59843cfd5004ce92e4ba207.

[2024-06-11_17-10-21_audit_0.log](https://github.com/user-attachments/files/15794870/2024-06-11_17-10-21_audit_0.log)

Replication steps:

1. Clone repo & checkout the correct commit (https://github.com/BrightSpots/rcv/commit/bf032c75bf905449e59843cfd5004ce92e4ba207).
2. Run program per README.md instructions.
3. Load [src/test/resources/network/brightspots/rcv/test_data/generic_csv_test/generic_csv_test_config.json](https://github.com/BrightSpots/rcv/blob/bf032c75bf905449e59843cfd5004ce92e4ba207/src/test/resources/network/brightspots/rcv/test_data/generic_csv_test/generic_csv_test_config.json) into RCTab.
4. Tabulate the election based on that config file.
5. In the `src/test/resources/network/brightspots/rcv/test_data/generic_csv_test/output`, open up the log file. You should see a lot of extra `FINE` log statements.

artoonie:
Closes #844 

I'm not sure why the audit log includes `FINE` level logging at all -- it seems more useful to limit it to `INFO`. Thoughts on just removing all `FINE`s, rather than hunting down why certain UI elements log to `FINE`?

### DIFF
--- a/src/main/java/network/brightspots/rcv/CastVoteRecord.java
+++ b/src/main/java/network/brightspots/rcv/CastVoteRecord.java
@@ -149,7 +149,7 @@ class CastVoteRecord {
       logStringBuilder.append(" [value] ").append(fractionalTransferValue);
     }
 
-    Logger.fine(logStringBuilder.toString());
+    Logger.info(logStringBuilder.toString());
   }
 
   Map<Integer, List<Pair<String, BigDecimal>>> getCdfSnapshotData() {
--- a/src/main/java/network/brightspots/rcv/CastVoteRecord.java
+++ b/src/main/java/network/brightspots/rcv/CastVoteRecord.java
@@ -48,11 +48,11 @@ class CastVoteRecord {
   // a set is used to handle overvotes
   CandidateRankingsList candidateRankings;
   // contest associated with this CVR
-  private String contestId;
+  private final String contestId;
   // tabulatorId parsed from Dominion CVR data
-  private String tabulatorId;
+  private final String tabulatorId;
   // batchId parsed from Dominion CVR data
-  private String batchId;
+  private final String batchId;
   // the ballot status for the current round, which will change as tabulation progresses.
   private StatusForRound currentRoundStatus = StatusForRound.ACTIVE;
   // tells us which candidate is currently receiving this CVR's vote (or fractional vote)
@@ -149,7 +149,7 @@ class CastVoteRecord {
       logStringBuilder.append(" [value] ").append(fractionalTransferValue);
     }
 
-    Logger.info(logStringBuilder.toString());
+    Logger.auditable(logStringBuilder.toString());
   }
 
   Map<Integer, List<Pair<String, BigDecimal>>> getCdfSnapshotData() {
--- a/src/main/java/network/brightspots/rcv/Logger.java
+++ b/src/main/java/network/brightspots/rcv/Logger.java
@@ -14,8 +14,8 @@
  * log message
  *  |
  *  v
- * Tabulation handler (FINE) -> tabulation "audit" file
- *  When a tabulation is in progress this captures all FINE level logging including audit info.
+ * Tabulation handler (AUDIT) -> tabulation "audit" file
+ *  When a tabulation is in progress this captures all AUDIT level logging including audit info.
  *
  * Execution handler (INFO) -> execution file
  *  Captures all INFO level logging for the execution of a session.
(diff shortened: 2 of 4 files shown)


yezr:
`FINE` logging is used in different places throughout RCTab that I would not want to exclude. Like what happens to a ballot in each round
image

If we do make this change, can we first ensure we aren't using `FINE` logging in any of our logs
image

artoonie:
Hmm, my hunch is that this was designed so that these messages don't show up in the GUI, but do show up in the audit log. That seems intentional and I guess we should keep it that way.

What if we just made that explicit? I've added a new logger type called "AUDIT" which is intended to show up only in the audit log, but not in the GUI. Thoughts?

src/main/java/network/brightspots/rcv/CastVoteRecord.java
```diff
@@ -149,7 +149,7 @@ void logRoundOutcome(
       logStringBuilder.append(" [value] ").append(fractionalTransferValue);
     }
 
-    Logger.fine(logStringBuilder.toString());
+    Logger.info(logStringBuilder.toString());
```
HEdingfield, line 152: I had to revert this because it made the console even busier than it already is, even causing RCTab to crash when running `2013_minneapolis_mayor_scale`.

src/main/java/network/brightspots/rcv/Logger.java
```diff
@@ -83,7 +84,7 @@ class Logger {
 
   static void setup() {
     logger = java.util.logging.Logger.getLogger("");
-    logger.setLevel(Level.FINE);
+    logger.setLevel(AUDIT_LEVEL);
```
HEdingfield, line 87: What about line 132. Does it make sense that it should still do: `tabulationHandler.setLevel(Level.FINE)`?

src/main/java/network/brightspots/rcv/CastVoteRecord.java
artoonie: Good catch, that was unintentional

src/main/java/network/brightspots/rcv/Logger.java
artoonie: No it does not! Good catch as well. The whole point was to reduce the audit log spam from FINEs :) fixed

HEdingfield (approved):
LGTM, though it doesn't seem to reduce log file size much at all when running `2013_minneapolis_mayor`, so not sure it fully addresses #844?

artoonie:
I believe the only spam that was an issue in #844 is the `ReadOnlyBooleanProperty` spam (@nurse-the-code can confirm?), given yezr's comment above that we want the rest of the FINE details, and the Issue text noting it applies specifically to the Generic CSV.

yezr:
I updated #844 to be explicit that the issue we are trying to fix is the `ReadOnlyBooleanProperty` FINE logging noise.

PR #848, opened 2024-06-15
fix broken test

artoonie:


src/test/resources/network/brightspots/rcv/test_data/more_winners_than_candidates/more_winners_than_candidates_config.json
```diff
@@ -15,7 +15,7 @@
     "firstVoteColumnIndex" : "2",
     "firstVoteRowIndex" : "2",
     "idColumnIndex" : "1",
-    "precinctColumnIndex" : "2",
+    "precinctColumnIndex" : "",
```
HEdingfield, line 18: Whoa weird, how did that happen? Bad merge?

HEdingfield (approved):

src/test/resources/network/brightspots/rcv/test_data/more_winners_than_candidates/more_winners_than_candidates_config.json
artoonie: I think another PR merged in the interim enables “tabulate by precinct” for this CVR type, but when this test was made, the precinct column was ignored. 

We might consider Merge Queues which would prevent this. It’s a newly released GH feature but I think something Google has had internally for a while.

PR #852, opened 2024-06-19
WIP: Add tiebreak option: previous rounds, then config permutation

artoonie:
As a demo for some folks, not intended to merge without discussion

### DIFF
--- a/src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
+++ b/src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
@@ -34,6 +34,8 @@ Tiebreak Mode (required): Ties in ranked-choice voting contests can occur when e
 
   * Previous round counts (then random): The tied candidate with the least votes in the previous round loses the tie. If there is a tie in the previous round, the tie is broken randomly. Requires a random seed.
 
+  * Previous round counts (then config permutation): The tied candidate with the least votes in the previous round loses the tie. If there is a tie in the previous round, the tie is broken using the order of candidates in the config file.
+
   * Previous round counts (then stop counting and ask): The tied candidate with the least votes in the previous round loses the tie. If there is a tie in the previous round, user is prompted to select any tied candidate to eliminate or, in multi-winner contests only, elect.
 
   * Use candidate order in the config file: Use the order of candidates in the config file to determine tiebreak results. Candidates lower in the list lose the tiebreaker.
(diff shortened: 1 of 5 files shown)


src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
```diff
@@ -34,6 +34,8 @@ Tiebreak Mode (required): Ties in ranked-choice voting contests can occur when e
 
   * Previous round counts (then random): The tied candidate with the least votes in the previous round loses the tie. If there is a tie in the previous round, the tie is broken randomly. Requires a random seed.
```
bruceleban, line 35: Suggest:
> The tied candidate with the least votes in the previous round loses the tie. If there is a tie in the previous round, it looks at the round before that. If there are still tied candidates after examining all previous rounds, the tie is broken randomly. Requires a random seed.

This and following suggestions repeat text. Could restructure these to say instead:
> Looks at previous rounds to break ties (see below). If there are still tied candidates after examining previous rounds, the tie is broken randomly. Requires a random seed.

src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
```diff
@@ -34,6 +34,8 @@ Tiebreak Mode (required): Ties in ranked-choice voting contests can occur when e
...
 
+  * Previous round counts (then config permutation): The tied candidate with the least votes in the previous round loses the tie. If there is a tie in the previous round, the tie is broken using the order of candidates in the config file.
```
bruceleban, line 37: Suggest:
> The` tied candidate with the least votes in the previous round loses the tie. If there is a tie in the previous round, it looks at the round before that. If there are still tied candidates after examining all previous rounds, the tie is broken using the order of candidates in the config file.

src/main/resources/network/brightspots/rcv/hints_winning_rules.txt
```diff
@@ -34,6 +34,8 @@ Tiebreak Mode (required): Ties in ranked-choice voting contests can occur when e
...
+
   * Previous round counts (then stop counting and ask): The tied candidate with the least votes in the previous round loses the tie. If there is a tie in the previous round, user is prompted to select any tied candidate to eliminate or, in multi-winner contests only, elect.
```
bruceleban, line 39: Suggest:
> The` tied candidate with the least votes in the previous round loses the tie. If there is a tie in the previous round, it looks at the round before that. If there are still tied candidates after examining all previous rounds, the user is prompted to select any tied candidate to eliminate or, in multi-winner contests only, elect.

artoonie:
Closing this, can re-open if discussion with external stakeholders progresses

PR #857, opened 2024-06-25
Improvements to Inactive Ballot tracking, and corrected reporting of undervotes

### LINKED ISSUE #780: [2.0-RC] summary.csv "Inactive Ballots Total" is different from the sum of each inactive ballot category
RCTab Version 1.4.0
config: [2013_minneapolis_mayor_config.json](https://github.com/BrightSpots/rcv/files/13814322/2013_minneapolis_mayor_config.json)
summary.csv: [2024-01-02_16-21-57_summary.csv](https://github.com/BrightSpots/rcv/files/13814325/2024-01-02_16-21-57_summary.csv)

## What Happens
I was documenting the list of changes made to the `summary.csv` in 1.4.0 and noticed a discrepancy. For each round we show 4 different categories of inactive ballots: Overvotes, Skipped Rankings, Exhausted Choices and Repeated Rankings. Then, we have an 'Inactive Ballots Total' sum. When I add up all of the individual categories I repeatedly get less than the 'Inactive Ballots Total' cell.

image

In this specific example the Round 1 inactives add up correctly `sum(B59:62)=B63`, but each of the transfer and Round columns after that do not sum correctly (column G & H for example). 

## What I would expect to happen
I would expect for each of the transfer and Round columns that the 4 inactive category totals sum to the 'Inactive Ballots Total' cell. I confirmed with @chughes297 that the 37 number in the transfer column after Round 1 is correct for this example.

### LINKED ISSUE #854: Split `Inactive Ballots by Exhausted Choices` bucket into `Inactive Fully Ranked Ballots by Exhausted Choices` and `Inactive Partially Ranked Ballots by Exhausted Choices`
Currently we have a row for `Inactive Ballots By Exhausted Choices` in the `summary.csv`. We'd like to split this into two buckets

- first bucket, `Inactive Fully Ranked Ballots by Exhausted Choices` is for exhausted ballots that used ALL of their rankings
- second bucket, `Inactive Partially Ranked Ballots by Exhausted Choices` is for exhausted ballots that used less than the number of rankings that they were allowed

We will use the number configured in `Maximum Number of Candidates That Can Be Ranked` to differentiate. This will be required! We don't want to assume the number of maximum rankings. It must be explicitly configured.

Image

artoonie:
This was originally designed to handle both #780 and #854. However, #854 has additional questions and has been deferred out of 2.0.

Still, this PR made some improvements that resolved #780 in a nice way, so I'm modifying this PR to just resolve #780 without the "partially ranked" vs "fully ranked" split in #854.

Therefore, this PR has code that is outside of the scope of #780 but makes it easy to accomplish #854 in the future -- things like tracking whether a ballot used all of its known rankings.


src/test/resources/network/brightspots/rcv/test_data/2013_minneapolis_mayor_scale/2013_minneapolis_mayor_scale_expected_summary.csv
```diff
@@ -1,63 +1,63 @@
...
+Inactive Ballots by Skipped Rankings,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0,611,,0
+Inactive Fully Ranked Ballots by Exhausted Choices,0,,234,234,,0,234,,13,247,,13,260,,52,312,,65,377,,65,442,,26,468,,195,663,,182,845,,195,1040,,143,1183,,208,1391,,351,1742,,273,2015,,273,2288,,247,2535,,364,2899,,611,3510,,598,4108,,1157,5265,,1313,6578,,2093,8671,,1300,9971,,1235,11206,,3978,15184,,4368,19552,,3393,22945,,3068,26013,,6955,32968,,18278,51246,,101751,152997,,0
+Inactive Partially Ranked Ballots by Exhausted Choices,0,,234,234,,78,312,,52,364,,130,494,,78,572,,156,728,,91,819,,195,1014,,221,1235,,182,1417,,221,1638,,338,1976,,78,2054,,169,2223,,390,2613,,104,2717,,78,2795,,572,3367,,91,3458,,130,3588,,117,3705,,286,3991,,1404,5395,,546,5941,,338,6279,,2626,8905,,429,9334,,2535,11869,,819,12688,,2548,15236,,7267,22503,,25753,48256,,0
```
artoonie, line 62: This concerns me -- it must be a fluke of the construction of these files, so I should investigate. Why is the Fully Ranked Exhausted exactly equal Partially Ranked Exhausted?

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -1181,7 +1183,7 @@ && isCandidateContinuing(cvr.getCurrentRecipientOfVote())) {
           if (config.getMaxSkippedRanksAllowed() != Integer.MAX_VALUE
               && maxAllowedRanking - rank > config.getMaxSkippedRanksAllowed()) {
             recordSelectionForCastVoteRecord(
-                cvr, roundTally, null, StatusForRound.INACTIVE_BY_UNDERVOTE, "");
+                cvr, roundTally, null, StatusForRound.INACTIVE_BY_UNUSED_RANKINGS, "");
```
yezr, line 1186: I'm confused by this. The comments make this seem like this is from skipping too many ranks and should then be `INACTIVE_BY_SKIPPED_RANKING`? We look for skipped rankings elsewhere explicitly. Before it was treated as an undervote? Was that our strict 0 ranking undervote? But we can only get here if there is a rank...

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -376,14 +376,27 @@ private void generateSummarySpreadsheet(
...
+                     StatusForRound.INACTIVE_BY_UNUSED_RANKINGS),
         };
 
     for (Pair<String, StatusForRound> statusToPrint : statusesToPrint) {
-      csvPrinter.print("Inactive Ballots by " + statusToPrint.getKey());
+      // Skip any "Inactive by X" if no ballots are associated with that bucket
+      RoundTally lastRound = roundTallies.get(numRounds);
+      BigDecimal numUndervotesLastRound = lastRound.getBallotStatusTally(statusToPrint.getValue());
+      if (numUndervotesLastRound.equals(BigDecimal.ZERO)) {
+        continue;
```
yezr, line 396: I think the summary_extended.csv split will solve this problem. I worry that consumers could be confused if they don't see explicit 0's.

src/main/java/network/brightspots/rcv/CastVoteRecord.java
```diff
@@ -224,7 +224,8 @@ Map<String, BigDecimal> getWinnerToFractionalValue() {
   // it is active or not.
   enum StatusForRound {
     ACTIVE,
-    INACTIVE_BY_UNDERVOTE,
+    INACTIVE_BY_NO_RANKINGS,
+    INACTIVE_BY_UNUSED_RANKINGS,
```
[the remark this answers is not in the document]
artoonie: Maybe this gets less verbose if the `INACTIVE` becomes implicit, as it's not `ACTIVE`:
* ACTIVE
* DID_NOT_RANK_ANY
* EXHAUSTED_CHOICES_PARTIALLY_RANKED
* EXHAUSTED_CHOICES_FULLY_RANKED
* INVALIDATED_BY_OVERVOTE
* INVALIDATED_BY_SKIPPED_RANKINGS
* INVALIDATED_BY_REPEATED_RANKING

src/main/java/network/brightspots/rcv/ResultsWriter.java
artoonie: The alternative is the following pseudocode:

```
doShowPartiallyRankedInactive = config.userHasProvidedAMaxRanking()
foreach (CVRSource source) {
  if (source.doesThisSourceOverrideConfigsMaxRanking) {
    doShowPartiallyRankedInactive = true;
  }
}
```

But keep in mind, with this logic in a multi-source contest, only some sources will trigger this. If the config is set to MAX, a Dominion config is set to 6, and an ES&S config leaves the max unspecified, only ballots from Dominion will end up in this bucket.

Should we prevent one source from having a MAX set and another not having it set? Should we change language to clarify this possibility?

src/main/java/network/brightspots/rcv/Tabulator.java
artoonie: I have clarified the comment and ENUMs to determine if this is `EXHAUSTED_CHOICE_PARTIALLY_RANKED` or `EXHAUSTED_CHOICE_FULLY_RANKED`

src/test/resources/network/brightspots/rcv/test_data/2013_minneapolis_mayor/2013_minneapolis_mayor_expected_summary.csv
```diff
@@ -1,63 +1,63 @@
...
+Current Round Threshold,39657,,,39638,,,39635,,,39633,,,39627,,,39622,,,39614,,,39608,,,39599,,,39583,,,39568,,,39552,,,39533,,,39522,,,39502,,,39477,,,39462,,,39450,,,39414,,,39387,,,39359,,,39309,,,39247,,,39112,,,39041,,,38981,,,38726,,,38541,,,38312,,,38162,,,37796,,,36810,,,31898,,
+Inactive Ballots by Overvotes,103,,1,104,,0,104,,0,104,,0,104,,0,104,,0,104,,0,104,,0,104,,1,105,,1,106,,1,107,,0,107,,0,107,,0,107,,0,107,,0,107,,0,107,,0,107,,0,107,,0,107,,1,108,,1,109,,1,110,,0,110,,0,110,,1,111,,1,112,,2,114,,1,115,,2,117,,6,123,,17,140,,0
+Inactive Ballots by Skipped Rankings,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0
+Inactive Ballots by Repeated Rankings,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0
```
yezr, line 61: Is it possible to keep the same order throughout the summary files for repeated and exhausted? So that we can minimize diffs

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -405,14 +408,14 @@ private void generateSummarySpreadsheet(
     }
 
     csvPrinter.print("Inactive Ballots Total");
-    // Undervotes should not be included in the Inactive Ballots count, even though we treat them
-    // as such internally. Subtract undervotes (which are static throughout a contest) from the
-    // inactive ballot totals.
-    BigDecimal numUndervotes =
-        roundTallies.get(1).getBallotStatusTally(StatusForRound.INACTIVE_BY_UNDERVOTE);
+    // DID_NOT_RANK_ANY_CANDIDATES should not be included in the Inactive Ballots count, even though
+    // we treat them as such internally. Subtract no-rankings (which are static throughout a
```
yezr, line 412: What does this mean? That we treat no-rank ballots as inactive internally?

src/test/resources/network/brightspots/rcv/test_data/2013_minneapolis_mayor/2013_minneapolis_mayor_expected_summary.csv
```diff
@@ -1,63 +1,63 @@
...
+Inactive Ballots by Exhausted Choices,0,,36,36,,6,42,,5,47,,11,58,,10,68,,17,85,,12,97,,17,114,,32,146,,28,174,,32,206,,37,243,,22,265,,40,305,,51,356,,29,385,,25,410,,72,482,,54,536,,56,592,,98,690,,123,813,,269,1082,,142,1224,,121,1345,,508,1853,,369,2222,,456,2678,,299,2977,,731,3708,,1965,5673,,9808,15481,,0
```
yezr, line 62: I will dig in to find out exactly where these diffs are coming from, but do we know where this increase in exhausted ballots is coming from?

yezr: Is it the thing you described in [this comment](https://github.com/BrightSpots/rcv/issues/780#issuecomment-1894608847): that previously we weren't including exhausted ballots that we thought had been fully ranked?

artoonie: We excluded ballots that we thought had been partially ranked. We lumped them into "undervotes" ([here](https://github.com/BrightSpots/rcv/pull/857/files#diff-f37dd0767c8316149ae7da21b20e0b3b0c2394a3220ad31dc01de70425232a3eL1184)), which we excluded.

src/main/java/network/brightspots/rcv/ResultsWriter.java
artoonie: Good call, this comment is outdated now that we have a separate "Did not rank" category.

yezr:
note for when this is merged that RCVRC needs to document how this changes the description for ballots in the audit log and potentially the totally unranked ballots in the summary.csv

yezr:
@artoonie can we change the undervote description in the `summary.csv` to `Number of Undervotes (No Rankings)"

src/test/resources/network/brightspots/rcv/test_data/csv_missing_header_test/csv_missing_header_test_expected_summary.json
```diff
@@ -100,4 +100,4 @@
     "totalNumBallots" : "26",
     "undervotes" : 0
   }
-}
+}
```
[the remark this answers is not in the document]
artoonie: A "no EOL" symbol -- removing the trailing \n from the file to line up with most of the other JSONs

src/test/resources/network/brightspots/rcv/test_data/2013_minneapolis_mayor/2013_minneapolis_mayor_expected_summary.csv
```diff
@@ -1,63 +1,63 @@
...
-Inactive Ballots by Skipped Rankings,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0,47,,0
-Inactive Ballots by Exhausted Choices,0,,18,18,,0,18,,1,19,,1,20,,4,24,,5,29,,5,34,,2,36,,15,51,,14,65,,15,80,,11,91,,16,107,,27,134,,21,155,,21,176,,19,195,,28,223,,47,270,,46,316,,89,405,,101,506,,161,667,,100,767,,95,862,,306,1168,,336,1504,,261,1765,,236,2001,,535,2536,,1406,3942,,7827,11769,,0
-Inactive Ballots by Repeated Rankings,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0
-Inactive Ballots Total,150,,37,187,,6,193,,5,198,,11,209,,10,219,,17,236,,12,248,,17,265,,33,298,,29,327,,33,360,,37,397,,22,419,,40,459,,51,510,,29,539,,25,564,,72,636,,54,690,,56,746,,99,845,,124,969,,270,1239,,142,1381,,121,1502,,509,2011,,370,2381,,458,2839,,300,3139,,733,3872,,1971,5843,,9825,15668,,0
+Contest Information
```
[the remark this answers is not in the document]
artoonie: Line ending change (must've previously been created before we made all line endings the same).

If you select "Hide Whitespace" in the diff view it goes away: 
image

artoonie:
All comments addressed, let me know if there's more needed from RCVRC's side

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -1172,20 +1174,8 @@ && isCandidateContinuing(cvr.getCurrentRecipientOfVote())) {
         // if this is the last ranking we are out of rankings and must exhaust this cvr
         // determine if the reason is skipping too many ranks, or no continuing candidates
```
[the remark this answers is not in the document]
yezr: It says this comment is outdated, but I still see this in the latest 
`// determine if the reason is skipping too many ranks, or no continuing candidates`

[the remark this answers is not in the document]
artoonie: It should show as deleted, perhaps github hadn't fully refreshed when you checked? You should see it deleted now: 
image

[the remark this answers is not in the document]
yezr: Oh ok I kept navigating here by going to the PR -> Conversations -> Unresolved Conversations. Turns out using that navigation method automatically filters the PR Files tab to the commit that the original comment was on, which filters out the new commit that fixes this comment. Seems counterintuitive, a feeling  [shared by others](https://github.com/orgs/community/discussions/23138)

image

Anyways, this is fixed with the latest commit!

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -992,10 +992,14 @@ private Map<String, BigDecimal> getInactiveJsonMap(RoundTally roundTally) {
...
-          new Pair<>("exhaustedChoices", StatusForRound.INACTIVE_BY_EXHAUSTED_CHOICES),
-          new Pair<>("repeatedRankings", StatusForRound.INACTIVE_BY_REPEATED_RANKING)
+          new Pair<>("overvotes",
+                     StatusForRound.INVALIDATED_BY_OVERVOTE),
+          new Pair<>("skippedRankings",
+                     StatusForRound.INVALIDATED_BY_SKIPPED_RANKING),
+          new Pair<>("repeatedRankings",
+                     StatusForRound.INVALIDATED_BY_REPEATED_RANKING),
+          new Pair<>("exhaustedChoices",
+                     StatusForRound.EXHAUSTED_CHOICE),
```
yezr, line 1002: exhausted is extra here? Why was that not here before? It looks like this list is used to drive the properties of the summary.json. It looks like `summary.json` had this property before, just always 0's. Maybe this just provides the correct lookup for the ResultsWriter to fill it in? https://github.com/BrightSpots/rcv/pull/857/files#diff-581edbf64dc2eb31f841777e3cfdc5e462066e8cd7de90c026e68538b8b8a46cL94-R94

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -405,14 +408,11 @@ private void generateSummarySpreadsheet(
     }
 
     csvPrinter.print("Inactive Ballots Total");
-    // Undervotes should not be included in the Inactive Ballots count, even though we treat them
-    // as such internally. Subtract undervotes (which are static throughout a contest) from the
-    // inactive ballot totals.
-    BigDecimal numUndervotes =
-        roundTallies.get(1).getBallotStatusTally(StatusForRound.INACTIVE_BY_UNDERVOTE);
+    BigDecimal numNoRankings =
+        roundTallies.get(1).getBallotStatusTally(StatusForRound.DID_NOT_RANK_ANY_CANDIDATES);
```
yezr, line 412: along with https://github.com/BrightSpots/rcv/pull/857/files#diff-f37dd0767c8316149ae7da21b20e0b3b0c2394a3220ad31dc01de70425232a3eR1073 this is the crux of the change

src/main/java/network/brightspots/rcv/ResultsWriter.java
yezr: oops! it is just a different order

yezr (approved):

PR #864, opened 2024-07-02
don't treat non-text files as text in git line ending conversion

### LINKED ISSUE #863: EOL issues w/ git clone
when we clone and immediately do a git status have a long list of files that are marked as modified. I attached my list of files which is very similar but a little different from Malachi's.

It looks like it's a line ending issue of some kind. I think maybe since [#790](https://github.com/BrightSpots/rcv/pull/790) we checked in a new `.gitattributes` that sets line endings for certain files. I attached the list of files that are different right when I clone and then git status. Malachi's are similar but a little different.

[mathew's git-status.txt](https://github.com/user-attachments/files/16072537/git-status.txt) immediately after clone
[malachi's git-status.txt](https://github.com/user-attachments/files/16072542/git-status.1.txt)

What I see for git status. Notice the warning about CRLF
image

I'm leaving this out of 2.0 until we test the golden hash generation. I think this will affect getting the correct golden hash. If so, I'll add it in to 2.0.

artoonie:
closes #863 

I tested by cloning a fresh repo with just this branch and no files came back dirty. May be worth testing on windows too, but I don't currently have access to a Windows machine. You can test it by:
`git clone -b feature/issue-863_cloned-line-endings https://github.com/BrightSpots/rcv`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-* text eol=lf
+* text=auto


nurse-the-code:
I checked it on Windows. It looks good to me.

nurse-the-code:
I also checked it out on MacOS. It looked good there, too.

nurse-the-code:
Will this cause any issues with the golden hash?

yezr:
> Will this cause any issues with the golden hash?

No, according to Armin here https://github.com/BrightSpots/rcv/issues/863#issuecomment-2206904017

tarheel (approved):

PR #872, opened 2024-07-10
Updates to the per-slice CSV

### LINKED ISSUE #871: Tabulate by slice updates
Followup to #807. After doing extended testing of Tabulate by Batch I discussed some proposals w/ RCVRC HQ related to Tabulate by Batch and Tabulate by Precinct. These are referred to below collectively as 'Tabulate by Slice' 

- Include contest level Elect/Eliminate rows in all tabulate by slice summary files similar to the full contest `summary.csv`. Currently, neither Tabulate by Precinct or Tabulate by Batch include them. The hope is this will reduce confusion. The example being seeing a candidate with the most Round 1 votes eliminated in the by slice summary because of the contest level Elect/Eliminate decision.
- Remove threshold information from tabulate by slice summary files. Since the Elected/Eliminated decisions are at the contest level the threshold information can be confusing.
  - The `Final Threshold` number in row 9 for both.
  - The `Current Round Threshold` row. Currently only in the precinct summary files. 
- Use the same candidate order in Tabulate by Slice summary files to match the contest wide summary file. In my test contest that I used for #870 the sub-jurisdiction summary files had candidates listed in a different order than the contest summary file. I believe each of the Tabulate by Slice summary files were using the first round vote tallies limited to only their universe. Can we make it so that the Tabulate by Slice files have the same candidate order as the full contest summary?

artoonie:
Closes #871. Makes several changes as outlined there.


src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -533,19 +543,27 @@ private void addContestInformationRows(CSVPrinter csvPrinter,
...
     }
+
     csvPrinter.printRecord("Winner(s)", String.join(", ", winners));
-    csvPrinter.printRecord("Final Threshold", winningThreshold);
+
     if (!isNullOrBlank(sliceId)) {
+      // Only silces print the slice information
       csvPrinter.printRecord(slice, sliceId);
+    } else {
+      // Only non-silces print threshold information
```
RankWeis, line 553:
```suggestion
      // Only non-slices print threshold information
```

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -533,19 +543,27 @@ private void addContestInformationRows(CSVPrinter csvPrinter,
         winners.add(config.getNameForCandidate(candidateName));
       }
     }
+
     csvPrinter.printRecord("Winner(s)", String.join(", ", winners));
-    csvPrinter.printRecord("Final Threshold", winningThreshold);
+
     if (!isNullOrBlank(sliceId)) {
+      // Only silces print the slice information
```
yezr, line 550:
```suggestion
      // Only slices print the slice information
```

yezr:
## Testing
- confirmed candidate order
- confirmed elect/eliminate in tabulate by precinct
- confirmed removal of both thresholds in tabulate by precinct
- confirmed additional text at the bottom of summary.csv for precinct files

I don't see any tabulate by batch files currently in tests and I think we should include at least one. I tried to do this by using the contest that had already changed (Minneapolis Mayor 2013), turning on "Tabulate by Batch", and set the `Batch Column` setting in the CVR config to 1 in order to just re-use the precinct ids as batch ids. That way they should be equivalent except for the minor tabulate by batch/precinct differences. I was then gonna check in one batch summary file, the updated config and change the equivalent test to expect more files. 

However, when running the tabulation with `Precinct Column` and `Batch Column` set to 1 I got warnings that every row didn't have a batch id. I looked it the code and that's because the StreamingCvrReader only allows a cell to be one of precinct id, batch id, ballot id or an actual ranking - never more than one of those things.

image


I changed `Batch Column` to 4 to use the extra count column that ES&S CVRs spit out. Effectively, every batch id = 1. That resulted in this file that exactly matches the full contest `summary.csv` except for the tabulate by batch specific edits, as expected. 
[2024-09-06_15-15-03_1_batch_summary.csv](https://github.com/user-attachments/files/16912147/2024-09-06_15-15-03_1_batch_summary.csv)

I think we need to check one batch slice file in. What I did here seems kind of clunky. We could use the smaller contest I made to test tabulate by batch earlier 
[CSV CVRs - tab by batch test.csv](https://github.com/user-attachments/files/16912184/CSV.CVRs.-.tab.by.batch.test.csv)


Other than adding in at least one batch summary the changes LGTM! Ready for a live code review!

yezr (dismissed):
code review comments addressed
- two tabulate by batch files have been added by @artoonie 
- addActionRows logic now explicit inline, removed boolean in function call
- explicit check that the candidates we have within the slice match the candidate list sent to us from the `Tabulator`
 
Tested that the batch tabulation contest runs on my machine. LGTM! approved

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -310,16 +316,25 @@ private void generateSummarySpreadsheet(
...
-    List<String> sortedCandidates = roundTallies.get(1).getSortedCandidatesByTally();
+    // Check that all candidates are included in the candidate order
+    Set<String> expectedCandidates = roundTallies.get(1).getCandidates();
+    Set<String> providedCandidates = new HashSet<>(candidateOrder);
+    if (!expectedCandidates.equals(providedCandidates)) {
+      throw new IllegalArgumentException(
+              "Candidate order must include all candidates in the contest. "
+                      + "\nExpected: " + expectedCandidates
+                      + "\nProvided: " + providedCandidates);
+    }
```
alyssahursh, line 334: Should this stanza come first in the method? My concern is that at this point in the method, we've already created a file and started writing to it, so we'd be in kind of an awkward state if this throws.

artoonie: agreed!

yezr (dismissed):
LGTM

yezr (approved):
approving #883 and #884 merge

PR #873, opened 2024-07-11
debugging info in golden hash file

artoonie:
Underneath the golden hash, include the file-by-file hash

### DIFF
--- a/.github/actions/sha-of-zip.bat
+++ b/.github/actions/sha-of-zip.bat
@@ -58,8 +58,9 @@ powershell -Command "& {[IO.File]::WriteAllText(\"%HASHFILE_SORTED%\", $([IO.Fil
 :: echo the final hash
 C:\Windows\System32\certutil.exe -hashfile %HASHFILE_SORTED% SHA%SHA_A% | findstr /v ":"
 
-:: For debugging, enable printing the file-by-file hash
-:: echo File-by-file hash
-:: type "%HASHFILE_SORTED%"
+:: For easier debugging, print the file-by-file hash
+echo ##########
+echo This checksum was created by a SHA-%SHA_A% of the following file, after removing the filenames and sorting by their SHAs:
+sort "%HASHFILE_SORTED%"
 
 endlocal
--- a/.github/actions/sha-of-zip.sh
+++ b/.github/actions/sha-of-zip.sh
@@ -15,7 +15,9 @@ parentPath=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 # Make a temporary directory to extract zip, and a temporary file to hold SHAs
 tempDirectory=$(mktemp -d)
 tempAllChecksumsFile=$(mktemp)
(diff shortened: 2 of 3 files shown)


nurse-the-code:
@artoonie thanks for creating this PR! This made the debugging of the golden hash issues a lot easier.

.github/actions/sha-of-zip.bat
```diff
@@ -58,8 +58,9 @@ powershell -Command "& {[IO.File]::WriteAllText(\"%HASHFILE_SORTED%\", $([IO.Fil
 :: echo the final hash
 C:\Windows\System32\certutil.exe -hashfile %HASHFILE_SORTED% SHA%SHA_A% | findstr /v ":"
 
-:: For debugging, enable printing the file-by-file hash
-:: echo File-by-file hash
-:: type "%HASHFILE_SORTED%"
+:: For easier debugging, print the file-by-file hash
+echo ##########
+echo This checksum was created by a SHA-%SHA_A% of the following file, after removing the filenames and sorting by their SHAs:
```
nurse-the-code, line 63: Should we sort by SHA or by file name?

artoonie: Sorting by SHA makes it a little more resilient to where the golden hash script lives without needing to guess at the root folder of the repo. For example, if we included filenames, then `file.txt = hash` and `rcv/file.txt = hash` would result in different golden hashes.

In the golden hash file, underneath the golden hash, we will now print the filename and the hash which should resolve the debugging need.

I'm not opposed to putting in some effort to include the filenames in the golden hash itself, but I think this PR already gives us all the benefits of that task?

nurse-the-code: Ok, looking at this again, we are hashing a list of all of the hashed. Do I understand this correctly?

I think I may have been a little confused in my previous statement/question. I think I was trying to say that I thought the final output file with the golden hashes should list the hashes of the other hashes sorted by file name.

The reason why I wanted that was in case the golden hash doesn't match the expected value, you could easily compare it to another golden hash output file and find the non-matching SHA.

I hope I am explaining myself well. Let me know if you have any questions.

artoonie: That should be exactly what this PR does, though let me know if you see otherwise! For example, here's one of the output files: 
[goldenhash.txt](https://github.com/user-attachments/files/16664650/goldenhash.txt)

yezr: I found the last build run from this branch [here](https://github.com/BrightSpots/rcv/actions/runs/9897495841/job/27342112757) and confirmed the golden hash file has the hashes sorted by filename 
[rctab_feature_golden-hash-debugging_Windows_X64.zip.golden.sha512.txt](https://github.com/user-attachments/files/16803336/rctab_feature_golden-hash-debugging_Windows_X64.zip.golden.sha512.txt)

yezr (approved):

PR #875, opened 2024-07-25
fix some math on the per-slice inactive ballots

### LINKED ISSUE #870: [1.3.999] sub-jurisdiction summary.csv not reporting inactive buckets
**RCTab Version**: 1.3.999
**config**: [tab by batch and precinct some empty w repeat ranking.json](https://github.com/user-attachments/files/16089798/tab.by.batch.and.precinct.some.empty.w.repeat.ranking.json)
**cvr**: [tab by batch test cvr empty precinct and batch w: repeat ranking.xlsx](https://github.com/user-attachments/files/16089802/tab.by.batch.test.cvr.empty.precinct.and.batch.w.repeat.ranking.xlsx)

### What happened
Created a contest to test Tabulate By Batch. That CVR has A & B tying in the full contest and the following sub-jurisdictions
- **Precinct 1 // Batch x** marked on the same records to confirm that precinct and batch output files should be effectively the same. A wins these first place votes by one
- **Unmarked Precinct // Batch** empty precinct/batch columns to make sure that edge case is handled correctly. It creates `missing_precinct_id` and `missing_batch_id` summary files. B wins these first place votes by 1
image

It correctly splits vote totals across the candidates within the sub-jurisdiction summary files. However, there is a single vote (id=18) that is a C repeat ranking to test that inactive buckets are being correctly filled in sub-jurisdiction files. In the full contest `summary.csv` it is correctly transferred to the `Inactive Ballot by Repeat Rankings` bucket during the post-Round 1 transfer when C is eliminated.
image

However, the `Inactive Ballots by Repeat Rankings` bucket in both the precinct and batch`summary.csv` files remains 0.
image

### What I would Expect
In unmarked precinct and batch `summary.csv`'s I would expect the inactive bucket (both for repeat ranking and the sum inactive bucket) for the post-Round 1 transfer to correctly capture the repeat ranking. 

https://github.com/BrightSpots/rcv/issues/265 and its PR https://github.com/BrightSpots/rcv/pull/335 might be related?

artoonie:
Closes #870 

In the first commit, I fix a bug where the residual surplus is not calculated for by-slice structures. However, it added a lot of complexity for no benefit: the residual surplus is not printed in by-slice files. I have removed it, but in its place, I have updated a test for by-slice tabulation on multi-winner elections which we didn't have before. That will ensure that we'll have a test to ensure that residual surplus continues to not be included in the output by-slice files.

Also fixes a GUI bug where you couldn't close the tabulation window if an exception was thrown (should really be a separate PR, but it's small...)


yezr:
confirmed with testing that the inactive buckets are now being reported correctly in the by slice summary files

src/test/resources/network/brightspots/rcv/test_data/minneapolis_multi_seat_threshold/minneapolis_multi_seat_threshold_cvr.xlsx
yezr, line 1: image
so this is just adding in some precincts to test?

src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -600,7 +600,7 @@ void test2018MaineGovPrimaryDem() {
   @Test
   @DisplayName("testMinneapolisMultiSeatThreshold")
   void testMinneapolisMultiSeatThreshold() {
-    runTabulationTest("minneapolis_multi_seat_threshold");
```
yezr, line 603: So this CVR has been altered to include precinct numbers. There are three files checked in: two precinct_summary.csv's and a precinct_summary.json. What exactly are we trying to get test coverage of by including those? Why only one .json?

src/test/resources/network/brightspots/rcv/test_data/2013_minneapolis_mayor/2013_minneapolis_mayor_expected_MINNEAPOLIS_W-1_P-01_precinct_summary.csv
```diff
@@ -13,8 +13,8 @@ Precinct,MINNEAPOLIS W-1 P-01
 Contest Summary
 Number to be Elected,1
 Number of Candidates,36
-Total Number of Ballots,517
-Number of Undervotes (No Rankings),0
```
yezr, line 17: ok looking at the CVR it does look like there is a single undervote which I would expect to be correctly identified now with the fix.
image
But I'm trying to hunt down why the Total increases by 2. I confirmed in the CVR that there are 519 rows with `W-1 P-01` precinct. 
image

The other precinct file for this contest also increased by two...

src/test/resources/network/brightspots/rcv/test_data/minneapolis_multi_seat_threshold/minneapolis_multi_seat_threshold_cvr.xlsx
artoonie: Correct, since none of our existing Tabulate-by-Silce tests were STV

src/test/resources/network/brightspots/rcv/test_data/2013_minneapolis_mayor/2013_minneapolis_mayor_expected_MINNEAPOLIS_W-1_P-01_precinct_summary.csv
artoonie: It looks like:
1. One additional undervote from voter 49004
2. One additional overvote from voter 6938

src/test/resources/network/brightspots/rcv/test_data/2017_minneapolis_mayor/2017_minneapolis_mayor_expected_MINNEAPOLIS_W-13_P-13_precinct_summary.csv
```diff
@@ -40,6 +40,6 @@ Active Ballots,953,,,952,,,943,,,939,,,866,,,789,,
 Current Round Threshold,52243,,,52200,,,50933,,,49874,,,46790,,,40838,,
 Inactive Ballots by Overvotes,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0
 Inactive Ballots by Skipped Rankings,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0
-Inactive Ballots by Exhausted Choices,0,,0,0,,0,0,,0,0,,0,0,,0,0,,0
+Inactive Ballots by Exhausted Choices,1,,1,2,,9,11,,4,15,,73,88,,77,165,,0
```
yezr, line 43: How do I get a first round Exhausted ballot? If the only marking is an overvote and the overvote rule is lenient enough to not inactivate the ballot

src/test/resources/network/brightspots/rcv/test_data/2013_minneapolis_mayor/2013_minneapolis_mayor_expected_MINNEAPOLIS_W-1_P-01_precinct_summary.csv
yezr: thank you for the point in the right direction. Confirmed that the "Total Number of Ballots" in the slice file is the active ballots in round 1. Since previously we weren't sending inactive ballot counts down to the slices the round 1 count was previously just active ballots. Examples of inactive round 1 ballots are 
- complete undervotes 
- ballots that have only overvotes AND the overvote rule is not so strict that they are inactivated

src/test/resources/network/brightspots/rcv/test_data/2017_minneapolis_mayor/2017_minneapolis_mayor_expected_MINNEAPOLIS_W-13_P-13_precinct_summary.csv
artoonie: This is a funny consequence of EXHAUSTED encapsulating both FULLY_RANKED and PARTIALLY_RANKED.

Specifically, in this case, the first round is an "overvote" and the overvote decision is "skip to next rank". We mark these as exhausted:
```
            // If the final ranking is an overvote, even if we're trying to skip to the next rank,
            // we consider this inactive by exhausted choices -- not an overvote.
```

I don't think there's an obvious answer to what these should be marked as. EXHAUSTED seems good as any.

yezr: I think exhausted is the right bucket for these. I wanted to make sure I documented this for when I'm looking back later and forget why I can have an exhausted ballot in Round 1

yezr (approved):

Issue #856, opened 2024-06-24
Two different results report files, one with ALL details and one with a summary of the data


yezr:
We’ve been adding more granular inactive ballot buckets to help with reporting. These granular buckets are important for reporting on RCV contests, but are starting to take a lot of real estate and grab maybe too much attention in the `summary.csv`. So we’re proposing to split the current `summary.csv` into two separate files…

### summary_report.csv
Will include most of the 2.0 updates like round by round vote percentages and round by round transfers but will NOT include the split out inactive buckets. It will have just one "Inactive Ballots Total" row similar to all output before 2.0.
- Number of undervotes in round 0
- For each candidate, each round
  - % of total votes for each round
  - Number of ballots transferred from eliminated candidate(s) in the previous round
- ~Inactive ballots in each round broken down by category~
  - ~Overvote~
  - ~Skipped Ranking~
  - ~Repeat Ranking~
  - ~Exhausted Choices~
- Per-Round winning threshold
- RCTab version number
- Contest Summary
- Number to be Elected
- Number of Candidates
- Total Number of Ballots

### detailed_report.csv 

will include all of the current 2.0 updates.

### detailed_report.json
`summary.json` will include everything and be renamed `detailed_report.json`.

### Tabulate by Slice
 Tabulate by slice files will include everything and be renamed `[sliceId]_[sliceType]_detailed_report.[csv|json]`

artoonie:
Should an extended summary, a simplified summary, or both be generated for slices?

yezr:
tabulate by slice file name shouldn't change and should include all the information as it currently does. So as it is right now, no updates to tab by slice are necessary

artoonie:
Noted. Flagging potential confusion as `x_summary_extended.csv` will correspond with `x_by_slice_summary.csv`

yezr:
> Noted. Flagging potential confusion as `x_summary_extended.csv` will correspond with `x_by_slice_summary.csv`

Noting this, can we rename the current tabulate by slice files `x_by_slice_summary_extended.csv` to prevent this confusion.   The original motivation for this ticket was at the contest level: for jurisdictions that don't want to have to be responsible for inactive bucket data explanations they would have an option to just post the summary file and not the summary_extended file. My thinking is that if someone is already digging into the tabulate by slice files they are going to want extended information no matter what.

artoonie:
done!

yezr:
title and original description updated based on conversations with RCVRC HQ

PR #884, opened 2024-09-04
STV last round use "Final Round Surplus" rather than inactive

### LINKED ISSUE #855: Add 'Final Round Surplus' row in summary.csv
**RCTab Version**: 1.3.999
Config [vmsg surplus transfer test.json](https://github.com/user-attachments/files/15888521/vmsg.surplus.transfer.test.json)
CVR [vmwg cvr.xlsx](https://github.com/user-attachments/files/15888524/vmwg.cvr.xlsx)

In STV, when the final candidate(s) are elected we have a final transfer round to remove any votes over the threshold that are unneeded to elect that candidate. This step ensures that all elected candidates are shown with the same vote total in the final round (otherwise, the last candidate(s) to be elected would show a higher vote total than candidates elected before them). Currently, those surplus votes over the threshold are moved to the Inactive Ballots bucket in the final transfer round. We should make a new bucket, `Final Round Surplus`, under Residual Surplus, and put any surplus votes unneeded to elect in there. This would show up in both summary.csv and summary_extended.csv

I added an example below. The sum of everything in the `a` box should equal `b`. Only the final round where surplus transfers are made from the last elected candidate(s) should have a number for  Final Round Surplus, others should be empty.  
 

Image

artoonie:
There are a few options for implementation here.

I chose to treat the final round surplus similar to how we treat inactive ballots, rather than how we treat residual surplus. Residual surplus is treated as votes that sort of "disappear," showing up in the transfer list but not elsewhere. In this case, we want the final round to not make these votes disappear, but rather, land in a non-candidate and non-inactive bucket.

I played around with other ideas but this felt both the most correct and led to the most natural implementation.

Closes #855


yezr:
### Testing
- Confirmed for all the races that had `summary.csv` changes that the new value in `Final Round Transfer` exactly matched the difference that no longer went to other inactive buckets

Need to figure out what is up with src/test/resources/network/brightspots/rcv/test_data/test_set_allow_only_one_winner_per_round/test_set_allow_only_one_winner_per_round_expected_summary.json issue during code review

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -1013,17 +1028,20 @@ private Map<String, BigDecimal> updateCandidateNamesInTally(RoundTally roundSumm
...
-          new Pair<>("overvotes",
-                     StatusForRound.INVALIDATED_BY_OVERVOTE),
-          new Pair<>("skippedRankings",
-                     StatusForRound.INVALIDATED_BY_SKIPPED_RANKING),
-          new Pair<>("repeatedRankings",
-                     StatusForRound.INVALIDATED_BY_REPEATED_RANKING),
-          new Pair<>("exhaustedChoices",
-                     StatusForRound.EXHAUSTED_CHOICE),
-        };
+    List<Pair<String, StatusForRound>> statusesToPrint = new ArrayList<>();
```
alyssahursh, line 1031: There's something about this code (both the old and the new) that has me scratching my head. What would you think about moving these "key" strings into the `StatusForRound` enum as, say, `jsonKey`? Then the logic here might look like this:

```
private Map<String, BigDecimal> getInactiveJsonMap(RoundTally roundTally) {
    List<StatusForRound> statusesToPrint = new ArrayList<>();
    statusesToPrint.add(INVALIDATED_BY_OVERVOTE);
    statusesToPrint.add(INVALIDATED_BY_SKIPPED_RANKING);
    statusesToPrint.add(EXHAUSTED_CHOICE);
    statusesToPrint.add(INVALIDATED_BY_REPEATED_RANKING);

   if (config.usesSurpluses() && roundTally.getRoundNumber() == numRounds) {
      statusesToPrint.add(FINAL_ROUND_SURPLUS);
    }

    return statusesToPrint.stream()
        .collect(Collectors.toMap(StatusForRound::jsonKey, status -> roundTally.getBallotStatusTally(status)));
}
```

alyssahursh (approved):
Nice to meet you today, Armin! Sorry to drop so abruptly without a proper goodbye. I had a few extra comments for you on this PR. If you find them useful, great! If not, that's fine too :)

yezr (approved):
LGTM. All code review comments addressed
- `StatusForRound` enum now includes single source of truth for which ones are active/inactive and their `audit.log` text
- logic for whether or not the configuration of a contest will require a Final Round Surplus is pushed up into the `ContestConfig`
- active/inactive count in `Summary` fixed to properly account for `FINAL_ROUND_TRANSFER`

yezr:
I approved thinking that I had seen all of Alyssa's comments but for some reason some of them were not showing when I was looking at the diff of just the two commits since our code review. The ones I weren't seeing were the last two that were suggestions. I thought the only outstanding PR conversations were simply comments on how the updates fixed exactly what we had talked about in the code review. Since those two are suggestions for updates and not requirements for changes I will leave it to Armin to resolve those

I used the shift + click to select the last two commits but it didn't two of Alyssa's four comments were not visible in that diff screen. Weird...
image

src/main/java/network/brightspots/rcv/ResultsWriter.java
artoonie: Note that we have different strings for different outputs -- the JSON file uses camelCase and the CSV uses Title Plain Text (with spaces). I fear that this difference doesn't belong inside StatusForRound itself (e.g. `plaintextString` vs `camelCaseString`), though could see an argument for standardizing the text across the two formats. That would have a lot of downstream side-effects. Thoughts?

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -1007,22 +1006,12 @@ private void recordSelectionForCastVoteRecord(
...
-          "invalidated by skipped ranking" + additionalLogText;
-      case INVALIDATED_BY_REPEATED_RANKING -> outcomeDescription =
-          "invalidated by repeated ranking" + additionalLogText;
-      default ->
-      // Programming error: we missed a status here
-      throw new RuntimeException("Unexpected ballot status: " + statusForRound);
+    if (statusForRound == StatusForRound.ACTIVE) {
+      outcomeDescription = selectedCandidate;
+    } else {
+      outcomeDescription = statusForRound.getPlaintext() + additionalLogText;
```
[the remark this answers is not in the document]
artoonie: Done! And made the subsequent line `final` as well to match.

src/main/java/network/brightspots/rcv/ResultsWriter.java
yezr: Though #856 doesn't require different strings, I could see other future output formats possibly requiring other strings to match to each of the `StatusForRound` enums. The `.json` vs `.csv` formats are similar enough that it seems silly to have both, but other formats could be different enough and this spot seems as good a place as any.

alyssahursh: @artoonie You have more experience in this code than I do, so I'll certainly defer to your judgment. I'm just running off intuition here, and my intuition says that all of the different ways of titling these statuses should be held somewhere in common rather than spread around in different files. My concern is that by spreading these different strings around, we're likely to encourage a third or fourth formatting standard to crop up unnecessarily. But my familiarity with this code base is very thin, and I'm not here to block you!

artoonie: Got it! I tried it out and I do think it's cleaner as you suggested, though we previously had three different string types not two:
- camel case, brief (e.g. "overvote")
- title case, using "invalidated" (e.g. "Invalidated by Overvote")
- title case, using "inactive" (e.g. "Inactive by Overvote")

The "invalidated" phrase was only used in the logs. @yezr, I don't recall if that was an intentional choice to use it in the audit logs but not in the output CSV. Are you okay using "inactive" for both, or shall I add a third string to the enum in the last commit here?

yezr: Using "inactive" for both will work.

artoonie: Great, then I'd like a final review on the last commit from Alyssa then I think we're ready to merge all three?

Thanks for the great feedback, Alyssa! This process has been super helpful.

src/main/java/network/brightspots/rcv/ResultsWriter.java
```diff
@@ -1028,23 +1024,18 @@ private Map<String, BigDecimal> updateCandidateNamesInTally(RoundTally roundSumm
...
         && roundTally.getRoundNumber() == numRounds) {
-      statusesToPrint.add(new Pair<>("finalRoundSurplus",
-              StatusForRound.FINAL_ROUND_SURPLUS));
+      statusesToPrint.add(StatusForRound.FINAL_ROUND_SURPLUS);
     }
-    for (Pair<String, StatusForRound> statusToPrint : statusesToPrint) {
+    for (StatusForRound statusToPrint : statusesToPrint) {
       inactiveMap.put(
-          statusToPrint.getKey(), roundTally.getBallotStatusTally(statusToPrint.getValue()));
+          statusToPrint.getCamelCaseKey(), roundTally.getBallotStatusTally(statusToPrint));
```
alyssahursh, line 1038: I see that you didn't love the stream suggestion! Since you're using a modifiable map here, you could use the static final immutable list described above like this:
```
private Map<String, BigDecimal> getInactiveJsonMap(RoundTally roundTally) {
    Map<String, BigDecimal> result = new HashMap<>();
    for (StatusForRound statusToPrint : STATUSES_TO_PRINT) {
        result.put(statusToPrint.getCamelCaseKey(), roundTally.getBallotStatusTally(statusToPrint));
    }

    if (config.usesSurpluses() && roundTally.getRoundNumber() == numRounds) {
        result.put(FINAL_ROUND_SURPLUS.getCamelCaseKey(), roundTally.getBallotStatusTally(FINAL_ROUND_SURPLUS));
    }

    return result;
}
```

Again, not asking for changes in this PR! Just pointing out what I see :) If it's useful to you, great! Overall, I think the changes that have already been made are a step in the right direction.

alyssahursh (commented):
I know that I was able to approve prior versions, but I somehow can't approve today ("Only users with explicit access to this repository may approve pull requests.") I left two more comments, but I'm not requesting changes. Consider this comment my approval!

yezr:
> I know that I was able to approve prior versions, but I somehow can't approve today ("Only users with explicit access to this repository may approve pull requests.") I left two more comments, but I'm not requesting changes. Consider this comment my approval!

I got to get the approval stuff figured out. Right now anyone can approve and that allows a merge. We don't have a more intermediate "I am done reviewing all the changes" that still allows a later "official" approval for the merge to make its way to `dev`. For now if you can just comment in the PR like this to flag that a review is finished. 

The update to make this list `final` since then LGTM. Approved

PR #885, opened 2024-09-15
fix releases: need unique filenames for checksums.csv

artoonie:


nurse-the-code:
Perhaps this could be accomplished by having the file name be a complete path?

nurse-the-code:
## ChatGPT Summary of this PR:

The PR appears to be related to the release process for a Gradle-based project, where checksums, ZIP file names, and platform-specific extensions are generated as part of the build and release process.

**Key Change:**

In the workflow, the line:
```bash
echo "FILEPATH=cache/checksums.csv" >> $GITHUB_OUTPUT
```
was changed to:
```bash
echo "FILEPATH=cache/${{ steps.basefn.outputs.FILEPATH }}.checksums.csv" >> $GITHUB_OUTPUT
```

**Explanation:**

1. Original Line: The original line hard-coded the output file path for checksums as cache/checksums.csv. This means that every build would generate a checksum file with this static name, which could potentially cause issues if there are multiple releases or builds in parallel or cached files are reused.
2. Updated Line: The new line dynamically generates a checksum filename by appending .checksums.csv to a base file name, which is extracted from a previous step (referenced via ${{ steps.basefn.outputs.FILEPATH }}). This dynamic filename ensures that each build/release gets its unique checksum file, avoiding conflicts and collisions during multiple builds or releases.

**Motivation:**

The primary motivation of this change is likely to prevent collisions when multiple builds/releases occur. By using a dynamic filename that is based on the artifacts being generated in the build process, the release workflow becomes more robust.

**Broader Context in the PR:**

- The rest of the workflow seems to involve generating various artifacts (ZIP files, checksums, executables) based on platform-specific extensions (.exe, .deb, .dmg).
- The dynamic nature of filenames in this PR aims to improve the stability of the automated release process.

yezr (approved):
LGTM to try on our next automated build

yezr:
@artoonie is there a way to test this earlier than a fortnight from now

artoonie:
@yezr  I added a button to allow manual dispatch of the event. I have not tried it before but it should allow us to test the deployment both on develop and on other branches: https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow

But it needs to be merged to develop before that button is (theoretically) enabled.

nurse-the-code:
@yezr a commit was added to this after you approved it.

yezr:
thanks for the heads up, I reviewed both. And moving forward with the changes we made this morning it will rescind the approval with any code update

PR #899, opened 2024-11-11
898: Dominion CvrExport parsing

### LINKED ISSUE #898: Dominion CvrExport_N.json parsing
RCTab Version: 1.3.999

When parsing CVR files from Dominion RCTab assumes either
- There is a single file named `CvrExport.json`
- There are multiple files named `CvrExport_N.json` starting at `CvrExport_0.json` and going in increasing sequential order

We have an example where [neither are the case](https://drive.google.com/drive/u/0/folders/1x2gn8ZwdL9Y8ZdCXN3nf-CjMSWK9OTXq). In this example, the files go from `CvrExport_182.json` to `CvrExport_767.json` with gaps in the sequential numbering. 

We should make the CVR parsing more robust to handle this case.

yezr:
fixes #898


src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -230,30 +232,48 @@ private void validateNamesAreInContest(List<CastVoteRecord> castVoteRecords)
...
+        // Use regex to find all CvrExport_N.json files
+        File[] files = Paths.get(cvrPath).toFile().listFiles();
+        if (files != null) {
+          for (File file : files) {
+            Matcher matcher = pattern.matcher(file.getName());
+            if (matcher.matches()) {
+              matchedCvrFiles.add(file);
+            }
+          }
+        }
```
artoonie, line 253: Everything above, up to the else, can be simplified with:

```
    String regexPath = CVR_EXPORT_PATTERN.replaceAll("%d", "\\\\d+");
    File cvrDirectory = new File(cvrPath);
    File[] files = cvrDirectory.listFiles((dir, name) -> name.matches(pattern));
```

which has the added benefit of not duplicating the `Cvr_Export` string

src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -230,30 +232,48 @@ private void validateNamesAreInContest(List<CastVoteRecord> castVoteRecords)
...
+
         int recordsParsed = 0;
+        int filesParsed = 0;
         int recordsParsedAtLastLog = 0;
-        int cvrSequence = 0;
-        Path cvrFilePath = Paths.get(cvrPath, String.format(CVR_EXPORT_PATTERN, cvrSequence));
-        while (cvrFilePath.toFile().exists()) {
-          HashMap json = JsonParser.readFromFile(cvrFilePath.toString(), HashMap.class);
+
+        for (File file : matchedCvrFiles) {
```
artoonie, line 265: Could order of files read matter here? I fear it might with some of the randomizers, but not totally sure if any randomizer works at the CVR level.

For safety, I might sort the `matchedCvrFiles` just in case it does matter, now or later, so we generate the records in a deterministic order.

artoonie (approved):

PR #900, opened 2024-11-11
Dominion CVR file search should handle gaps and nonzero starts

### LINKED ISSUE #898: Dominion CvrExport_N.json parsing

artoonie:
Closes #898 -- some dominion exports do not start at zero and the filenames may not be sequential

The test simply renames Index 0 to Index 6, so that we test both non-sequential numbers and non-zero starting indices


artoonie:
Closing in favor of #899

PR #904, opened 2024-11-18
Update style to look a touch more modern

artoonie:
This updates the theme a touch to add padding and colors and make it look a tiny bit more modern.

| Before    | After |
| -------- | ------- |
| Screenshot 2024-11-18 at 12 31 34 PM  | Screenshot 2024-11-18 at 12 31 04 PM  |
| Screenshot 2024-11-18 at 12 31 39 PM | Screenshot 2024-11-18 at 12 31 09 PM   |


artoonie:
FYI I've fixed the issues shown in the screenshots above

artoonie:
Good eye! All addressed here:

image


As for (3), you mean changing the overall background color? I don't see much of a visual difference between white and the light gray we have on my screen.

HyltonE:
One word of warning when making changes like this: please try to test in all 3 OSes (Windows, Linux, MacOS) before committing, as we can see UX regressions when messing with padding, fonts, etc.

Not 100% required, but definitely recommended.

HEdingfield:
One word of warning when making changes like this: please try to test in all 3 OSes (Windows, Linux, MacOS) before committing, as we can see UX regressions when messing with padding, fonts, etc.

Not 100% required, but definitely recommended.

PR #908, opened 2024-11-25
Have dominion autoload use both names and codes (alt)

### LINKED ISSUE #866: Dominion: auto-load candidate name and alias correctly
**RCTab version**: 1.3.999
**config**: [Gov Final Two Rules.json](https://github.com/user-attachments/files/16088567/Gov.Final.Two.Rules.json)
**cvr**: [Alaska 2022 CVR](https://drive.google.com/file/d/10SXByEThrS902xPGOdIPGPfZeYGYWmSe/view?usp=drive_link)

### What Happens
Right now for Dominion candidate auto-load it populates the candidate name field with the candidate code from that CVR. That's not super useful as users will need their results file to show the candidate's actual name, not their candidate code, in the results. In order to get the candidate names affiliated with those candidate codes, users need to:
1. review their CVR data to match up the candidate code numbers populated into the Candidates tab 
2. record those candidate names into RCTab
3. move the candidate codes over to the aliases/code column

This doesn't really save any time as finding that candidate name info takes just as much time to do after you have candidate codes as before you have any candidate codes.

image

### What should happen
Candidate auto-load from Dominion files should load candidate names into the candidate name column and candidate codes into the code column. 
image

artoonie:
Alternate approach to #901 
Closes #866

Rather than reading the full CVR then cross-checking the CandidateManifest, only use the candidate manifest.

This speeds up autoloading, but has the downside of including `(undeclared)` as a candidate.

This splits autoloading and undeclared candidate searching into two different steps. Theoretically this should not be a problem: there should be no candidate found in the CVRs that are not found in the candidate manifest.

Let's discuss which approach feels better. I think this is a little cleaner than the other PR.


yezr:
I'm using the `dominion_multi_file` contest in the test data. 

I get every single candidate in the CVRs using this code. Can we use the configured contest id to filter to just the contest we're interested in?

Auto-Loaded with this code
alt-candidate list

Old configured list for the test data
old candidate list

src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -51,7 +51,7 @@ class DominionCvrReader extends BaseCvrReader {
   private Map<Integer, String> precinctPortions;
   // map of contest ID to Contest data
   private Map<String, Contest> contests;
-  private List<Candidate> candidates;
+  private HashMap<String, Candidate> candidates;
```
yezr, line 54: same as another comment: is it possible to signal what this string is in the HashMap? Maybe by changing the variable to candidateCodeToCandidate or candidatesByCode or something like that.

yezr:
Added to [3.0 dreams](https://docs.google.com/document/d/1M1drXbR7uoRSnKRQwMZHRFZ4GTjagIeRH7KHc2cCGf0/edit?tab=t.0) our discussion about reviewing the `GatherUnknownCandidateCounts()` implementation.

yezr (approved):
Approved following 11/14 code review!

PR #913, opened 2025-01-09
Standardize CVR identification across audit logs and RCTab output

nurse-the-code:
**Key Changes:**
- Clean up `CastVoteRecord` identifier logic
  - `getId` - guaranteed unique identifier per ballot across tabulations. coalesce(`computedId`, `suppliedId`)
  - `getSuppliedId` - vendor provided Id. Not reliably unique. Nullable
- Uses `getId` throughout audit log and `rctab_cvr.csv`
- `rctab_cvr.csv` now explicitly includes both. `getId` provides the `"RCTab CVR Id"` and `getSuppliedId` provides the `"Vendor Id"` (formerly `"Ballot Id"`)
- Changes Dominion `computedId` delimiter from pipe (`|`) to dash (`-`) for improved readability

**Testing Needed:**
1. Verify ID consistency:
   - ✅ Compare IDs in audit log with `rctab_cvr.csv` output
   - ✅ Ensure id returned from `getId` method is consistent across multiple tabulations
   - ✅ Check ID format follows tabulator-batch-record pattern for records from Dominion CVR sources
2. Test fallback behavior:
   - ✅ Cases where computedId is null
   - ✅ Cases where computedId is empty or whitespace
   - ✅ Cases where only suppliedId is available
3. Test with each CVR provider
   - ✅ CDF/Unisyn
   - ✅ Clear Ballot
   - ✅ Dominion (tested with 2024 Alaska US House race)
   - ✅ ES&S (in test suite)
   - ✅ Hart
   - ✅ generic CSV
4. Review and update failing tests:
   - ✅ Done for now


src/main/java/network/brightspots/rcv/CastVoteRecord.java
```diff
@@ -128,6 +128,12 @@ boolean doesUseLastAllowedRanking() {
     return usesLastAllowedRanking;
   }
 
+  // This represents the canonical ID used for audit logs and RCTab CVR
+  String getPrimaryId() {
+    return !isNullOrBlank(computedId) ? computedId : suppliedId;
+  }
```
nurse-the-code, line 134: This is a refactoring of the `computedId` and `suppliedId` logic used below in the `logStringBuilder` (which should print a unique CVR Id to the audit log.

src/main/java/network/brightspots/rcv/CastVoteRecord.java
```diff
@@ -128,6 +128,12 @@ boolean doesUseLastAllowedRanking() {
...
+
+  // TODO: Before merging PR determine if we still need getId & the Record ID it prints to the RCTab CVR
   String getId() {
     return suppliedId != null ? suppliedId : computedId;
   }
```
nurse-the-code, line 139: This is the logic that is used to generate the "Record Id" that gets printed out in the RCTab CVR. It is not clear to me if this was intended to be a unique CVR identifier, because it uses the same `computedId` and `suppliedId` that my `getPrimaryId` uses.

But this value returned by `getId()` is not unique to every CVR (tested on 2024 Alaska US House race).

nurse-the-code: @yezr thinks we don't need this any more.

nurse-the-code: Either fix this with the new logic or remove this method.

nurse-the-code: Need to fix this line:
`csvPrinter.print(castVoteRecord.getId());`
in `OutputWriter.java`

nurse-the-code: Move the code in here to a `getSuppliedId` method and what is "Record ID" in the RCTab CVR should be "Vendor Id"

src/main/java/network/brightspots/rcv/CastVoteRecord.java
```diff
@@ -138,13 +145,9 @@ void logRoundOutcome(
...
     logStringBuilder.append("[Round] ").append(round).append(" [CVR] ");
-    if (!isNullOrBlank(computedId)) {
-      logStringBuilder.append(computedId);
-    } else {
-      logStringBuilder.append(suppliedId);
-    }
+    logStringBuilder.append(getPrimaryId());
     if (outcomeType == VoteOutcomeType.IGNORED) {
-      logStringBuilder.append(" [was ignored] ");
+      logStringBuilder.append(" [ was ignored] ");
```
nurse-the-code, line 150: remove this change

src/main/java/network/brightspots/rcv/OutputWriter.java
```diff
@@ -696,8 +696,9 @@ String writeRcTabCvrCsv(
       CSVFormat format = CSVFormat.DEFAULT.builder().setNullString("").build();
       csvPrinter = new CSVPrinter(writer, format);
       // print header:
-      // ContestId, TabulatorId, BatchId, RecordId, Precinct, Precinct Portion, rank 1 selection,
-      // rank 2 selection, ... rank maxRanks selection
+      // RCTab CVR Id, ContestId, TabulatorId, BatchId, RecordId, Precinct, Precinct Portion,
+      // rank 1 selection, rank 2 selection, ... rank maxRanks selection
+      csvPrinter.print("RCTab CVR Id");
```
nurse-the-code, line 701: move this down a couple lines to be after the `"Contest Id"`

src/main/java/network/brightspots/rcv/OutputWriter.java
```diff
@@ -737,6 +738,7 @@ String writeRcTabCvrCsv(
         }
 
         CastVoteRecord castVoteRecord = castVoteRecords.get(i);
+        csvPrinter.print(castVoteRecord.getPrimaryId());
```
nurse-the-code, line 741: move this down a couple lines to be after the "Contest Id"

src/main/java/network/brightspots/rcv/OutputWriter.java
```diff
@@ -981,6 +984,7 @@ private Map<String, Object> generateCvrSnapshotMap(
             entry("@type", "CVR.CVRContest"));
 
     return Map.ofEntries(
+        // TODO: Do we want this to use the getSuppliedId or the getId method?
         entry("@id", generateCvrSnapshotId(sanitizeStringForOutput(cvr.getId()), round)),
```
nurse-the-code, line 988: Our choice here should be the same as in the `TODO` comment added before the `sanitizedId` definition.

src/main/java/network/brightspots/rcv/OutputWriter.java
```diff
@@ -875,6 +877,7 @@ private List<Map<String, Object>> generateCdfMapForCvrs(List<CastVoteRecord> cas
     for (CastVoteRecord cvr : castVoteRecords) {
       List<Map<String, Object>> cvrSnapshots = new LinkedList<>();
       cvrSnapshots.add(generateCvrSnapshotMap(cvr, null, null));
+      // TODO: Do we want this to use the getSuppliedId or the getId method?
```
nurse-the-code, line 880: In researching this question, I looked at the [NIST Cast Vote Records Common Data Format Specification v 1.0](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.1500-103.pdf) document. Here are some relevant screenshots I found.

- on page 18:
    Screenshot 2025-01-22 at 05 52 02
- on page 31:
    Screenshot 2025-01-22 at 05 53 46


This prints out the `BallotPrePrintedId` from the CDF's CVR class. It would seem to me that this should be a vendor-supplied id (e.g. `suppliedId`). I am not sure that we were using this correctly prior to this PR, because we would sometimes print the `computedId` (which was never pre-printed on any ballot). I am also not sure what this should look like if there is no vender supplied id.

src/test/resources/network/brightspots/rcv/test_data/conversions_from_ess/conversions_from_ess_expected.csv
```diff
@@ -1,11 +1,11 @@
...
+../_shared/simple_ess_cvr.xlsx,ess,,simple_ess_cvr.xlsx-1,,,,,,Mookie Blaylock,undervote,undervote
+../_shared/simple_ess_cvr.xlsx,ess,,simple_ess_cvr.xlsx-2,,,,,,Mookie Blaylock,undervote,undervote
+../_shared/simple_ess_cvr.xlsx,ess,,simple_ess_cvr.xlsx-3,,,,,,Mookie Blaylock,undervote,undervote
+../_shared/simple_ess_cvr.xlsx,ess,,simple_ess_cvr.xlsx-4,,,,,,Mookie Blaylock,undervote,undervote
+../_shared/simple_ess_cvr.xlsx,ess,,simple_ess_cvr.xlsx-5,,,,,,Yinka Dare,undervote,undervote
+../_shared/simple_ess_cvr.xlsx,ess,,simple_ess_cvr.xlsx-6,,,,,,Yinka Dare,undervote,undervote
+../_shared/simple_ess_cvr.xlsx,ess,,simple_ess_cvr.xlsx-7,,,,,,Yinka Dare,undervote,undervote
+../_shared/simple_ess_cvr.xlsx,ess,,simple_ess_cvr.xlsx-8,,,,,,George Gervin,undervote,undervote
+../_shared/simple_ess_cvr.xlsx,ess,,simple_ess_cvr.xlsx-9,,,,,,George Gervin,Yinka Dare,undervote
+../_shared/simple_ess_cvr.xlsx,ess,,simple_ess_cvr.xlsx-10,,,,,,Sedale Threatt,George Gervin,undervote
```
nurse-the-code, line 11: For ES&S CVRs, it looks like our new `getSuppliedId` method (slightly changed from the original `getId` method results in no `"Vendor Id"` (formerly `"Record Id"`) getting printed to the `rctab_cvr.csv`. Instead, that same value is printed as the `"RCTab CVR Id"` (using the new `getId` method).

This is because the old `getId` method (and the new `getId method as well) returns a `computedId` for ES&S CVRs. The new `getSuppliedId` returns an empty string, because ES&S CVRs don't have `suppliedId` associated with them in RCTab.

src/test/java/network/brightspots/rcv/CastVoteRecordTests.java
```diff
@@ -0,0 +1,196 @@
...
+      // Act
+      String expectId = defualtSuppliedId;
+      String actualId = cvr.getSuppliedId();
+
+      // Assert
+      assertEquals(expectId, actualId);
+    }
+
+    @Test
+    @DisplayName("returns empty string when suppliedId is null")
```
yezr, line 155: I could see us wanting to know when suppliedId is not....well, supplied :) When it is null, that is a flag that it was not supplied by the vendor. Can you see a reason for returning an empty string here when that happens instead? 

Does this also result in the change you mention [here](https://github.com/BrightSpots/rcv/pull/913/files#r1925212693)?

src/main/java/network/brightspots/rcv/OutputWriter.java
```diff
@@ -696,14 +696,15 @@ String writeRcTabCvrCsv(
       CSVFormat format = CSVFormat.DEFAULT.builder().setNullString("").build();
       csvPrinter = new CSVPrinter(writer, format);
       // print header:
-      // ContestId, TabulatorId, BatchId, RecordId, Precinct, Precinct Portion, rank 1 selection,
-      // rank 2 selection, ... rank maxRanks selection
+      // RCTab CVR Id, ContestId, TabulatorId, BatchId, RecordId, Precinct, Precinct Portion,
```
nurse-the-code, line 699: reorder the variables in this comment

src/test/java/network/brightspots/rcv/CastVoteRecordTests.java
nurse-the-code: Per this and our conversation yesterday, I changed it so that the `CastVoteRecord.getSuppliedId` method no longer returns an empty string when the `suppliedId` is null. Instead, the null handling is done where the `getSuppliedId` method is called.

src/main/java/network/brightspots/rcv/OutputWriter.java
```diff
@@ -744,7 +745,7 @@ String writeRcTabCvrCsv(
         csvPrinter.print(castVoteRecord.getId());
         csvPrinter.print(castVoteRecord.getTabulatorId());
         csvPrinter.print(castVoteRecord.getSlice(TabulateBySlice.BATCH));
-        csvPrinter.print(castVoteRecord.getSuppliedId());
+        csvPrinter.print(Objects.toString(castVoteRecord.getSuppliedId(), ""));
```
yezr, line 748: just confirming: is this safe for when `suppliedId` = `null` ?

nurse-the-code: https://learn.microsoft.com/en-us/dotnet/api/java.util.objects.tostring?view=net-android-34.0

The second parameter in `Objects.toString` is a `null` parameter.

src/main/java/network/brightspots/rcv/DominionCvrReader.java
```diff
@@ -321,7 +321,7 @@ private int parseCvrFile(
       String computedId =
           Stream.of(tabulatorId, batchId, Integer.toString(recordId))
               .filter(s -> s != null && !s.isBlank())
-              .collect(Collectors.joining("|"));
+              .collect(Collectors.joining("-")); // using a dash since this is an Id
```
artoonie, line 324: Not sure I understand why - vs | is important, but the comment makes me think it matters? Is it because it will be used in filenames? If so, perhaps that should be explicit.

src/test/java/network/brightspots/rcv/CastVoteRecordTests.java
```diff
@@ -0,0 +1,196 @@
...
+              rankings
+      );
+
+      // Act
+      String expectId = idWithSpecialCharacters;
+      String actualId = cvr.getSuppliedId();
+
+      // Assert
+      assertEquals(expectId, actualId);
+    }
```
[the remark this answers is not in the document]
yezr: agreed. To test that should we edit any of our current tests to include some data with special characters then confirm they are correct in the output? 

And trying to understand what is "correct", I see this in the output writer
```
static String sanitizeStringForOutput(String s) {
    return s == null ? "" : s.replaceAll("[^a-zA-Z0-9_\\-.]", "_");
  }
 ```

Which says anything not one of the following is replaced with `_`

- a letter
- a number
- an underscore
- a hyphen
- a period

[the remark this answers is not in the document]
Comment (artoonie) [reply src/test/java/network/brightspots/rcv/CastVoteRecordTests.java:194]: Agreed, this could be an easy enough change, though I fear it will get lost in the tests if we just randomly add a comma to a candidate name in one of the existing tests.

Something I do in RCVis that I think could work nicely here: create a function which takes an existing test, changes it just a touch, then tests something specific: https://github.com/artoonie/rcvis/blob/da0a7ee7fbc96e76dd11305ccc49bc69da2d36b4/common/testUtils.py#L117

Here, that could be used to take an existing test in the _shared directory, modify a candidate name to add a comma, then instead of checking an exact expected_* file, we test more explicitly that the CSVs have the candidate name properly escaped.

Anyway, we can punt that to another issue?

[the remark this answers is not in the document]
Comment (yezr) [reply src/test/java/network/brightspots/rcv/CastVoteRecordTests.java:194]: created #922 and prioritized for after 2.0

File: src/main/java/network/brightspots/rcv/DominionCvrReader.java (hunk 12 above)
Comment (yezr) [reply src/main/java/network/brightspots/rcv/DominionCvrReader.java:324]: It's an aesthetic choice. Originally I had used pipes as they aren't really used in the wild (certainly less so than `-`), and I wanted to signal that it was computed. I'm going to revert to pipe to signal that.

File: src/main/java/network/brightspots/rcv/OutputWriter.java [hunk 14]
```diff
@@ -696,17 +697,19 @@ String writeRcTabCvrCsv(
...
+      // RCTab CVR Id, ContestId, TabulatorId, BatchId, RecordId, Precinct, Precinct Portion,
+      csvPrinter.print("Source Filepath"); // CvrSource.getFilePath()
+      csvPrinter.print("CVR Provider"); // CvrSource.getProvider()
+      csvPrinter.print("Contest Id"); // CastVoteRecord.getContestId()
+      csvPrinter.print("RCTab CVR Id"); // CastVoteRecord.getId()
+      csvPrinter.print("Tabulator Id"); // CastVoteRecord.getTabulatorId
+      csvPrinter.print("Batch Id"); // CastVoteRecord.getSlice(TabulateBySlice.BATCH)
+      csvPrinter.print("Vendor Id"); // CastVoteRecord.getSuppliedId
+      // CastVoteRecord.getSlice(ContestConfig.TabulateBySlice.PRECINCT)
       csvPrinter.print("Precinct");
```
artoonie, line 709: Probably can make this line match the rest on one line (and therefore make it more readable) by using
`csvPrinter.print("Precinct"); // CastVoteRecord.getSlice(TabulateBySlice.PRECINCT)`, or if not, just `(PRECINCT)`

src/main/java/network/brightspots/rcv/DominionCvrReader.java
artoonie: I believe we discovered another reason why pipes won't work, but I forget what it was. Can we change the comment to be the reason you discovered?

artoonie (approved):
Some [nit]s but this looks good

artoonie (approved):

PR #919, opened 2025-02-14
Hart Auto-Load

### LINKED ISSUE #918: Hart: Auto load candidate name and alias correctly
**RCTab Version**: 2.0
**Contest**: Hart Cedar Park School Board in repo

## Current Behavior
This is similar to #866 with Dominion contests but for Hart configured contests. When I auto-load from an empty candidate list I get only candidate codes. Those codes are incorrectly loaded into the `Name` column, not the `Aliases/Codes` column. 

Image

## Expected Behavior
Candidate names and aliases are loaded successfully into their respective places. 
Image

yezr:
fixes #918. Cribs a lot of the logic from #908


yezr:
The logic works, but I would love a review of the code syntax.

src/main/java/network/brightspots/rcv/HartCvrReader.java
```diff
@@ -109,10 +117,31 @@ private void readCastVoteRecord(List<CastVoteRecord> castVoteRecords, Path path)
...
-            String candidateName = option.Id;
-            if (candidateName.equals(source.getUndeclaredWriteInLabel())) {
-              candidateName = Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL;
+            String candidateCode = option.Id;
+            String candidateName = option.Name;
+            if (candidateCode.equals(source.getUndeclaredWriteInLabel())) {
+              candidateCode = Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL;
+            }
+            if (this.candidateCodesToCandidates.containsKey(candidateCode)) {
+              if (!this.candidateCodesToCandidates.get(candidateCode).Name.equals(candidateName)) {
```
artoonie, line 126: These two ifs can be simplified a bit:

```
Candidate existingCandidate = this.candidateCodesToCandidates.get(candidateCode);
if (existingCandidate != null && existingCandidate.Name.equals(candidateName)) {
```

src/main/java/network/brightspots/rcv/HartCvrReader.java
```diff
@@ -109,10 +117,31 @@ private void readCastVoteRecord(List<CastVoteRecord> castVoteRecords, Path path)
...
+                String message = """
+                        Candidate Code %s associated with more than one candidate name.
+                        Originally associated with name '%s'.
+                        In CVR at '%s' it is associated with '%s'.
+                        """.formatted(candidateCode,
```
artoonie, line 131: This syntax will cause the error message to have a ton of spaces throughout the error message. Using `+`s will avoid that:
```
                String message =
                        "Candidate Code %s associated with more than one candidate name.  " +
                        "Originally associated with name '%s'. " + 
                        "In CVR at '%s' it is associated with '%s'."
                        .formatted(candidateCode,
```

src/main/java/network/brightspots/rcv/HartCvrReader.java
```diff
@@ -109,10 +117,31 @@ private void readCastVoteRecord(List<CastVoteRecord> castVoteRecords, Path path)
...
+                        this.candidateCodesToCandidates.get(candidateCode).Name,
+                        path.getFileName(), candidateName);
+                Logger.severe(message);
+                throw new CastVoteRecord.CvrParseException();
+              }
+            } else {
+              if (!candidateCode.equals(Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL)) {
```
artoonie, line 138: Combine into `else if {` rather than `else { if {`

src/main/java/network/brightspots/rcv/HartCvrReader.java
```diff
@@ -109,10 +117,31 @@ private void readCastVoteRecord(List<CastVoteRecord> castVoteRecords, Path path)
...
+                this.candidateCodesToCandidates.put(candidateCode,
+                        new Candidate(candidateName, candidateCode));
+              }
```
artoonie, line 141: I think you can simplify this whole block using `computeIfAbsent`:

```
          for (Option option : contest.Options) {
            final String candidateCode = option.Id;
            final String candidateName = option.Name;
            final boolean isUndeclaredWriteIn = candidateCode.equals(
                    source.getUndeclaredWriteInLabel());

            if (!isUndeclaredWriteIn) {
              this.candidateCodesToCandidates.computeIfAbsent(candidateCode, k ->
                      new Candidate(candidateName, candidateCode));

              String candidateNameFromMap = this.candidateCodesToCandidates.get(candidateCode).Name;
              if (!candidateNameFromMap.equals(candidateName)) {
                String message = String.format(
                        "Candidate Code %s associated with more than one candidate name. "
                                + "Originally associated with name '%s'. "
                                + "In CVR at '%s' it is associated with '%s'.",
                        candidateCode, candidateNameFromMap, path.getFileName(), candidateName
                );
                Logger.severe(message);
                throw new CastVoteRecord.CvrParseException();
              }
            }

            // Hart RCV election ranks are indicated by a string read left to right:
            // each digit corresponds to a rank and is set to 1 if that rank was voted:
            // 0100 indicates rank 2 was voted
            // 0000 indicates no rank was voted (undervote)
            // 0101 indicates ranks 2 and 4 are voted (overvote)
            for (int rank = 1; rank < option.Value.length() + 1; rank++) {
              String rankValue = option.Value.substring(rank - 1, rank);
              if (rankValue.equals("1")) {
                if (isUndeclaredWriteIn) {
                  rankings.add(new Pair<>(rank, UNDECLARED_WRITE_IN_OUTPUT_LABEL));
                } else {
                  rankings.add(new Pair<>(rank, candidateCode));
                }
              }
            }
          }
```

src/main/java/network/brightspots/rcv/HartCvrReader.java
```diff
@@ -147,6 +176,31 @@ private void readCastVoteRecord(List<CastVoteRecord> castVoteRecords, Path path)
...
+
+    Set<String> knownNames = config.getCandidateNames();
+    if (this.candidateCodesToCandidates.entrySet().isEmpty()) {
+      try {
+        //Reading the CVRs will load this.candidateCodestoCandidates
+        readCastVoteRecords(castVoteRecords);
+      } catch (CastVoteRecord.CvrParseException e) {
+        Logger.severe("Error gathering Unknown Candidates\n%s", e);
+        return new HashSet<>();
+      } catch (IOException e) {
```
artoonie, line 191: You can collapse the two catches into one:
```
      } catch (CastVoteRecord.CvrParseException | IOException e) {
```

src/main/java/network/brightspots/rcv/HartCvrReader.java
```diff
@@ -147,6 +176,31 @@ private void readCastVoteRecord(List<CastVoteRecord> castVoteRecords, Path path)
...
   }
 
+  @Override
+  public Set<RawContestConfig.Candidate> gatherUnknownCandidates(
+          List<CastVoteRecord> castVoteRecords) {
+
+    Set<String> knownNames = config.getCandidateNames();
+    if (this.candidateCodesToCandidates.entrySet().isEmpty()) {
+      try {
+        //Reading the CVRs will load this.candidateCodestoCandidates
```
artoonie, line 186:
```suggestion
        // Reading the CVRs will load this.candidateCodestoCandidates
```

src/main/java/network/brightspots/rcv/HartCvrReader.java
yezr: great heads up. I updated to use this to put the candidate code/name into the lookup if the code isn't there. then use the lookup to compare against the current code/name to see if the current CVR's candidate code/name matches the lookup. if not, it throws.

src/main/java/network/brightspots/rcv/HartCvrReader.java
yezr: along with the computeIfAbsent update this should be addressed

yezr:
@artoonie I addressed all your comments. I closed all the ones that I don't think need further discussion. There are two open that implement the map.ComputeIfAbsent you suggested that I'm interested in your feedback on

src/main/java/network/brightspots/rcv/HartCvrReader.java
artoonie: This looks good to me

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -1905,8 +1905,13 @@ protected LoadedCvrData call() {
               cvrStatics = session.parseAndCountCastVoteRecords(this::updateProgress);
               succeeded();
             } catch (TabulatorSession.CastVoteRecordGenericParseException e) {
-              Logger.severe("Failed to read CVRs: %s", e.getMessage());
+              String message = "Failed to read CVRs";
+              if (!isNullOrBlank(e.getMessage())) {
+                message += ": " + e.getMessage();
+              }
+              Logger.severe(message);
```
artoonie, line 1912: If getting an NPE here, I believe this could be simplified with:
`Logger.severe("Failed to read CVRs: " + e.getMessage());`
which would just say "Failed to read CVRs: null" if null. (And that's easier to include in the list of errors that we need for the TDP)

src/main/java/network/brightspots/rcv/HartCvrReader.java
```diff
@@ -109,10 +118,31 @@ private void readCastVoteRecord(List<CastVoteRecord> castVoteRecords, Path path)
         ArrayList<Pair<Integer, String>> rankings = new ArrayList<>();
         if (contest.Options != null) {
           for (Option option : contest.Options) {
-            String candidateName = option.Id;
-            if (candidateName.equals(source.getUndeclaredWriteInLabel())) {
-              candidateName = Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL;
+
+            //Can be null with some write-ins
+            String candidateNameClean = Objects.requireNonNullElse(option.Name, "");
```
artoonie, line 123: How far can we get without setting this to empty string? If it's null, I think it makes more sense to keep it null as long as possible -- unless that causes a ton of issues everywhere. Another idea is to set the null candidate name to the ID instead.

src/main/java/network/brightspots/rcv/HartCvrReader.java
```diff
@@ -109,10 +118,31 @@ private void readCastVoteRecord(List<CastVoteRecord> castVoteRecords, Path path)
...
+            Candidate candidate = new Candidate(candidateNameClean, option.Id);
+            if (candidate.Code.equals(source.getUndeclaredWriteInLabel())) {
+              candidate.Code = Tabulator.UNDECLARED_WRITE_IN_OUTPUT_LABEL;
+            } else {
+              this.candidateCodesToCandidates.putIfAbsent(candidate.Code,
+                      candidate);
```
artoonie, line 129: [nit] I think these two lines would fit on one line

src/main/java/network/brightspots/rcv/HartCvrReader.java
```diff
@@ -147,6 +177,28 @@ private void readCastVoteRecord(List<CastVoteRecord> castVoteRecords, Path path)
...
+      }
+    }
+
+    // Return the candidate codes that are not in the knownNames set
+    return candidateCodesToCandidates.entrySet().stream()
+            .filter(entry -> !knownNames.contains(entry.getValue().Name))
+            .map(entry -> new RawContestConfig.Candidate(entry.getValue().Name, entry.getKey()))
+            .collect(Collectors.toSet());
+  }
+
```
artoonie, line 201: This code seems useful to move to `BaseCvrReader` to replace `gatherUnknownCandidates` there -- any downsides to that I don't see?

src/main/java/network/brightspots/rcv/GuiConfigController.java
yezr: I was thinking that any place where we show `null` explicitly to an end user could be alarming and I was trying to avoid that. 

I didn't even think about the list of errors reporting requirements. But I think here either way we are going to have to search down the different `e.getMessage()` potential descriptions. The only difference with this change would be that the case where there is no explicit message is a little bit different. I think it's worth it to remove the `null` reference.

src/main/java/network/brightspots/rcv/HartCvrReader.java
yezr: Started digging in to this. When trying to move it to `BaseCvrReader` it looks like it would need to know about the `candidateCodesToCandidates` map. The reason is because that map is lazy loaded. If we get here and it is empty we take a pass through the CVRs to populate it. Once it's populated we use it for the comparison to see which haven't been configured.

artoonie: Looked into this and agree -- not worth changing. The two Candidate classes are different, so any consolidation would be a bigger lift. Keep as is!

artoonie (approved):

Issue #914, opened 2025-01-09
Output Folder Structure Cleanup


yezr:
The current RCTab folder setup can be unintuitive for some users. We’ve added hashes and extra tabulate by slice options that can clutter the output folder. This ticket proposes a redesign to

- Make it simple and easy for all users to see directly the round by round results first, with the option to navigate to more info (logs, hashes etc.) second
- Distinguish more clearly .hash files from regular files
- Distinguish more clearly separate tabulations of the same contest

<img width="848" alt="image" src="https://github.com/user-attachments/assets/3884f4fd-a3dc-43da-86d3-ec179eb7560f" />

Some notes
- The top level output folder is still the one that is configured in RCTab. 
- Each run creates a new `YYYY-MM-DD_HH-mm Results` folder as a child to the parent `output` folder. Though unlikely, we must protect against collisions if more than one tabulation is run on the same minute. We've implemented appending an auto-incrementing number starting at 1 for collisions.
- `Open Results Folder` button in RCTab after running tabulation should show the contents of the `YYYY-MM-DD-HH-mm Results` folder for that run
- Open to changing checksums folder name/location if it would be easier in code to standardize it e.g. making any hash go into a `Checksums` folder in the current location

artoonie:
Thoughts on the root level checksums directory being `YYYY-MM-DD_HH-mm Results Checksums` instead of `Checksums`? That creates some consistency: all `Checksums` directories are `[parent dir name] Checksums`. Easier to code for one, as we don't need a special case, but I think also easier to parse for anyone automatically checking checksum code.

Also: thoughts on removing the space between the timestamp and "Results" in the root directory? Avoiding spaces in directory names is no longer necessary, but is still nice-to-have.

yezr:
> Thoughts on the root level checksums directory being YYYY-MM-DD_HH-mm Results Checksums instead of Checksums? That creates some consistency: all Checksums directories are [parent dir name] Checksums. Easier to code for one, as we don't need a special case, but I think also easier to parse for anyone automatically checking checksum code.

So the name of each checksum folder has the entire parent folder name in it? Sounds good to me

> Also: thoughts on removing the space between the timestamp and "Results" in the root directory? Avoiding spaces in directory names is no longer necessary, but is still nice-to-have.

The timestamps use both `-` and `_` to delineate it's parts. For me, it is nice to have something different, the space, to delineate the timestamp and the folder description.

PR #920, opened 2025-02-20
Feature/issue 914 output folder

artoonie:
Here's what this looks like:
image

Some additional discussion in #914 about design decisions here


yezr:
I get errors when I run the same tabulation during the same minute. 
image

An idea for handling collisions is in #914 
> Each run creates a new YYYY-MM-DD_HH-mm Results folder as a child to the parent output folder. Though unlikely, we must protect against collisions if more than one tabulation is run on the same minute. A solution could be appending _run2, _run3 to the tabulation folder name for subsequent runs within the same minute.

artoonie:
Fixed -- just added a dash-number for same-second collisions

yezr:
LGTM! confirmed it handles same minute naming collisions

yezr (approved):
LGTM!

PR #925, opened 2025-03-14
threshold for Continue Until 2 freezes after elected

### LINKED ISSUE #924: Bug: fix "Final threshold", and reconsider "Current round threshold"
When continuing tabulation down to two in IRV, the "Final threshold" should be reflective of the threshold at the time when the winner was elected, not the final round.

Additionally, we should rethink whether "Current round threshold" is based on the current round until a candidate is elected, then "freezes" on that number as tabulation continues, or whether it continues to change after someone is elected. I lean towards freezing it in place.

artoonie:
Closes #924 -- the threshold should not continue to change after a winner is elected in "Continue Until Two" contests. The final threshold and the current round threshold should "freeze" on the round in which someone hits the threshold.

To verify that there is not an off-by-one on the round in which the freezing occurs, I have added a test where the threshold changes on the round before and after the candidate gets elected.


yezr:
LGTM. Ready for code review!

yezr (approved):

PR #877, opened 2024-08-15
Upgrade to Gradle 8.10 and JDK 21

alyssahursh:
#### Motivation
* JDK 20 is a feature release and is not recommended for use in production: https://www.oracle.com/java/technologies/javase/jdk20-archive-downloads.html
* JDK 21 is the latest long-term support (LTS) release of the JDK: https://www.oracle.com/java/technologies/downloads/#java21
* Gradle 8.4+ is required for JDK 21

#### Commits
This PR contains two atomic commits, each of which build and run successfully
* Upgrade to Gradle 8.10 (automated result of running `gradle wrapper --gradle-version 8.10`)
* Upgrade to JDK 21

#### Testing
* `./gradlew build` and `./gradlew run` both succeed

#### Note
After accepting this commit, you will need to install JDK 21 and update JAVA_HOME to point to the file system location where JDK 21 is installed

PR #722, opened 2023-06-26
Upgrade deps

HEdingfield:
Upgrades:
* JDK to 20.0.1
* Gradle to 8.1.1
* spotbugs-gradle-plugin to 5.0.14
* spotbugs to 4.7.3
* checkstyle to 10.12.1
* jlink to 2.26.0
* commons.csv to 1.10.0
* poi-ooxml to 5.2.3
* jackson to 2.15.2
* jupiter to 5.9.3
* javafxplugin to 0.0.14
* javafx to 20.0.1

Closes #669.

PR #931, opened 2025-04-29
Feature/issue 669 update all deps

### LINKED ISSUE #669: Update JDK and all other deps for v1.4.0

artoonie:
Note: this PR is based on #877 and the comments there apply. Closes #669.

The commits here:
1. Update plugins
2. Update code for latest spotbugs
3. Merge develop
4. Final fixes
4. Additional changes during Code Review

There is a warning about a deprecation when compiling now. We'll have to live with it -- it's caused by `org.javamodularity:moduleplugin:1.8.12` which is a transitive dependency of `org.openjfx.javafxplugin:org.openjfx.javafxplugin.gradle.plugin:0.0.14` which we cannot upgrade without moving to JDK 22, which is not LTS. 

I also had to change `org.junit.jupiter:junit-jupiter-engine` from `testRuntime` to `testImplementation`, not fully sure why, something about java or gradle or something being stricter? Would love a second opinion on the consequences of that.


src/main/java/network/brightspots/rcv/SecurityConfig.java
```diff
@@ -19,15 +19,15 @@
...
 
-  // Is the user allowed to save output files to their Users directory?
-  // Since user accounts retain delete and create permissions to their user account folders,
-  // this should be disallowed to truly ensure output files are read-only.
-  private static boolean CAN_OUTPUT_FILES_SAVE_TO_USERS_DIRECTORY = true;
+  // ThreadLocal variables for thread-safe overrides
+  private static final ThreadLocal<Boolean> hartSignatureValidationEnabled =
+          ThreadLocal.withInitial(() -> DEFAULT_HART_SIGNATURE_VALIDATION_ENABLED);
+  private static final ThreadLocal<Boolean> canOutputFilesSaveToUsersDirectory =
+          ThreadLocal.withInitial(() -> DEFAULT_CAN_OUTPUT_FILES_SAVE_TO_USERS_DIRECTORY);
```
yezr, line 30: Is this required with the new Java version? Can we keep the old comment text? I find it helpful to explain what exactly these configuration options are doing.

src/test/java/network/brightspots/rcv/ContestConfigTests.java
```diff
@@ -58,9 +58,9 @@ void testVersionComparison() {
     // jackson-core to address.
     // https://github.com/FasterXML/jackson-core/issues/1050
     // Once that issue is addressed, the first and third tests below should flip from False to True
-    assertFalse(isVersionNewer("1.4.0-beta", "1.4.0-alpha"));
```
yezr, line 61: awesome so it looks like the issue referenced is fixed in the new version? Can we remove the comment above if that is the case

yezr (commented):
some questions and suggestions

gradlew
```diff
@@ -55,7 +57,7 @@
 #       Darwin, MinGW, and NonStop.
 #
 #   (3) This script is generated from the Groovy template
-#       https://github.com/gradle/gradle/blob/HEAD/subprojects/plugins/src/main/resources/org/gradle/api/internal/plugins/unixStartScript.txt
```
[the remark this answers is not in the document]
artoonie: Yep, this file auto-updates

build.gradle
```diff
@@ -16,22 +16,25 @@ repositories {
...
+
+    testImplementation platform('org.junit:junit-bom:5.13.0-M2')
+    testImplementation 'org.junit.jupiter:junit-jupiter-api'
+    testImplementation 'org.junit.jupiter:junit-jupiter-engine'
+    testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
 }
 
 // ### Application plugin settings
 application {
-    mainModule = "${moduleName}"
```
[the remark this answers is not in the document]
artoonie: The new Java version doesn't allow dynamic naming. `mainModule = moduleName` works, but not the next line, so I made them both explicit.

.github/workflows/release.yml
```diff
@@ -76,10 +76,10 @@ jobs:
 
       - uses: actions/checkout@v3
 
-      - name: "Set up JDK 20.0.1"
+      - name: "Set up JDK 21.0.4"
```
HEdingfield, line 79: Here and elsewhere: can we bump up to 21.0.7 now, as it seems to be the latest GA?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-# Also manually update gradlew and gradlew.bat with appropriate version tag from:
-# https://github.com/gradle/gradle
```
HEdingfield, line 4: Why get rid of this comment?

src/main/java/network/brightspots/rcv/SecurityConfig.java
HEdingfield, line 1: Are the changes made here related to the dep upgrades, or were they opportunistic? If the latter, consider splitting into a separate PR.

Not a hard requirement if you think it's better to include here.

build.gradle
```diff
@@ -2,11 +2,11 @@ plugins {
...
     // Latest version at: https://github.com/spotbugs/spotbugs-gradle-plugin/tags
-    id "com.github.spotbugs" version "5.0.14"
+    id "com.github.spotbugs" version "6.1.10"
     id "idea"
     id "java-library"
     // Latest version at: https://github.com/beryx/badass-jlink-plugin/tags
-    id "org.beryx.jlink" version "2.26.0"
+    id "org.beryx.jlink" version "3.1.1"
     // Latest version at: https://github.com/openjfx/javafx-gradle-plugin/tags
     id "org.openjfx.javafxplugin" version "0.0.14"
```
HEdingfield, line 11: 0.1.0 looks newer here, why not upgrade?

README.md
```diff
@@ -32,7 +32,7 @@ On Linux, you may install the .deb file, then run `/opt/rcv/bin/RCTab` to launch
 
 #### Method 2 (Less Easy): Compile and Run Using Gradle
 
-1. Install [JDK 20](https://www.oracle.com/java/technologies/javase/jdk20-archive-downloads.html), and make sure your Java path is picking it up properly by
+1. Install [JDK 21](https://adoptium.net/temurin/releases/?version=21), and make sure your Java path is picking it up properly by
```
HEdingfield, line 35: Why this site instead of https://www.oracle.com/java/technologies/javase/jdk21-archive-downloads.html ?

build.gradle
```diff
@@ -16,22 +16,25 @@ repositories {
...
+    implementation 'commons-cli:commons-cli:1.9.0'
+    implementation 'org.apache.commons:commons-csv:1.14.0'
+    implementation 'org.apache.poi:poi-ooxml:5.4.1'
+    implementation 'com.fasterxml.jackson.core:jackson-core:2.19.0'
+    implementation 'com.fasterxml.jackson.core:jackson-annotations:2.19.0'
+    implementation 'com.fasterxml.jackson.core:jackson-databind:2.19.0'
+    implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-xml:2.19.0'
+    implementation "com.github.spotbugs:spotbugs:4.9.3"
+
+    testImplementation platform('org.junit:junit-bom:5.13.0-M2')
```
HEdingfield, line 28: I would stick with 5.12.2 for Jupiter to avoid `-M*`. See if that allows you to keep the old format we've been using consistently in this file for a few years.

build.gradle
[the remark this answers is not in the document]
HEdingfield: I remember fighting with this for a long time years ago, but can't remember exactly why I had to get it this way. It might have had to do with `checkstyle`.

Speaking of, it seems you forgot to update that in this PR. See #722 for an example of that and please include it here.

HEdingfield (changes requested):

src/main/java/network/brightspots/rcv/SecurityConfig.java
artoonie: Required for the latest version of spotbugs -- the old code was not thread-safe, and spotbugs now complains about it.

build.gradle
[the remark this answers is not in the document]
artoonie: Working on it. Just to confirm -- the old google_checks is the default, right? No modifications I need to copy over?

There are around 200 issues, many of them pretty annoying to try to handle (like adding javadoc for getters and setters, and having to suppress PascalCase classes used for parsing XMLs).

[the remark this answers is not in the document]
HEdingfield: > the old google_checks is the default, right? No modifications I need to copy over?

I'm not exactly sure what you mean by this. There's a comment in `build.gradle` that has instructions -- it should start with:

> Keep the below file updated along with subsequent versions of Checkstyle

If you're saying you updated it and now it's blocking you with a lot of new linting issues, I'd be ok with a fast follow PR to silence those.

build.gradle
```diff
@@ -16,22 +16,25 @@ repositories {
...
+    implementation 'com.fasterxml.jackson.core:jackson-core:2.19.0'
+    implementation 'com.fasterxml.jackson.core:jackson-annotations:2.19.0'
+    implementation 'com.fasterxml.jackson.core:jackson-databind:2.19.0'
+    implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-xml:2.19.0'
+    implementation "com.github.spotbugs:spotbugs:4.9.3"
+
+    testImplementation platform('org.junit:junit-bom:5.12.2')
+    testImplementation 'org.junit.jupiter:junit-jupiter-api'
+    testImplementation 'org.junit.jupiter:junit-jupiter-engine'
+    testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
```
HEdingfield, line 31: Did you confirm that these four lines are still necessary when swapping to 5.12.2?

Can you not just do:

```
    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.12.2'
    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.12.2'
```

Like we did before?

README.md
yezr: Doing some research on the Oracle JDK 21 long term release. It is free for a while, then it costs according to their [download site](https://www.oracle.com/java/technologies/downloads/#java21:~:text=JDK%2021%20will%20receive%20updates%20under%20the%20NFTC%2C%20until%20September%202026%2C%20a%20year%20after%20the%20release%20of%20the%20next%20LTS.%20Subsequent%20JDK%2021%20updates%20will%20be%20licensed%20under%20the%20Java%20SE%20OTN%20License%20(OTN)%20and%20production%20use%20beyond%20the%20limited%20free%20grants%20of%20the%20OTN%20license%20will%20require%20a%20fee.). The Temurin version is free always according to their site. For that reason I think we should stick with Temurin.

HEdingfield: Ah, good catch. I also see it was referenced [here](https://github.com/BrightSpots/rcv/pull/877#issuecomment-2292282128). [This](https://www.reddit.com/r/java/comments/1b7yp39/comment/ktlqrd5/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) is also a good rundown.

Swapping it back to Adoptium SGTM.

build.gradle
yezr: Reverting back to the old way I was able to build locally. Is that enough to test this? Or do I need to check it in and make a build, then put on a separate machine? I was going to do that eventually for testing everything.

HEdingfield: I think doing so locally is sufficient, because this is just for running tests anyway. Sounds good though, let's set this back to the way it was.

build.gradle
```diff
@@ -47,23 +47,20 @@ run {
...
-    // "${org.checkstyle.google.suppressionfilter.config}"
-    // with:
-    // "${config_loc}/checkstyle-suppressions.xml"
-    // Additional context: https://github.com/gradle/gradle/issues/21409
+    // You must then modify this file to add the following under MissingJavadocMethod:
+    // <property name="minLineCount" value="2"/>
     configFile = file("$projectDir/config/checkstyle/google_checks.xml")
+    // Additional contex for configPropertiest: https://github.com/gradle/gradle/issues/21409
+    configProperties = [
+        "org.checkstyle.google.suppressionfilter.config": file("$projectDir/config/checkstyle/checkstyle-suppressions.xml").absolutePath
```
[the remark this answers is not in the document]
HEdingfield: @artoonie

build.gradle
yezr: Turns out that these four lines are needed. The `test` gradle task fails, unable to run anything with the previous code. These new lines fix that issue.

build.gradle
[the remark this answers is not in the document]
artoonie: Done -- I generally prefer to avoid manually editing files that we want to regularly update. It keeps diffs clean, and I saw no downside to this since we don't use a remote cache. But no strong feelings either way, so I've reverted to the old way.

build.gradle
```diff
@@ -44,27 +47,32 @@ run {
 // ### Checkstyle plugin settings
 checkstyle {
     // Latest version at https://github.com/checkstyle/checkstyle/tags
-    toolVersion = '10.12.1'
+    toolVersion = '10.23.1'
+
     // Keep the below file updated along with subsequent versions of Checkstyle (make sure to choose
     // the tag matching the toolVersion above):
     // https://github.com/checkstyle/checkstyle/blob/checkstyle-10.3.2/src/main/resources/google_checks.xml
+
```
HEdingfield, line 55: Nit: prefer not to add this line break, unless you // it. I think it's easier for someone to overlook the instructions below it unless it's all merged together, or the line breaks are clearly indicating a continuation of the comment.

build.gradle
```diff
@@ -44,27 +47,32 @@ run {
...
     // NOTE: After updating the local version of config/checkstyle/google_checks.xml again in the
-    // future, it may be necessary to subsequently replace the following line in the local copy to
-    // prevent suppressions from being ignored:
+    // future, you will need to make two manual changes:
+    //
+    // 1. Allow relative path for suppressions
```
HEdingfield, line 59: Nit: 1. Replace the following line in the local copy to prevent suppressions from being ignored:

build.gradle
```diff
@@ -44,27 +47,32 @@ run {
...
     // "${org.checkstyle.google.suppressionfilter.config}"
     // with:
     // "${config_loc}/checkstyle-suppressions.xml"
     // Additional context: https://github.com/gradle/gradle/issues/21409
+
+    // 2. Allow omitting javadoc on short functions
+    // Add the following property into the MissingJavadocMethod module:
```
HEdingfield, line 66: Nit: 2. Add the following property into the MissingJavadocMethod module to allow omitting Javadoc on short functions:

build.gradle
```diff
@@ -44,27 +47,32 @@ run {
...
-    // future, it may be necessary to subsequently replace the following line in the local copy to
-    // prevent suppressions from being ignored:
+    // future, you will need to make two manual changes:
+    //
+    // 1. Allow relative path for suppressions
     // "${org.checkstyle.google.suppressionfilter.config}"
     // with:
     // "${config_loc}/checkstyle-suppressions.xml"
     // Additional context: https://github.com/gradle/gradle/issues/21409
+
```
HEdingfield, line 64: Nit: same here, please either add // or eliminate the line break it entirely.

build.gradle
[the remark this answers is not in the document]
HEdingfield: But the "new" method also requires doing this, doesn't it?

> You must then modify this file to add the following under MissingJavadocMethod:

I guess it's one less line to modify, but with that potential caching downside.

HEdingfield (dismissed):

yezr (approved):
LGTM! all PR review comments have been resolved

yezr:
@HEdingfield can you approve again after I made the final commit? It was a minor one to fix a comment nit that had the unfortunate effect of dismissing your approval

HEdingfield (approved):

Issue #944, opened 2025-08-18
Bottoms Up Percentage Threshold: be clear about how decimal places for vote arithmetic affects threshold


yezr:
## What Happens

`Decimal Places For Vote Arithmetic (Multi-Winner Only)` is being used in the threshold calculation for Bottoms Up Percentage Threshold.

In ContestConfig.java, the getMultiSeatBottomsUpPercentageThreshold() method is 
being used to convert an integer percentage (like 15) into its decimal representation (0.15).

```
        //src/main/java/network/brightspots/rcv/ContestConfig.java:1010
        BigDecimal getMultiSeatBottomsUpPercentageThreshold() {
            return getMultiSeatBottomsUpPercentageThresholdRaw() != null
                && !getMultiSeatBottomsUpPercentageThresholdRaw().isBlank()
                ? divide(new BigDecimal(getMultiSeatBottomsUpPercentageThresholdRaw()), new BigDecimal(100))
                : null;
        }
```

Instead of performing a standard division, this method calls a custom divide() helper method within the same class.
```
        // src/main/java/network/brightspots/rcv/ContestConfig.java:1045
        BigDecimal divide(BigDecimal dividend, BigDecimal divisor) {
            return dividend.divide(divisor, getDecimalPlacesForVoteArithmetic(), RoundingMode.DOWN);
        }
```

It rounds the result of the division down to the number of decimal places specified in decimalPlacesForVoteArithmetic. If the `multiSeatBottomsUpPercentageThreshold` is `15` and `decimalPlacesForVoteArithmetic` is `1`, the calculation becomes:

`new BigDecimal(15).divide(new BigDecimal(100), 1, RoundingMode.DOWN)`

0.15 is rounded down to 1 decimal place, resulting in 0.1. This incorrectly changes the intended 15% threshold to 10%.

This percentage is then used in Tabulator.java to calculate the winning threshold:

    // src/main/java/network/brightspots/rcv/Tabulator.java:424
    if (config.isMultiSeatBottomsUpWithThresholdEnabled()) {
      winningThreshold =
          currentRoundTotalVotes.multiply(config.getMultiSeatBottomsUpPercentageThreshold());
    } 

## What Should Happen
Assume that if someone puts in `15%` that they mean `.15`, if they put in `15.0001` they mean `.150001` in the arithmetic. In `getMultiSeatBottomsUpPercentageThreshold()` use however many decimal places are needed for every digit configured in `Percentage Threshold`

[...]

Issue #946, opened 2025-08-19
Bottoms Up Percentage Threshold: Defaulting to Static


yezr:
## What Happens Now
When using Bottoms Up Percentage Threshold as your winning method, the `First Round Determines Threshold` setting is unchecked and greyed out. That makes Bottoms Up Percentage Threshold use a dynamic X% of active ballots as the threshold in each round. 

<img width="853" height="352" alt="Image" src="https://github.com/user-attachments/assets/12b24b9d-d978-490c-ae37-a9856daeb4b3" />

## What We Want To Happen
Since Bottoms Up is a multi-winner RCV method, we should use a static threshold. Default Bottoms Up Percentage Threshold to have `First Round Determines Threshold` enabled. Because there might be a jurisdiction that wants dynamic in the future, update the `First Round Determines Threshold` to make it toggleable. 

Once toggleable, we need to make the % of vote calculation match the configured setting. We need to change the % of vote calculation in the output writer. Previously it made any multi-winner method use Round 1 active ballots as the divisor.

But if we are using Bottoms Up Percentage threshold, which can now be configured either static or dynamic, we need to look at what is configured. If configured to static threshold we should use the Round 1 active ballots. If configured to dynamic threshold we should use the current Round active ballots.

yezr:
updated the ticket to say that we want it defaulted to static threshold and toggleable to dynamic

PR #947, opened 2025-08-20
Automatically determine the number of decimal places to use in the bottoms-up threshold

artoonie:
Currently, if decimal precision is set to 1, a threshold of 15% gets turned into 10%. That's weird. See #944 

This PR automatically determines the number of decimal places used in the threshold based on the threshold value entered. That means 15% is 0.15, and 15.0% is 0.150.

Added tests to demonstrate the functionality.

### DIFF
--- a/src/main/java/network/brightspots/rcv/ContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/ContestConfig.java
@@ -888,11 +888,60 @@ class ContestConfig {
     return rawConfig.rules.multiSeatBottomsUpPercentageThreshold;
   }
 
+  public static int numSigFigsInString(String number) {
+    if (number == null || number.isEmpty()) {
+      throw new IllegalArgumentException("No number provided");
+    }
+
+    boolean atMostOneDecimal = number.indexOf('.') == number.lastIndexOf('.');
+    if (!atMostOneDecimal) {
+      throw new IllegalArgumentException("Number " + number + " has too many decimal points");
+    }
+
+    String[] parts = number.split("\\.");
+    if (parts.length == 0 || parts.length > 2) {
+      throw new IllegalArgumentException("Number " + number + " is not a valid decimal number");
+    }
+
+    // Remove leading zeros from integer part
+    String intPart = parts[0];
+    intPart = intPart.replaceFirst("^0+", "");
+
+    // All zeros are significant in the fractional part
+    String fracPart = parts.length > 1 ? parts[1] : "";
+
+    // Ensure both sides are integers by casting them to ints and ensuring no errors
+    try {
+      if (!intPart.isEmpty()) {
+        Integer.parseInt(intPart);
+      }
+      if (!fracPart.isEmpty()) {
+        Integer.parseInt(fracPart);
+      }
+    } catch (NumberFormatException e) {
+      throw new IllegalArgumentException(number + " is not a valid fraction");
+    }
+
+    return intPart.length() + fracPart.length();
+  }
+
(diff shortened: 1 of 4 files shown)


src/test/resources/network/brightspots/rcv/test_data/multi_seat_bottoms_up_with_static_threshold/multi_seat_bottoms_up_with_static_threshold_expected_detailed_report.json
```diff
@@ -23,13 +23,13 @@
...
       "G" : "3",
-      "H" : "1"
+      "H" : "2"
     },
     "tallyResults" : [ {
       "eliminated" : "F",
       "transfers" : { }
     } ],
-    "threshold" : "3.1500"
+    "threshold" : "3.30"
```
artoonie, line 32: Note: the changes in this file are just what's caused by adding one additional vote to the CVR

yezr: image

artoonie: That's fine with me! I just wanted an easy-to-follow diff so we could ensure no other changes.

yezr:
With #946 being added in, I realized that we need to change the `% of vote` calculation in the output writer. Previously it made any multi-winner method use Round 1 active ballots as the divisor. 

But if we are using Bottoms Up Percentage threshold, which can now be configured either static or dynamic, we need to look at what is configured. If configured to static threshold we should use the Round 1 active ballots. If configured to dynamic threshold we should use the current Round active ballots. 

At `OutputWriter.java` line 445

I made an example with updated percentages [here](https://docs.google.com/spreadsheets/d/1AM-oqfW2JX8zawiyk5CRb7TmMBr15V-Ti1odFwvfHRU/edit?gid=671368869#gid=671368869) See the comment in Round 4 of the dynamic detailed report.

src/test/resources/network/brightspots/rcv/test_data/multi_seat_bottoms_up_with_static_threshold/multi_seat_bottoms_up_with_static_threshold_expected_detailed_report.json
yezr: I added those ballots and updated the shared CVR as well as the expected results files!

yezr:
That was the only place I saw the divisor math! thank you

yezr (approved):
LGTM!

yezr:
> For now, I just went with "All who reach threshold", but easy to change from here.

Can we change this to "All candidates over the X% threshold" for bottoms up using percent threshold and the new one detailed in #950

artoonie:
I'm realizing now that the other thresholds are expressed in terms of votes, so this is the only place in the file where the actual percentage threshold is used. I played around with some options but none seemed better than this.

One option is to include the threshold percentage in this line:
`Current Round Threshold,4.05,,,4.05,,,4.05,,,3.90,,,3.90,,`

But we'd have to decide whether we want that for all elections or just threshold bottoms-up.

yezr (approved):
LGTM!!!!!!!!!

PR #948, opened 2025-09-02
Track ties through code, print tie result in CSVs

### LINKED ISSUE #945: Show that a tie was resolved in the results report
## What Happens Now
Ties are addressed in the round by round logic according to however they are configured in the Winning Rules - TieBreak Mode setting. The results report will show that two candidates have the same count in a round. Other than that, nothing makes explicit that a tie happened or how it was resolved. 

## What We Want To Happen
We need to make it explicit in the results report that a tie occurred and was resolved. To do that we should

- Add an asterisk to the candidate name that was elected/eliminated in the tie. 
- Add the following text in the last line of the `detailed_report.csv` and `summary_report.csv`: `* Tie resolved in accordance with election law`

As we are changing the output, we should also increment the `CSV Format Version` to `2` in both detailed and summary reports for .csv output and the `JSONSummaryVersion` to 2 in .json output.

artoonie:
Resolves #945 

This PR needed a dependent PR to keep the diff clean and demonstrate that we only touch CSVs, not JSONs.


artoonie:
Good points in the review. Before, I was overoptimizing and setting the footnote used status only when we "naturally" discovered it was used in the course of writing out the file. While that happened to work, it was brittle: it would be too easy to try using `isFootnoteUsed` before it was set. I have changed it to have a separate iteration through the rounds to check if the footnote is used, and setting that in the constructor.

Comments addressed, I believe!

src/main/java/network/brightspots/rcv/Tabulator.java
```diff
@@ -802,19 +860,27 @@ private List<String> doRegularElimination(
...
+            wasTiebroken,
+            currentRound));
     return eliminated;
   }
 
+  // getter method for round decisions (used by OutputWriter)
+  Map<Integer, List<TallyDecision>> getRoundToDecisions() {
+    return roundToDecisions;
+  }
+
```
yezr, line 875: Looks like this was unnecessary since the Tabulator creates the OutputWriter just below

src/main/java/network/brightspots/rcv/OutputWriter.java
```diff
@@ -426,11 +428,15 @@ private void generateCsvReport(
     }
     csvPrinter.println();
 
-    csvPrinter.print(outputFileIdentifiers.isSlice() ? "Eliminated*" : "Eliminated");
-    printActionSummary(csvPrinter, roundToEliminatedCandidates);
+    Footnote tallyResultInSlice = new Footnote("†", outputFileIdentifiers.isSlice());
```
yezr, line 431: I'm not sure what this variable name is trying to communicate. Can it be `isSliceFootnote` similar to the tiebreak one

yezr (approved):
LGTM!

PR #958, opened 2025-10-13
fix validation error during bottoms-up

### LINKED ISSUE #956: Bottoms-Up validation error when number of winners isn't configured
Image

Using bottoms up, I accidentally didn't configure Number of Winners because I was testing out some threshold configurations. Validation fails as expected. However, the GUI validation error message doesn't say explicitly what is causing the validation to fail. 

It says
``` 2025-10-03 11:23:35 EDT SEVERE: Error during validation:
java.lang.NumberFormatException: For input string: ""
Validation failed!
```

It would be good to say explicitly what it is causing validation to fail, that it is the Number of Winners not being configured.

artoonie:
Closes #956 

We were getting an exception thrown when reading getNumberOfWinners before we validated that that field was valid.

Fix is to just ignore this exception if thrown.

Also included:
1. On development versions, include a stack trace. Having it hidden was annoying.
2. I also saw that `firstRoundDeterminesThreshold` defaulted to true, but must be false for multi-winner elections, so you have to immediately unset it. I can't imagine this was intentional. I've removed the ability to toggle this on bottoms-up.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,7 +1,8 @@
 name: "Build, Lint, and Test"
 
 on:
-  - push
+  - pull_request
+  - merge_group
 
 jobs:
   build:


yezr (dismissed):
Confirmed that the development version stack trace works. Confirmed that the error message in the UI when leaving number of winners blank has our user friendly message now!

yezr (approved):

PR #987, opened 2026-03-09
Do not duplicate autoloaded candidates across multiple CVRs

### LINKED ISSUE #986: Candidate Auto-Load doesn't make distinct list when configured with multiple CVRs
## What Happens
Configure multiple CVRs with the same candidates in ES&S, [like the 2024 CD2 CVRs from the Main SoS website](https://www.maine.gov/sos/elections-voting/election-results-data/election-results-2024#ranked). Auto-load candidates. The list that is returned has duplicates for each configured CVR. So for three CVRs I have three separate copies of the three distinct candidates in the CVRs, like this
- Golden, Jared Forrest
- Theriault, Austin
- Write-in
- Golden, Jared Forrest
- Theriault, Austin
- Write-in
- Golden, Jared Forrest
- Theriault, Austin
- Write-in

## What Should Happen
The list that the candidate auto-load returns should only have unique candidate/code pairs. So these CVRs should only have
- Golden, Jared Forrest
- Theriault, Austin
- Write-in

artoonie:
Resolves #986 

We were creating a hashset to prevent duplicate candidates, but that hashset was hashing Candidate objects. Implement .hashCode and .equals to prevent duplicates across multiple CVRs.

### DIFF
--- a/src/main/java/network/brightspots/rcv/RawContestConfig.java
+++ b/src/main/java/network/brightspots/rcv/RawContestConfig.java
@@ -325,6 +325,33 @@ public class RawContestConfig {
       return "Name: " + name + " Aliases: " + aliases;
     }
 
+    /** The key for dictionaries includes:
+      1. [excluded|not]
+      2. name
+      3. newline-separated aliases
+      This is unique for each candidate because newline is the only truly-disallowed character
+     */
+    @Override
+    public int hashCode() {
+      String nameString = getName() == null ? "" : getName();
+      String aliasesString = String.join("\n", getAliases());
+      return "%b%n%s%n%s".formatted(getExcluded(), nameString, aliasesString).hashCode();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (!(o instanceof Candidate other)) {
+        return false;
+      }
+      return this.getExcluded() == other.getExcluded()
+         && this.getName().equals(other.getName())
+         && this.getAliases().equals(other.getAliases());
+    }
+
+
     public String getName() {
       return name.getValue();
     }


src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -325,6 +325,33 @@ public String toString() {
...
     }
 
+    /** The key for dictionaries includes:
+      1. [excluded|not]
+      2. name
+      3. newline-separated aliases
+      This is unique for each candidate because newline is the only truly-disallowed character
+     */
+    @Override
+    public int hashCode() {
```
yezr, line 335: What happens if we add a new property to Candidate? It would be fine if it isn't included here, so long as these properties still identify a unique candidate?

src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -325,6 +325,33 @@ public String toString() {
...
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (!(o instanceof Candidate other)) {
+        return false;
+      }
+      return this.getExcluded() == other.getExcluded()
```
yezr, line 349: Same question as above: would this need to be updated if we add properties to the Candidate object?

src/main/java/network/brightspots/rcv/RawContestConfig.java
artoonie: We should almost certainly include any new property to these two functions, unless there's a good reason why two candidates that have different values for that field are considered equal (like `auditTrailCandidateFirstLoadedFromFile` or `lastUpdated` or something). I can't imagine realistically adding any field that _doesn't_ belong in these two functions.

yezr: How can we ensure this gets updated when/if the properties on the `Candidate` object change?

artoonie: I looked into two options:

1) **Using a java `record`**: does not work because this class is used in deserializing json and therefore it needs setters and is no longer immutable. There seem to be [some workarounds](https://stackoverflow.com/questions/74890806/parse-json-to-java-records-with-fasterxml-jackson) but the complexity seems to outweigh the benefits. 
2) **Using [EqualsVerifier](https://jqno.nl/equalsverifier/)** seems incompatible with Candidate, again because it subclasses SimpleStringProperty and therefore has some more complex properties under the hood: `superclass access check failed: class nl.jqno.equalsverifier.internal.reflection.ExpressionHelper$$DynamicSubclass$7ae9a33a$ (in module nl.jqno.equalsverifier) cannot access class com.sun.javafx.binding.ExpressionHelper (in module javafx.base) because module nl.jqno.equalsverifier does not read module javafx.base`

I'm not sure we're going to beat "try to avoid programmer error" here

artoonie:
TODO: Add tests, preferably using multiple CVR types with the same data

artoonie:
Multiple CVRs with same data created in #990 
Tests added in #991 

This is ready for review, and we should decide on a merge order. I recommend merging this first, then #990, then #991.

yezr (approved):
lgtm

PR #988, opened 2026-03-11
Support ES&S Multi-Contest CVRs

### LINKED ISSUE #981: ES&S Multiple Contests with different ballot styles
[CVR](https://github.com/user-attachments/files/25270081/1_8_26.rctab.run.w_.randy.ES.S.CVRs.multiple.contests.xlsx)
config: [es&s_multiple_contests_different_ballot_styles_config.json](https://github.com/user-attachments/files/25270084/es.s_multiple_contests_different_ballot_styles_config.json)

## Background
Currently, RCTab expects an ES&S CVR where all exported ballots have the contest it is configured to tabulate. Additionally, RCTab can be configured to interpret any text in the CVR's ranking data as an empty ranking, and tabulate accordingly. By default, ES&S CVRs contain the text `undervote` in a ranking that was not marked on a ballot. Single contest exports in ES&S satisfies these assumptions. 

Multi-Contest exports, however, show ` ` in the CVR's ranking data for contests that don't appear on a ballot. We'd like to support multi-contest exports. 

This support is important for timely reporting of RCV results. ES&S CVR exports can take ~30 minutes for ~300,000 ballots. For jurisdictions that have multiple contests to export, requiring an additional export for each contest can start to add up. We'd like to support multi-contest exports so that only one export is required for any number of contests. Additionally, a single CVR file for all contests instead of a separate CVR file for each contest can make contest configuration simpler, leading to more reliably successful configurations. 

## What Happens
We have an ES&S CVR with multiple contests and multiple ballot styles. In this ES&S CVR, when a ballot does not have a contest, the rankings for that contests are all blank. This is different from rankings that were not marked on a ballot where the contest was on the ballot, these are marked with the text `undervote`. 

Let's use the `DEM DELEGATE TO THE U.S. HOUSE OF REPRESENTATIVES FROM THE DISTRICT OF COLUMBIA` contest, whose five rankings are in columns `I` to `M`. 44 of the 147 records in the CVR have rankings for this contest. Currently, when configured correctly to look in columns `I` to `M` for the contest rankings, RCTab sees 103 rows where there are blanks for all 5 rankings. It reports that `Unrecognized candidate "" appears 515 time(s)!` 5 blank rankings x 103 ballots = 515 rankings

Image

## What we want to happen
- Ballots with all blank ranks for the contest's rankings don't show up in the universe at all, don't contribute to the threshold calculation and don't contribute to the `Number of Undervotes (No Rankings)` number
## Open Questions
- Must be very sensitive to filtering out ballots that are in the configured CVR. How can we report this out for the user to confirm? 
- Separate from this ticket, any other ways we can make configuring the multi-contests ES&S CVR easier? Like a column list with the first 5 rows of data visually shown so you can select which columns your rankings are in vs. currently configuring by explicit column index

artoonie:
Closes #981 

## ES&S CVR Format Updated Assumptions
As we got into this ticket, we started to see clearly that getting some of our assumptions confirmed about the structure of the ES&S CVR export format would help. We were able to speak to ES&S reps about the CVR export and confirmed some assumptions. 
- Most importantly, we confirmed that **it is always the text "undervote" when a ballot has an option to rank/vote in a contest and no ranking/vote is made**. 
- When a ballot is included in an CVR export and a contest is included in that export that doesn't appear on that ballot, the rankings/votes for that contest will all be blank.
- For write-ins, there is an option in the export to either 1) insert a picture of the write-in portion of the ballot in the cell of that ranking. That picture is interpreted in our parser as a blank OR 2) the text "Write-in". I'm going to find a CVR export that has the pictures in it to confirm that. If we could at least differentiate between a picture and a blank that might help us even more.

## Updates

Now that we can rely on the above assumptions, we've made the following changes.

This PR deprecates "Treat blank as UWI" and "Undervote label." For ES&S, the undervote label is hardcoded. For UWI, we now explicitly search the XLSX for write-in images. Blanks are invalid unless the entire row is blank. When that happens, we report it as part of an irrelevant contest in the audit logs.

This PR discovered that the 2015 Portland Mayor CVRs had images, but "treat blank as UWI" was false. That meant we mistakenly treated blanks as undervotes, not UWI. This PR fixes that in two ways:
1. 2015 Portland Mayor test: update the output files, now properly handling UWI
2. 2015 Portland Mayor Codes test: update the input CVR, removing the images and replacing with "undervote," and demonstrating the output files do not change.

There were several other ES&S xlsx files that included images in some far-away row or column. Those have been updated.

In order to include UWI in the output, you must input something in the UWI label field. This serves as a bit of a magical boolean: if the ES&S file includes images, you must input a non-null string here. That string is never used.

We were not testing the migrations before which led to several bugs that were invisible to the tests but prevented manually testing ES&S write-ins. To prevent that issue, I now include migrations in the test files. Because of this, we shouldn't rely on the "version=TEST" in the config JSON: the "version" field is overloaded and how do we know which migrations to run if the version is set to TEST? So I've directly detected if the config file is in the test directory instead of looking at the version field. I have not updated all configs to remove VERSION but I can do that in a separate PR (or here).

Given two fewer config options on the right-hand column, the GUI columns became imbalanced. I moved "Contest ID" from the left column to the right to balance it.
image

Each change in this PR is simple, but due to the cascading effects, there are several intertwined tasks. Let me know if the review is cumbersome and I can split this PR into multiple chained PRs.


artoonie:
This PR found a bug:

While Apple Numbers only shows non-empty rows, the underlying XLSX XML data can store empty rows. That led to some CVRs having hidden empty rows which we erroneously marked as undervotes. We now correctly detect these rows as rows to be completely ignored and thrown out.

artoonie:
TODO: Audit log should say "read X ballots, of which Y were ignored" or similar, rather than X=only non-empty ballots

artoonie:
> TODO: Audit log should say "read X ballots, of which Y were ignored" or similar, rather than X=only non-empty ballots

Rather than modifying that row, I opted to add a warning so (A) it's more visible, and (B) we don't need to propagate what is likely source-specific values up from the ES&S reader through the generic reader and into the tabulator session.

It says "Ignored %d rows with no votes for any candidates"

yezr:
Testing I found that setting `Treat Blanks As Undeclared Write-In` to `true` has some unexpected behavior now. When it sees a fully blank record it calls `handleEmptyCells` at the beginning of `endCvr()`. It writes into each of those blank ranks an undeclared write-in ranking. 

However, the `hasSeenAnyNonBlankCandidateCells` boolean is only updated in the cell data call back, which never gets fired because all cells are blank. The ranking of all Undeclared Write-In ranks is then filtered out. 

Maybe we need to set the `hasSeenAnyNonBlankCandidateCells` to true when we set a UWI?

artoonie:
It seems that `Treat Blanks As Undeclared Write-In` is fundamentally incompatible with this approach. Either blanks mean UWI or it means Unrelated Contest. I wonder if we should go back to the drawing board and consider looking at the other columns to disambiguate. Or, better yet -- find out what's actually possible in the ES&S data format so we can support it without guessing.

If we want to keep this approach, I suggest changing the `Treat Blanks As Undeclared Write-In` from a checkbox to a dropdown: `Blanks are... > [Ignored, Treated as Undeclared Write-Ins]`

yezr:
> If we want to keep this approach, I suggest changing the Treat Blanks As Undeclared Write-In from a checkbox to a dropdown: Blanks are... > [Ignored, Treated as Undeclared Write-Ins]

I think you are right that it needs to be mutually exclusive. For our options
- default option of `Any Blank Invalid - Causes Tabulation to Fail`
- for any single blank being treated as a UWI, the current behavior when the `Treat Blanks As Undeclared Write-In` is checked, it now says `Any Blank Valid - Individual Blanks Treated As UWI`
- I think for the `Ignored` option we might change the description to make it explicit that it is only ignored if it is a full ranking of blanks. Maybe `Valid Only As An All Blank Ranking Ballot - Not included in tabulation (For Multi-Contest Exports)`

We can work on this vocab...

artoonie:
Makes sense. 

Suggested vocab:
1. Invalid (fail tabulation)
2. Undeclared Write-In (valid if only one blank per voter)
3. Irrelevant Contest (skip row if all rankings are blank)

With 3, the CSV field listing the number of voters who didn't vote at all (Undervotes-No Ranking) would always be zero, and that's invalid. Does it make sense to write "unknown" instead of zero for that field?

artoonie:
This change is getting larger than it seemed as first, so we'll definitely need to add tests to handle each of these options. The code isn't ready yet, but the functionality is more or less there if you want to give it an initial test @yezr to verify this is the right direction.

yezr:
As we got into this ticket, we started to see clearly that getting some of our assumptions confirmed about the structure of the ES&S CVR export format would help. I was able to speak to ES&S reps about the CVR export and confirmed some assumptions. 
- Most importantly, I confirmed that **it is always the text "undervote" when a ballot has an option to rank/vote in a contest and no ranking/vote is made**. 
- When a ballot is included in an CVR export and a contest is included in that export that doesn't appear on that ballot, the rankings/votes for that contest will all be blank.
- For write-ins, there is an option in the export to either 1) insert a picture of the write-in portion of the ballot in the cell of that ranking. That picture is interpreted in our parser as a blank OR 2) the text "Write-in". I'm going to find a CVR export that has the pictures in it to confirm that. If we could at least differentiate between a picture and a blank that might help us even more.

That should at least remove the need to handle blanks as skipped ranks.

yezr:
Finally got an example ES&S CVR with write-in snippets. 
[ess_cvr_with_images.xlsx](https://github.com/user-attachments/files/26576241/ess_cvr_with_images.xlsx)

It's a plurality contest, but it should have the same structure as an RCV CVR.

artoonie:
This gets fun-- images aren't associated with cells, they're simply placed relative to cells but are global items.

Step 1: Go through all images, find which cells they seem to be associated with, and mark those cells as UWI
Step 2: When parsing, check the list of UWI images we found

That means blanks can either only be "invalid" or "irrelevant contest".

Code needs clean-up, but app is ready to test @yezr

yezr:
What we know now about the ES&S CVR structure
- We know that there is only one possible reason that we would ever see blanks in an ES&S CVR. If a contest included in the export is not a contest on a ballot that is included in the export, then ALL rankings for that contest will be blank. Any other blanks, like a single blank in a contest that has at least one ranking, would be a halting error. 
- We know that if a ballot has an opportunity to rank a contest, but choosing not to use one or more rankings in that contest, the text `undervote` will be in the ranking cell. 

Also, with the changes we've made we are now able to "see" images in the CVR. Before, these were being parsed as empty cells. Now, we are able to differentiate between an empty cell (an unused ranking), and a cell that has an image (a ranking for a write-in candidate). 

I think then we can get rid of some config options in RCTab
- We can get rid of the `Undervote Label` configuration textbox altogether. We know that this will always be the text `undervote`. 
- We can get rid of any `Treat Blank As _____` configuration options. The only time a blank is not a halting exception that stops tabulation is if ALL rankings for a contest are blank. In that case, the contest was not on the ballot and that ballot should not be included in tabulation. Any other blank is a malformed CVR and should halt tabulation. 
- I think we can also get rid of the `Undeclared Write-In Label` and `Treat Blank as Undeclared Write-in`? Now we know that we can flag something as UWI if the ranking cell is an image or the text `Write-in`.

src/main/java/network/brightspots/rcv/ContestConfig.java
```diff
@@ -1276,15 +1265,12 @@ private void processCandidateData() {
...
-          || source.getTreatBlankAsUndeclaredWriteIn()) {
-        includeUwi = true;
-        break;
+      if (!isNullOrBlank(source.getUndeclaredWriteInLabel())) {
+        return true;
       }
     }
-    return includeUwi;
+    return false;
   }
```
artoonie, line 1283: TODO -- think about how to handle this.

Currently the logic is based on whether UWIs are configured. For ES&S, there is no configuration needed: the presence of an image is sufficient. So we'd need to refactor this logic to occur after we read the CVRs.

artoonie: @yezr I'm glad I didn't ignore this, here's what was happening:

UWI text was previously enabled in the UI, so our current tests still specify the UWI label field for ES&S. Then, this function returned true. That means our test files previously-created would all pass, but no new configs with UWIs would work.

I will need a migration that removes the UWI label from ES&S sources. To do that, I'll need to update the migrations to work on test source files.

I think eventually we'll need a better way to determine if a config file is part of the test suite -- setting the version to TEST overloads what we use the version for.

For now, take a look at my last commit. If it looks reasonable, I can update tests to match.

artoonie: Okay, this is now complete! The last commit also had to replace all blanks with the phrase "undervote" because previously, while it would write SEVERE to the logs, it wouldn't actually fail.

src/main/java/network/brightspots/rcv/StreamingCvrReader.java
```diff
@@ -369,10 +420,133 @@ public void headerFooter(String s, boolean b, String s1) {
     // close zip file without saving
     pkg.revert();
 
+    if (numRowsIgnoredBecauseAllBlank > 0) {
+      Logger.warning("Ignored %d rows with no votes for any candidates.",
```
yezr: I could see someone confusing this language with a full undervote. Can we be more clear here that we are ignoring because the ES&S formatting indicates this ballot did not have our configured contest on it? Something like 

`Ignored %d rows because the ES&S CVR format indicates the configured contest did not appear on those ballots.`

src/main/java/network/brightspots/rcv/GuiConfigController.java
```diff
@@ -1642,7 +1611,6 @@ private boolean isAnyPropertyValueFunctionMissing(EditableColumn[] columnsAndPro
   private void loadConfig(ContestConfig config) throws ConfigVersionIsNewerThanAppVersionException {
     clearConfig();
     RawContestConfig rawConfig = config.getRawConfig();
-    ContestConfigMigration.migrateConfigVersion(config);
```
yezr, line 1645: This is not needed any more? Does the migration happen elsewhere?

src/main/java/network/brightspots/rcv/RawContestConfig.java
```diff
@@ -453,9 +442,8 @@ public static class ContestRules {
 
     // These are deprecated (moved to individual CVRs), but we need to leave them in place here for
     // the purpose of supporting automatic migration from older config versions.
-    public boolean treatBlankAsUndeclaredWriteIn;
+    // This only includes fields that were moved or renamed, not deprecated
```
yezr: The new comment says these are for fields that are not deprecated, the comment on 443 says they are deprecated. On 443 should it maybe say `These are moved to individual CVRs, but we need to leave them in place here for`

src/main/java/network/brightspots/rcv/StreamingCvrReader.java
```diff
@@ -106,9 +125,11 @@ final class StreamingCvrReader extends BaseCvrReader {
             : null;
     this.overvoteDelimiter = source.getOvervoteDelimiter();
     this.overvoteLabel = source.getOvervoteLabel();
-    this.skippedRankLabel = source.getSkippedRankLabel();
     this.undeclaredWriteInLabel = source.getUndeclaredWriteInLabel();
-    this.treatBlankAsUndeclaredWriteIn = source.getTreatBlankAsUndeclaredWriteIn();
+  }
+
+  private static long hashForCell(int row, int col) {
```
yezr: We discussed this in conversation, but can you include a comment here about what exactly this is doing?

src/test/resources/network/brightspots/rcv/test_data/2015_portland_mayor/2015_portland_mayor_expected_detailed_report.json
```diff
@@ -30,17 +30,46 @@
       "Miller, Markos S." : "3",
       "Rathband, Jed" : "2",
       "Strimling, Ethan K." : "1",
-      "Undeclared Write-ins" : "0",
+      "Undeclared Write-ins" : "1",
```
yezr, line 33: FYI for later: there is an image in the first rank for one ballot that has chris vail as its second choice. This config uses batch elimination. Previously the image was interpreted as a blank cell, which was interpreted as a skip.

src/test/resources/network/brightspots/rcv/test_data/test_set_treat_images_as_undeclared_write_in/2015_portland_mayor_codes_cvr.xlsx
yezr, line 1: this maybe got moved on accident into this folder? Is it a copy? The original file in test_data/2015_Portland_mayor_codes shows as +/- in github and still seems to be in it's original location....maybe because it was removed and added in subsequent changes in this PR?

src/test/resources/network/brightspots/rcv/test_data/2015_portland_mayor_codes/2015_portland_mayor_codes_cvr.xlsx
yezr, line 1: looks like just removed all images and replaced with an undervote?

Image

src/test/resources/network/brightspots/rcv/test_data/_shared/bottoms_up_cvr.xlsx
yezr, line 1: Image

src/test/resources/network/brightspots/rcv/test_data/_shared/simple_ess_cvr.xlsx
yezr, line 1: Image

src/test/resources/network/brightspots/rcv/test_data/_shared/simple_sequential_cvr.xlsx
yezr, line 1: Image

src/test/resources/network/brightspots/rcv/test_data/_shared/tiebreak_cvr.xlsx
yezr, line 1: not sure how these are different?

Image

src/test/resources/network/brightspots/rcv/test_data/continue_tabulation_test/continue_tabulation_test_cvr.xlsx
yezr, line 1: Image

src/test/resources/network/brightspots/rcv/test_data/continue_until_two_threshold_freeze/continue_until_two_threshold_freeze_cvr.xlsx
yezr, line 1: Image

src/test/resources/network/brightspots/rcv/test_data/ess_multi_contest/ess_multi_contest_config.json
```diff
@@ -0,0 +1,86 @@
+{
+  "tabulatorVersion" : "TEST",
```
yezr, line 2: should we check in a v2.1.0? Should we have any more TEST version configs?

src/test/resources/network/brightspots/rcv/test_data/duplicate_test/duplicate_test_cvr.xlsx
yezr, line 1: Image

src/test/resources/network/brightspots/rcv/test_data/first_round_determines_threshold_test/first_round_determines_threshold_test_cvr.xlsx
yezr, line 1: Is there a reason to add an image to this one since we test image to UWI elsewhere? 

Image

src/test/resources/network/brightspots/rcv/test_data/first_round_determines_threshold_tiebreaker_test/first_round_determines_threshold_tiebreaker_test_expected_detailed_report.json
```diff
@@ -49,7 +49,7 @@
     "finalThreshold" : "9",
     "numCandidates" : 4,
     "numWinners" : 1,
-    "totalNumBallots" : "19",
-    "undervotes" : 2
+    "totalNumBallots" : "17",
```
yezr, line 52: This was two "invisible" totally blank rows that were previously being interpreted as a full overvote. We noticed them now because they were being filtered instead. So we just got rid of them in the CVR

src/test/resources/network/brightspots/rcv/test_data/first_round_determines_threshold_test/first_round_determines_threshold_test_expected_detailed_report.json
```diff
@@ -18,38 +18,59 @@
...
     "threshold" : "9"
   } ],
   "summary" : {
     "finalThreshold" : "9",
-    "numCandidates" : 4,
+    "numCandidates" : 5,
     "numWinners" : 1,
-    "totalNumBallots" : "19",
-    "undervotes" : 3
+    "totalNumBallots" : "16",
```
yezr, line 52: same as threshold tiebreaker test: these were three "invisible" all empty rows that were previously being read as an undervote, we noticed they were being filtered out, so we removed them altogether

src/test/resources/network/brightspots/rcv/test_data/more_winners_than_candidates/more_winners_than_candidates_config.json
```diff
@@ -52,7 +52,7 @@
     "decimalPlacesForVoteArithmetic" : "4",
     "minimumVoteThreshold" : "0",
     "maxSkippedRanksAllowed" : "1",
-    "maxRankingsAllowed" : "8",
+    "maxRankingsAllowed" : "3",
```
yezr, line 55: was just misconfigured. there are only 3 columns in the CVR. Previously, these were just being skipped over. Now they cause an error, so it has been configured to be explicit now.

src/test/resources/network/brightspots/rcv/test_data/stop_tabulation_early_test/stop_tabulation_early_test_config.json
```diff
@@ -48,7 +48,7 @@
     "decimalPlacesForVoteArithmetic" : "4",
     "minimumVoteThreshold" : "0",
     "maxSkippedRanksAllowed" : "0",
-    "maxRankingsAllowed" : "3",
+    "maxRankingsAllowed" : "1",
```
yezr, line 51: tiebreak CVR that this uses only has one ranking column. Previously would be skipped over, now it errors when it sees those two blank columns. Configured correctly now to prevent errors

src/test/resources/network/brightspots/rcv/test_data/test_set_treat_blank_as_undeclared_write_in/test_set_treat_blank_as_undeclared_write_in_cvr.xlsx
yezr, line 1: should we get rid of this whole folder? since it has effectively been moved to the treat_image_as_UWI?

src/test/resources/network/brightspots/rcv/test_data/tiebreak_generate_permutation_test/tiebreak_generate_permutation_test_cvr.xlsx
yezr, line 1: and can this get moved to the shared folder? I think there is a copy of this somewhere. I think it's the /_shared/tiebreak_cvr.xlsx

src/test/resources/network/brightspots/rcv/test_data/tiebreak_previous_round_counts_then_random_test/tiebreak_previous_round_counts_then_random_test_cvr.xlsx
yezr, line 1: Image

yezr (commented):
commented on each of the .XLSX cvrs that changed to make sure the differences are as expected. Most is just for posterity to ensure we know changes, some have questions attached

src/main/java/network/brightspots/rcv/GuiConfigController.java
artoonie: Yes, it happens in loadContestConfig now

src/main/java/network/brightspots/rcv/RawContestConfig.java
artoonie: Ah, s/deprecated/deleted. Clarified!

src/test/resources/network/brightspots/rcv/test_data/2015_portland_mayor_codes/2015_portland_mayor_codes_cvr.xlsx
artoonie: Yeah I treated these two differently, noted in the top-level PR comment:
`2015 Portland Mayor Codes test: update the input CVR, removing the images and replacing with "undervote," and demonstrating the output files do not change.`

src/test/resources/network/brightspots/rcv/test_data/test_set_treat_blank_as_undeclared_write_in/test_set_treat_blank_as_undeclared_write_in_cvr.xlsx
artoonie: It is effectively removed from git, so if it still exists it's probably just a local `output` folder on your machine. git won't delete non-empty directories.

src/test/resources/network/brightspots/rcv/test_data/tiebreak_generate_permutation_test/tiebreak_generate_permutation_test_cvr.xlsx
artoonie: There are similar names in other CVRs but none that match exactly this

src/test/resources/network/brightspots/rcv/test_data/2015_portland_mayor_codes/2015_portland_mayor_codes_cvr.xlsx
yezr: thank you armin for the clarification!

src/test/resources/network/brightspots/rcv/test_data/tiebreak_generate_permutation_test/tiebreak_generate_permutation_test_cvr.xlsx
yezr: you're right!

src/test/resources/network/brightspots/rcv/test_data/_shared/tiebreak_cvr.xlsx
artoonie: Rows 46, 65, 72 have images. The word wrap is probably just a difference from a Google Sheet vs Microsoft Excel export

src/test/resources/network/brightspots/rcv/test_data/test_set_treat_images_as_undeclared_write_in/2015_portland_mayor_codes_cvr.xlsx
artoonie: Deleted

src/test/resources/network/brightspots/rcv/test_data/_shared/tiebreak_cvr.xlsx
yezr: Ah! I never went lower than 9! I see those hanging out down there in the old version now. I also tested some different variations of max rankings and saw how the new logic handles the old one. It all works!

src/test/resources/network/brightspots/rcv/test_data/first_round_determines_threshold_test/first_round_determines_threshold_test_cvr.xlsx
yezr: reverted this, it was not intended

src/main/java/network/brightspots/rcv/ContestConfigMigration.java
```diff
@@ -57,126 +58,142 @@ static boolean isConfigVersionNewerThanAppVersion(String configVersion) {
...
-        for (CvrSource source : rawConfig.cvrFileSources) {
-          source.setUndeclaredWriteInLabel(rules.undeclaredWriteInLabel);
-        }
+  private static void migrate201To210(ContestConfig config) {
+    RawContestConfig rawConfig = config.getRawConfig();
+
+    for (CvrSource source : rawConfig.cvrFileSources) {
+      if (StringUtil.isNotBlank(source.getSkippedRankLabel()) && ContestConfig.Provider.ESS
+              == ContestConfig.Provider.getByInternalLabel(source.getProvider())) {
+        Logger.warning("ES&S no longer supports custom undervote labels. Ignoring.");
```
yezr: Image

This seems a little ominous. Can we check to see if the undervote label is `undervote`?

If it is `undervote`, then we can just carry on: everything should work.

If it isn't `undervote`, let's do the warning message with some more context like `Migrating config to v2.1.0 - the previously configured undervote label [previousUndervoteLabel] is not compatible with the ES&S default of "undervote". Please review your CVR to ensure undervotes are properly identified with "undervote"` Since we now make the assumption that all undervotes are the text `undervote`.

yezr (changes requested):
this was dangling in github for some reason. last request for a change I promise!!!!

src/main/java/network/brightspots/rcv/ContestConfigMigration.java
artoonie: I'll note that this only happens if you're using an old config on a new version. EAs would always be creating new configs from scratch, so they're unlikely to ever see this error. Happy to make this change but wanted to flag this before I do.

yezr: I could see the possibility an admin using a config from an old election to get started on a new one. Like using it as a template to copy over the voter error rules but then changing the pertinent contest details for the new contest. 

Thanks for flagging! I think the change is still pertinent.

artoonie: Done!

src/main/java/network/brightspots/rcv/ContestConfigMigration.java
```diff
@@ -93,9 +93,15 @@ private static void migrate201To210(ContestConfig config) {
...
-              == ContestConfig.Provider.getByInternalLabel(source.getProvider())) {
-        Logger.warning("ES&S no longer supports custom undervote labels. Ignoring.");
+      if (StringUtil.isNotBlank(source.getSkippedRankLabel())
+              && ContestConfig.Provider.ESS
+                == ContestConfig.Provider.getByInternalLabel(source.getProvider())
+              && !"undervote".equals(source.getSkippedRankLabel())) {
+        Logger.warning(
+                "Migrating config to v2.1.0 - the previously configured undervote label "
+                + source.getSkippedRankLabel()
+                + " is not compatible with the ES&S default of \"undervote\". Please review your "
```
yezr, line 103: nit: can we wrap the source.getSkippedRankLabel() with quotes, similar to `\"undervote\"`

artoonie: done!

yezr (approved):
Thank you so much Armin for the thorough back and forth on this big one. Also huge thank you for the revised original PR description. Quit a bit changed throughout, and being able to go back months from now and get all the context from the original comment instead of digging through the bazillion updates is super helpful

LGTM

PR #990, opened 2026-03-24
test data: same data, different formats

artoonie:
These tests are designed to aid with testing autoload (PR #987)

This PR takes an existing CDF CVR, duplicates it, and copies the data into the ES&S format. Still TODO: handle other formats. I think this PR can be merged without having all providers in here.

How this works:
- A base config file has an empty CVR sources section, `base_config`
- There are `stem_config.json` files which populate the CVR sources to generate a runnable config
- We generate the config for each stem and tabulate
- We compare all the output files, JSON and CSV, to ensure equivalence between providers


yezr:
This is great to enable the minor config changes needed for these new tests!

 The `sameDataDifferentFormats` test that was added here, does that test any tabulator functionality specifically? Or is that mainly just a basic example of how to use this new harness to change the configs? I guess for now it is a test that whatever we did to copy the CVR data from one provider to another works? And if we can count on that to be the same, then we can use it for the GuiTests that are coming with #991?
 
 That CVR copying code could also be interesting to explicitly check the assumptions we make for each provider? Though that could bloat this repo.

artoonie:
This essentially does not test anything new, other than the confirmation that there is no provider-specific information in the outputs.

The goal is just to create a basic example of this new test harness, which should eventually include all providers.

Creating the ES&S data was easy, as I just used the RCTab CVR format and modified it a touch. I started with an existing CDF file because that data is the hardest to manually create.

src/test/java/network/brightspots/rcv/TabulatorTests.java
```diff
@@ -491,6 +494,78 @@ static void setup() {
...
+            fileCompare(firstCsv, detailedCsvPaths.get(i).toString()),
+            "detailed_report.csv differs between \"" + stems[0] + "\" and \"" + stems[i] + "\"");
+      }
+
+      // Clean up output folders on success
+      for (TabulatorSession session : sessions) {
+        cleanOutputFolder(session);
+      }
+    } finally {
+      for (File tempConfig : tempConfigs) {
```
yezr, line 563: Should we leave the temp configs so we can see what they are for troubleshooting later?

yezr (approved):

src/test/java/network/brightspots/rcv/TabulatorTests.java
artoonie: When debugging them I just comment out the File.delete line, but leaving them in leaves git clutter. In general the tests all try to clean up after themselves.

PR #991, opened 2026-03-25
Add tests for autoload

artoonie:
Uses the new test harness in #990 to add a regression test for #987.

While the tests fail without #987 merged into this PR, they would pass once #987 is merged in. Not merging them for now to reduce dependencies between PRs until we decide merge order.

This PR:
1. Creates Common.java for shared test helpers
2. Creates GuiFunctionalityTests.java for non-tabulation tests, currently only testing autoload but I expect we can use it to test other functionality that only affects the GUI
3. In GuiConfigController.java, splits out the autoload code into a separate function that can be tested. (For a cleaner diff, make sure you "Hide Whitespace" in github. It's mostly moving code to another function without real changes.)
4. Updates TabulatorTests to use the new functionality in Common.java
5. Modifies the CVRs to remove aliases. They're tested elsewhere and they complicate autoload. \CDF and Dominion have "IDs" which autoload will find, creating differences between those formats and ES&S where IDs aren't used. Simpler to simply remove the concept of IDs for these specific tests.


src/test/java/network/brightspots/rcv/GuiFunctionalityTests.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  @Test
+  @DisplayName("test autoload when one candidate is missing")
+  void testAutoloadWithOneMissingCandidate() throws Exception {
+    Common.runForEachProvider(
+        configPath -> {
+          String configPathStr = configPath.toString();
+          ContestConfig config = ContestConfig.loadContestConfig(configPathStr);
+          assertSame(1,
```
yezr: Should we be more explicit about the Candidate objects that `gatherAutoloadedCandidates` returns? I'm thinking about the aliases earlier, should we be explicit here that, for example, we have a candidate object with no aliases?

src/test/java/network/brightspots/rcv/Common.java
```diff
@@ -0,0 +1,80 @@
...
+  static void runForEachProvider(ConfigConsumer consumer) throws Exception {
+    runForEachProvider(consumer, UnaryOperator.identity());
+  }
+
+  /**
+   * Like {@link #runForEachProvider(ConfigConsumer)}, but applies {@code configModifier} to the
+   * merged config JSON before writing it to disk. Use this to adjust the config for tests that
+   * need a non-default starting state (e.g. removing all candidates to test autoload).
+   */
+  static void runForEachProvider(
```
yezr, line 53: I can see using `runForEachProvider` with a different set of .configs (that maybe has a different set of CVRs). Is this extendable to use a different config? 

I see the config hard coded in here but somehow that functional interface allows a lambda that maybe could take different configs?

yezr (commented):
a question about `runForEachProvider`

src/test/java/network/brightspots/rcv/Common.java
artoonie: We could, but I didn't go that route to avoid having to maintain a bunch of config files. The configModifier should let us toggle any config option we want to test. If we want to test totally different CVRs too, that would be a good candidate for creating different configs and replicating the `same_data_different_formats` directory.

src/test/java/network/brightspots/rcv/GuiFunctionalityTests.java
artoonie: Done! Asserted that all functions return the same candidate. Didn't check that there are zero aliases as that's both incidental to what we're testing, and would cause failures elsewhere.

yezr (dismissed):

yezr (approved):
lgtm

PR #997, opened 2026-05-15
fix test-and-release workflow flakiness: update dependencies

artoonie:
The nightly build is failing. That should be fixed in later versions of the gradle wrapper validation: https://github.com/gradle/wrapper-validation-action/issues/40

Using setup-gradle uses wrapper-validation and we don't need both: https://github.com/gradle/actions/tree/main/wrapper-validation

This PR essentially updates gradle/actions/wrapper-validation@v3 to gradle/actions/setup-gradle@v6 and actions/setup-java@v3 to actions/setup-java@v5 and the other changes are necessary to support that.

### DIFF
--- a/.github/actions/zip/action.yml
+++ b/.github/actions/zip/action.yml
@@ -27,4 +27,14 @@ runs:
     - name: "Generate zip for Windows"
       if: runner.os == 'Windows'
       shell: powershell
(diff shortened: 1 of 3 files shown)


yezr (approved):
LGTM

